PR #35, opened 2021-01-11
SQL-136: Set default batch size to 0 instead of 10

pmeredit:


rychipman (approved):
LGTM, based on offline conversation that this change has addressed the perf problems described in HELP-21296

PR #36, opened 2021-01-12
SQL-152: Fix empty-batch bug when fetchSize is zero

rychipman:
If we set batchSize to zero on an AggregateIterable, the cursor will not return any results. Instead of unconditionally setting batchSize to fetchSize, we should do so only when it's non-zero.


pmeredit (approved):
LGTM!

PR #37, opened 2021-01-12
SQL-153: Create basic getObject implementation

rychipman:
DBeaver was failing because it was calling `getObject` on LONGVARCHAR columns (not sure why it didn't use `getString`, but that's beside the point). This PR provides an implementation for `getObject`

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoResultSet.java
+++ b/src/main/java/com/mongodb/jdbc/MongoResultSet.java
@@ -24,6 +24,7 @@ import java.sql.SQLXML;
 import java.sql.Statement;
 import java.sql.Time;
 import java.sql.Timestamp;
+import java.sql.Types;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.Calendar;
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -813,14 +814,98 @@ public ResultSetMetaData getMetaData() throws SQLException {
...
     public Object getObject(int columnIndex) throws SQLException {
-        throw new SQLFeatureNotSupportedException(
-                Thread.currentThread().getStackTrace()[1].toString());
+		checkBounds(columnIndex);
+		int zeroIndex = columnIndex - 1;
+		BsonValue out = current.getValues().get(columnIndex - 1).value;
+		int columnType = rsMetaData.getColumnType(columnIndex);
+		switch (columnType) {
+			case Types.ARRAY:
+				// not supported
```
pmeredit, line 823: java switches are like c switches, so this types.ARRAY will fall through to types.INT and then fail with an error that will be strange to users. I suggest putting all the unsupported cases at the end so that it goes to the Exception have you at the bottom. Other option is to put in breaks in all of these, but I think that's more verbose.

rychipman: updated! I was starting out by writing a `throw` in each of these cases, and later decided to just factor it out to the end (but forgot to add the breaks in here)! I'm going to add breaks instead of reordering so that this can be left in alphabetical order

pmeredit (approved):
LGTM! (out dated)

pmeredit (changes requested):
We should have a unit test for this similar to what we have for `getInt` or `getString`

src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -813,14 +814,121 @@ public ResultSetMetaData getMetaData() throws SQLException {
 
     @Override
     public Object getObject(int columnIndex) throws SQLException {
-        throw new SQLFeatureNotSupportedException(
-                Thread.currentThread().getStackTrace()[1].toString());
+		checkBounds(columnIndex);
+		int zeroIndex = columnIndex - 1;
+		BsonValue out = current.getValues().get(columnIndex - 1).value;
```
huan-Mongo, line 819: I guess this is the intention of the above variable?
```suggestion
		BsonValue out = current.getValues().get(zeroIndex).value;
```

huan-Mongo (approved):
LGTM! Only one minor comment above.

rychipman:
@pmeredit added unit testing, and caught a few small bugs along the way!

src/main/java/com/mongodb/jdbc/MongoResultSet.java
rychipman: updated!

src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -813,14 +814,121 @@ public ResultSetMetaData getMetaData() throws SQLException {
 
     @Override
     public Object getObject(int columnIndex) throws SQLException {
-        throw new SQLFeatureNotSupportedException(
-                Thread.currentThread().getStackTrace()[1].toString());
+		System.out.println("getObject(" + columnIndex + ")");
```
pmeredit, line 817: need to remove the debug code :)

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
```diff
@@ -229,14 +229,24 @@
         try {
             mongoStatement = new MongoStatement(mongoConnection, "test", true);
         } catch (SQLException e) {
-            e.printStackTrace();
```
pmeredit, line 232: Hmm odd, `throw new RuntimeException(e)` is definitely better. I wonder why we had that.

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
```diff
@@ -229,14 +229,24 @@
...
         strictMongoResultSet =
                 new MongoResultSet(mongoStatement, new MongoExplicitCursor(mongoResultDocs), false);
         relaxedMongoResultSet =
                 new MongoResultSet(mongoStatement, new MongoExplicitCursor(mongoResultDocs), true);
         closedMongoResultSet =
                 new MongoResultSet(mongoStatement, new MongoExplicitCursor(mongoResultDocs), true);
+
+		// call next() so that each result set is on the pre-populated row.
+		try {
+			relaxedMongoResultSet.next();
```
pmeredit, line 245: Nice refactor

pmeredit (approved):
LGTM!!

src/main/java/com/mongodb/jdbc/MongoResultSet.java
rychipman: good catch! done.

PR #40, opened 2021-01-26
SQL-156: Use $sql formatVersion 2

rychipman:
This PR updates the JDBC driver to use $sql formatVersion 2.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/Column.java
+++ b/src/main/java/com/mongodb/jdbc/Column.java
@@ -1,7 +1,5 @@
 package com.mongodb.jdbc;
 
-import org.bson.BsonValue;
-
 public class Column {
     public String database;
     public String table;
@@ -9,7 +7,6 @@ public class Column {
     public String column;
     public String columnAlias;
     public String bsonType;
-    public BsonValue value;
 
     public Column() {}
 
@@ -19,15 +16,13 @@ public class Column {
             String tableAlias,
             String column,
             String columnAlias,
-            String bsonType,
-            BsonValue value) {
+            String bsonType) {
         this.database = database;
         this.table = table;
         this.tableAlias = tableAlias;
         this.column = column;
         this.columnAlias = columnAlias;
         this.bsonType = bsonType;
-        this.value = value;
     }
 
     @Override
@@ -50,9 +45,6 @@ public class Column {
                 + '\''
                 + ", bsonType="
                 + bsonType
-                + '\''
-                + ", value="
-                + value
                 + '}';
     }
 }
--- a/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
@@ -813,16 +813,17 @@ public class MongoDatabaseMetaData implements DatabaseMetaData {
         // No procedures so we always return an empty result set.
         BsonValue n = new BsonNull();
         ArrayList<MongoResultDoc> docs = new ArrayList<>();
-        MongoResultDoc doc = new MongoResultDoc();
-        doc.values = new ArrayList<>();
-        doc.values.add(new Column("", "", "", "PROCEDURE_CAT", "PROCEDURE_CAT", "string", n));
-        doc.values.add(new Column("", "", "", "PROCEDURE_SCHEM", "PROCEDURE_SCHEM", "string", n));
-        doc.values.add(new Column("", "", "", "PROCEDURE_NAME", "PROCEDURE_NAME", "string", n));
-        doc.values.add(new Column("", "", "", "REMARKS", "REMARKS", "string", n));
-        doc.values.add(new Column("", "", "", "PROCEDURE_TYPE", "PROCEDURE_TYPE", "int", n));
-        doc.values.add(new Column("", "", "", "SPECIFIC_NAME", "SPECIFIC_NAME", "string", n));
-        doc.emptyResultSet = true;
-        docs.add(doc);
+
+        MongoResultDoc metaDoc = new MongoResultDoc();
+        metaDoc.columns = new ArrayList<>();
+        metaDoc.columns.add(new Column("", "", "", "PROCEDURE_CAT", "PROCEDURE_CAT", "string"));
+        metaDoc.columns.add(new Column("", "", "", "PROCEDURE_SCHEM", "PROCEDURE_SCHEM", "string"));
+        metaDoc.columns.add(new Column("", "", "", "PROCEDURE_NAME", "PROCEDURE_NAME", "string"));
+        metaDoc.columns.add(new Column("", "", "", "REMARKS", "REMARKS", "string"));
+        metaDoc.columns.add(new Column("", "", "", "PROCEDURE_TYPE", "PROCEDURE_TYPE", "int"));
+        metaDoc.columns.add(new Column("", "", "", "SPECIFIC_NAME", "SPECIFIC_NAME", "string"));
+
+        docs.add(metaDoc);
         return new MongoResultSet(null, new MongoExplicitCursor(docs), true);
     }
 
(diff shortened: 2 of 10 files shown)


src/main/java/com/mongodb/jdbc/Column.java
```diff
@@ -19,15 +16,13 @@ public Column(
             String tableAlias,
             String column,
             String columnAlias,
-            String bsonType,
-            BsonValue value) {
```
rychipman, line 23: `Column` now represents just the metadata for a column

src/main/java/com/mongodb/jdbc/MongoResultDoc.java
```diff
@@ -4,36 +4,20 @@
 import java.util.List;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
+import org.bson.BsonValue;
 
 public class MongoResultDoc {
-    public Boolean emptyResultSet;
-    public List<Column> values;
+    public List<Column> columns;
```
rychipman, line 10: MongoResultDoc will now be one of two things:
- The first metadata doc, in which case `columns` is non-null and `values` is null
- The remaining result-set docs, in which case `columns` is null and `values` is non-null

src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
```diff
@@ -503,10 +521,122 @@ public boolean isDefinitelyWritable(int column) throws SQLException {
     // --------------------------JDBC 2.0-----------------------------------
 
     @Override
+	// TODO
     public String getColumnClassName(int column) throws SQLException {
         checkBounds(column);
-        Object o = mongoResultDoc.values.get(column - 1).value;
-        return o.getClass().getName();
```
rychipman, line 509: It turns out that the previous implementation of this function was incorrect. It's supposed to return the name of the class that would be returned by `resultSet.getObject()`, which does _not_ return BSON values

src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
```diff
@@ -1,28 +1,46 @@
...
 import java.sql.SQLException;
 import java.sql.Types;
+import java.sql.Timestamp;
 import org.bson.BsonType;
 
 public class MongoResultSetMetaData implements ResultSetMetaData {
-    private MongoResultDoc mongoResultDoc;
-
+    private List<Column> columns;
+	private HashMap<String, Integer> columnPositions;
```
rychipman, line 14: I moved the column position cache inside this class, since column labels are really a property of result set metadata, not the result set itself

src/main/java/com/mongodb/jdbc/MongoStatement.java
```diff
@@ -74,8 +74,6 @@ public ResultSet executeQuery(String sql) throws SQLException {
             return resultSet;
         } catch (MongoExecutionTimeoutException e) {
             throw new SQLTimeoutException(e);
-        } catch (Exception e) {
-            throw new SQLException(e);
```
rychipman, line 78: Is there a reason we were doing this before? It just seems to obscure the source of the exception

src/test/java/com/mongodb/jdbc/MongoMock.java
```diff
@@ -62,11 +62,11 @@ protected static void resetMockObjs() throws NoSuchFieldException {
     }
 
     Column generateCol(
-            String database, String table, String column, String bsonType, BsonValue value) {
-        return new Column(database, table, table, column, column, bsonType, value);
+            String database, String table, String column, String bsonType) {
+        return new Column(database, table, table, column, column, bsonType);
     }
 
-    MongoResultDoc generateRow(boolean isEmpty) {
```
rychipman, line 69: split this into separate functions for generating the metadata doc and the actual docs representing each row

src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -62,29 +61,35 @@
...
         this.cursor = cursor;
         this.relaxed = relaxed;
+
         // dateFormat is not thread safe, so we do not want to make it a static field.
         TimeZone UTC = TimeZone.getTimeZone("UTC");
         dateFormat.setTimeZone(UTC);
+
+		// iterate the cursor to get the metadata doc
+		MongoResultDoc metadataDoc = cursor.next();
+		rsMetaData = new MongoResultSetMetaData(metadataDoc);
```
rychipman, line 92: we now construct the ResultSetMetaData from the first doc.

src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -805,35 +743,26 @@ public String getCursorName() throws SQLException {
     @Override
     public ResultSetMetaData getMetaData() throws SQLException {
         checkClosed();
-        checkAndCacheFirstDocAndMetaData();
-        if (current != null) {
-            rsMetaData = new MongoResultSetMetaData(current);
-        }
-        return rsMetaData;
```
rychipman, line 812: we've been able to remove a good amount of edge-case handling around the first doc and empty result sets now that empty result sets are no longer a special case

src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -93,36 +98,22 @@ MongoResultDoc getCurrent() {
         return current;
     }
 
+	private BsonValue getBsonValue(int columnIndex) throws SQLException {
```
rychipman, line 101: had to change all the lines in each `getX` function that were getting the column index from the label because the column position cache was moved to the ResultSetMetadta. While I was changing that, I factored this logic out into a single helper function, so that it's not duplicated in each getter

src/test/java/com/mongodb/jdbc/MongoResultSetMetaDataTest.java
```diff
@@ -280,20 +289,15 @@ void testGetColumnType() throws SQLException {
...
-                BsonDecimal128.class.getName(), resultSetMetaData.getColumnClassName(DECIMAL_COL));
+        assertEquals(null, resultSetMetaData.getColumnClassName(NULL_COL));
+        assertEquals(double.class.getName(), resultSetMetaData.getColumnClassName(DOUBLE_COL));
+        assertEquals(String.class.getName(), resultSetMetaData.getColumnClassName(STRING_COL));
+        assertEquals(String.class.getName(), resultSetMetaData.getColumnClassName(OBJECTID_COL));
+        assertEquals(boolean.class.getName(), resultSetMetaData.getColumnClassName(BOOLEAN_COL));
+        assertEquals(Timestamp.class.getName(), resultSetMetaData.getColumnClassName(DATE_COL));
+        assertEquals(int.class.getName(), resultSetMetaData.getColumnClassName(INTEGER_COL));
+        assertEquals(int.class.getName(), resultSetMetaData.getColumnClassName(LONG_COL));
+        assertEquals(BigDecimal.class.getName(), resultSetMetaData.getColumnClassName(DECIMAL_COL));
```
rychipman, line 300: updated to match the corrected `getColumnClassName` behavior

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
```diff
@@ -1558,19 +1545,26 @@ void testEmptyResultSetWhenGetMetadataCalledFirst() throws SQLException {
     }
 
     @Test
-    void differentMetaDataOnDifferentRows() throws SQLException {
+    void sameMetaDataOnDifferentRowsEvenWithDifferentBsonTypes() throws SQLException {
```
rychipman, line 1554: I've changed this test, because the old test was testing something that could never happen: getting different metadata info for the same column across multiple rows

src/main/java/com/mongodb/jdbc/Column.java
pmeredit: thanks for this comment!

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1952,165 +1981,111 @@ public boolean autoCommitFailureClosesAllResultSets() throws SQLException {
 
     @Override
     public ResultSet getClientInfoProperties() throws SQLException {
-        ArrayList<MongoResultDoc> rows = new ArrayList<>(4);
+        ArrayList<MongoResultDoc> docs = new ArrayList<>();
```
pmeredit, line 1984: Why did we remove the size hint on this one?

src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
pmeredit: Great catch

src/main/java/com/mongodb/jdbc/MongoStatement.java
pmeredit: It turns any other exception into a SQLException. Since it compiles without this, I'd say it's fine.

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
pmeredit: Yep, it used to be theoretically possible, now it is not.

pmeredit (approved):
LGTM, modulo the one tiny question about size hints.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
rychipman: I don't think this was intentional. My guess would be that I just deleted the original row and pasted in the new one, which resulted in the size hint getting lost!

src/main/java/com/mongodb/jdbc/MongoResultSet.java
pmeredit: This part makes me very happy. The handling between empty and one result was gross.

src/main/java/com/mongodb/jdbc/MongoResultSet.java
huan-Mongo, line 92:
```suggestion
	// iterate the cursor to get the metadata doc
	MongoResultDoc metadataDoc = cursor.next();
	rsMetaData = new MongoResultSetMetaData(metadataDoc); 
```

huan-Mongo (approved):
LGTM! Looks good in general except for some weird linting issues. I suggest you can run `./gradlew spotlessApply` and this should be auto fixed!

One small issue is in `MongoStatement`, I think `catch (Exception e) {` should still be preserved.

src/main/java/com/mongodb/jdbc/MongoStatement.java
rychipman: per slack discussion, we are only going to catch & convert exceptions that we _know_ could arise, not exceptions which are unexpected

PR #41, opened 2021-09-10
SQL-447: Set client metadata in JDBC driver

bucaojit:
This PR adds the driver name and version (major.minor) to the client metadata.  
Obtained from `getClass().getPackage().getImplementationTitle()` and `getClass().getPackage().getImplementationVersion()`, otherwise defaults to 'mongodb-jdbc' and 0.0.

Log from ADL:
```
"clientMetadata": {"driver":{"name":"mongo-java-driver|sync|mongodb-jdbc","version":"3.12.10|1.0"} ...
```

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -2,6 +2,7 @@ package com.mongodb.jdbc;
 
 import com.google.common.base.Preconditions;
 import com.mongodb.ConnectionString;
+import com.mongodb.MongoDriverInformation;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoClients;
 import com.mongodb.client.MongoDatabase;
@@ -48,7 +49,16 @@ public class MongoConnection implements Connection {
         this.url = cs.getConnectionString();
         this.user = cs.getUsername();
         this.currentDB = database;
-        mongoClient = MongoClients.create(cs);
+
+        StringBuilder version = new StringBuilder();
+        version.append(MongoDriver.MAJOR_VERSION).append(".").append(MongoDriver.MINOR_VERSION);
+        mongoClient =
+                MongoClients.create(
+                        cs,
+                        MongoDriverInformation.builder()
+                                .driverName(MongoDriver.NAME)
+                                .driverVersion(version.toString())
+                                .build());
         relaxed = conversionMode == null || !conversionMode.equals("strict");
         isClosed = false;
     }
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -86,6 +87,8 @@
             MAJOR_VERSION = 0;
             MINOR_VERSION = 0;
         }
+        String name = unit.getClass().getPackage().getImplementationTitle();
+        NAME = (name != null) ? name : "mongodb-jdbc";
```
rychipman, line 91: Under what circumstances would we expect `name` to be `null`?

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -48,7 +49,16 @@ public MongoConnection(ConnectionString cs, String database, String conversionMo
         this.url = cs.getConnectionString();
         this.user = cs.getUsername();
         this.currentDB = database;
-        mongoClient = MongoClients.create(cs);
+
+        StringBuilder version = new StringBuilder();
+        version.append(MongoDriver.MAJOR_VERSION).append(".").append(MongoDriver.MINOR_VERSION);
```
rychipman, line 54: I think we want the whole version (not just major and minor) here -- AFAICT, the only reason we've broken these two version components out into constants and not others is because the JDBC API only has methods for major & minor.

rychipman (changes requested):

src/main/java/com/mongodb/jdbc/MongoDriver.java
bucaojit: I found when running the gradle test during `gradle clean build`, the values returned for `getImplementationTitle()` and `getImplementationVersion()` are `null`.  

It returns the expected version and name values when running a program that uses the `mongodb-jdbc-1.0.4-SNAPSHOT.jar` JAR file.  I think it has something to do with the `META-INF\MANIFEST.MF` file which exists in the JAR.

src/main/java/com/mongodb/jdbc/MongoConnection.java
bucaojit: Sounds good.  
I was looking for a way to use the default value that MAJOR_VERSION and MINOR_VERSION specifies if VERSION is `null` (currently set to 0.0).  I can add a check to use VERSION if it is not null and MAJOR_VERSION.MINOR_VERSION in case it is null.

rychipman (approved):
LGTM!

pmeredit (approved):
LGTM!

PR #42, opened 2021-09-21
SQL-213: Rationalize get/setAutoCommit behavior

bucaojit:
This PR modifies the `getAutoCommit()` to always return false. 
Modifies `setAutoCommit()` to throw SQLFeatureNotSupportedException if input parameter is set to `true`, no-op in case of `false`.


rychipman (approved):
LGTM

pmeredit (approved):
LGTM!

PR #43, opened 2021-09-24
SQL-169: Split out JDBC driver unit and integration tests from build …

bucaojit:
…task

Modified Evergreen configuration to run unit test and integration test as separate tasks from `build`

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -34,7 +34,7 @@ buildvariants:
       - name: "build"
       - name: "test-unit"
       - name: "test-integration"
-      - name: "tdvt"
+      - name: "test-tdvt"
   - name: release
     display_name: "Release"
     expansions:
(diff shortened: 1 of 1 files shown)


.evg.yml
```diff
@@ -32,6 +32,8 @@ buildvariants:
     run_on: ubuntu1804-large
     tasks:
       - name: "build"
+      - name: "unittest"
+      - name: "integrationtest"
```
rychipman, line 36:
```suggestion
      - name: "test-unit"
      - name: "test-integration"
```
just a small naming suggestion with a couple small benefits:
- the test tasks get automatically grouped together by the UI if/when it sorts tasks lexicographically
- easier to filter down to the correctness-test tasks by prefix filtering

It might make sense to apply this same convention to the `tdvt` task name as well

.evg.yml
```diff
@@ -46,6 +48,18 @@ tasks:
     commands:
       - func: "build jdbc driver"
 
+  - name: "unittest"
+    depends_on:
+        "build"
```
rychipman, line 53: I would suggest that we remove the dependency on the `build` task here, for the integration tests, and for the tdvt tests.
These tasks don't actually depend on any artifacts generated by `build`, so I think all this really does is increase the makespan of patches

rychipman (changes requested):
looks good overall! just a couple small things

.evg.yml
bucaojit: Sounds good, I'll add the test prefix

.evg.yml
bucaojit: Removed the dependencies on `build` for unit, integration, and tdvt tests.

rychipman (approved):
LGTM

pmeredit (approved):
Great change! LGTM

PR #44, opened 2021-09-28
SQL-507: Refactor into base classes and MySQL and MongoSQL subclasses

mattChiaravalloti:
This PR refactors the base `Mongo*` classes into abstract classes and introduces completed `MySQL*` implementations and stub `MongoSQL*` implementations. There are no logic changes in this PR. All changes are purely refactors.


mattChiaravalloti:
Replaced @rychipman  with @bucaojit  as a reviewer, based on discussion in stand up.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -376,14 +324,7 @@ public SQLXML createSQLXML() throws SQLException {
                 Thread.currentThread().getStackTrace()[1].toString());
     }
 
-    private void validateConn() throws SQLException {
```
pmeredit, line 379: We don't need to make this abstract. We just need to change `"SELECT 1 from DUAL" to "SELECT 1"`, since that works in MySQL.

pmeredit (changes requested):
One change, looks great so far

src/main/java/com/mongodb/jdbc/MongoPreparedStatement.java
```diff
@@ -4,32 +4,18 @@
...
+public class MongoPreparedStatement implements PreparedStatement {
     private String sql;
+    private MongoStatement statement;
 
     public MongoPreparedStatement(
             String sql, MongoConnection conn, String databaseName, boolean relaxed)
             throws SQLException {
-        super(conn, databaseName, relaxed);
         this.sql = sql;
+        this.statement = new MySQLStatement(conn, databaseName, relaxed);
```
bucaojit, line 18: The design doc describes a slightly different approach where the constructor takes a MongoStatement object.  That would help in case we don't know here whether to construct a MySQL or MongoSQL statement

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -97,28 +97,14 @@ String getServerVersion() throws SQLException {
...
     @Override
-    public Statement createStatement() throws SQLException {
-        checkConnection();
-        try {
-            return new MongoStatement(this, currentDB, relaxed);
-        } catch (IllegalArgumentException e) {
-            throw new SQLException(e);
-        }
-    }
+    public abstract Statement createStatement() throws SQLException;
```
bucaojit, line 100: I have a general observation on this function and other ones like it: createStatement() is part of the Connection interface and does not need to be redeclared here.  The subclass will still need to implement all the interface functions.
It may be cleaner to remove the Overriden functions that are redeclared as abstract.

bucaojit (commented):
Great work!  Lots to review.  
I have a few comments for now and will continue reviewing tomorrow.

src/main/java/com/mongodb/jdbc/MongoConnection.java
mattChiaravalloti: That's a very good point, I'll do an audit of which ones don't need to be redeclared and make the updates 👍

src/main/java/com/mongodb/jdbc/MongoPreparedStatement.java
mattChiaravalloti: I figured I'd leave that to [SQL-511: Finish MongoSQL Statement](https://jira.mongodb.org/browse/SQL-511) . @pmeredit did you have an idea/preference of which ticket would cover this change?

src/main/java/com/mongodb/jdbc/MongoConnection.java
mattChiaravalloti: Oh right, I must have breezed past that part of the design when going through and checking off methods. I'll address this.

src/main/java/com/mongodb/jdbc/MongoPreparedStatement.java
bucaojit: Thanks for pointing that out.  I'm fine with it being handled in SQL-511.  👍

src/main/java/com/mongodb/jdbc/MongoConnection.java
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoConnection.java
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLConnection.java
```diff
@@ -0,0 +1,74 @@
+package com.mongodb.jdbc;
+
+import com.mongodb.ConnectionString;
+
+import java.sql.*;
+
+public class MongoSQLConnection extends MongoConnection implements Connection {
```
mattChiaravalloti, line 7: General Java style question:
For all of these new hierarchies, the base `abstract class` always `implements` some JDBC interface. The subclasses `MySQL*` and `MongoSQL*` always extend that base class and therefore inherit the interface. My question is: is it stylistically frowned upon to explicitly include these `implements XYZ` declarations in the subclasses, or is it nice to have it stated explicitly?

It works with or without this so it seems to come down to style and I can't really find anything about this online.

src/test/java/com/mongodb/jdbc/MongoMock.java
```diff
@@ -26,19 +28,31 @@
     @Mock protected static MongoCursor<MongoResultDoc> mongoCursor;
 
     @InjectMocks
-    protected static MongoConnection mongoConnection = new MongoConnection(uri, database, null);
+    protected static MongoConnection mongoConnection = new MySQLConnection(uri, database, null);
+
+    private static Field getDeclaredFieldFromClassOrSuperClass(Class c, String fieldName) throws NoSuchFieldException {
```
mattChiaravalloti, line 33: What do you think of this for fixing the problem of getting a field from a super class? I made it generic even though we really only need it for `MongoConnection` and really only need to go up one level of the hierarchy tree.

src/main/java/com/mongodb/jdbc/MongoSQLConnection.java
bucaojit: The few posts I see online give reasons for and against keeping the `implements` section.  
Arguments against are that it's not needed, redundant, and may affect some IDE's refactor features.  Arguments for it  are in case the parent class no longer implements the interface, it clearly shows what interface is implemented, and that JDK classes such as LinkedHashMap do it this way. 
My vote is to include the `implements interface` declarations.

src/main/java/com/mongodb/jdbc/MongoConnection.java
bucaojit: Thanks for making the change, it's clearer now what functionality is provided in the abstract class.

src/main/java/com/mongodb/jdbc/MongoSQLConnection.java
mattChiaravalloti: That's a good point that changing the base class to not implement the interface could lead to issues. I also vote for including them.

src/test/java/com/mongodb/jdbc/MongoMock.java
bucaojit: That's a good way to handle the problem of the field being in the class or any of the parent classes. 👍

bucaojit (approved):
Looks good!

I see the unit tests are working now. 
The changes match what is described in the design documents.  
I am fine with whether we keep or remove the additional 'implements X' in the subclass, prefer to keep.

src/main/java/com/mongodb/jdbc/MongoPreparedStatement.java
pmeredit: sgtm

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -172,158 +170,6 @@ public String getIdentifierQuoteString() throws SQLException {
...
-    public String getSystemFunctions() throws SQLException {
-        return "DATABASE,USER,SYSTEM_USER,SESSION_USER,VERSION";
-    }
-
-    @Override
-    public String getTimeDateFunctions() throws SQLException {
-        return MongoFunction.dateFunctionsString;
-    }
-
     @Override
```
pmeredit, line 173: Anything returning a `MySQLResultSet` needs to be made abstract and overridden in the subclass. I guess I missed a few in the design doc!

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -864,76 +710,23 @@ public ResultSet getProcedureColumns(
         metaDoc.columns.add(new Column("", "", "", "SPECIFIC_NAME", "SPECIFIC_NAME", "string"));
 
         docs.add(metaDoc);
-        return new MongoResultSet(null, new MongoExplicitCursor(docs), true);
+        return new MySQLResultSet(null, new MongoExplicitCursor(docs), true);
     }
 
-    private String patternCond(String colName, String pattern) {
+    protected String patternCond(String colName, String pattern) {
```
pmeredit, line 716: Since this is specific to MySQL, I'd just move it into the subclass.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -864,76 +710,23 @@ public ResultSet getProcedureColumns(
...
         if (pattern == null) {
             return " 1 "; //just return a true condition.
         }
         return " " + colName + " like '" + escapeString(pattern) + "' ";
     }
 
-    private String equalsCond(String colName, String literal) {
+    protected String equalsCond(String colName, String literal) {
```
pmeredit, line 723: Ditto

pmeredit (changes requested):
Not done yet, but here are a few things :)

src/main/java/com/mongodb/jdbc/MongoPreparedStatement.java
```diff
@@ -4,32 +4,18 @@
...
-import java.sql.Ref;
-import java.sql.ResultSet;
-import java.sql.ResultSetMetaData;
-import java.sql.RowId;
-import java.sql.SQLException;
-import java.sql.SQLFeatureNotSupportedException;
-import java.sql.SQLXML;
-import java.sql.Time;
-import java.sql.Timestamp;
+import java.sql.*;
```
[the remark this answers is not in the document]
mattChiaravalloti: Ah, I need to change my IDE settings -- I dislike that this is the default behavior. I'll undo this and get back all of the individual imports.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
mattChiaravalloti: Ok I'll look through and make sure I catch all of these! I can also add them to the design doc.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
mattChiaravalloti: Good catch. I went a little bulk-edit-happy and didn't fully inspect all of the helper methods I changed from `private` -> `protected`.

pmeredit:
@mattChiaravalloti I updated the Design Doc to cover _all_ the public methods that need to be moved out of `MongoDatabaseMetadata`. They are now in utf-8 order,  also

src/main/java/com/mongodb/jdbc/MongoPreparedStatement.java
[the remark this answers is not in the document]
mattChiaravalloti: done

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
mattChiaravalloti: done

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
mattChiaravalloti: done

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
mattChiaravalloti: done

mattChiaravalloti:
> @mattChiaravalloti I updated the Design Doc to cover _all_ the public methods that need to be moved out of `MongoDatabaseMetadata`. They are now in utf-8 order, also

thanks! updated

mattChiaravalloti:
The evergreen failures are the same ones that are already failing on master, btw

pmeredit (approved):
LGTM!

PR #45, opened 2021-10-06
SQL-510: Implement MongoSQLConnection

pmeredit:
Various refactoring of things that can actually be shared between MySQL and MongoSQL, as well.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
+++ b/src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
@@ -8,7 +8,10 @@ import java.sql.Statement;
 import java.sql.Types;
 import java.util.ArrayList;
 import java.util.regex.Pattern;
-import org.bson.*;
+import org.bson.BsonInt32;
+import org.bson.BsonNull;
+import org.bson.BsonString;
+import org.bson.BsonValue;
 
 public class MySQLDatabaseMetaData extends MongoDatabaseMetaData implements DatabaseMetaData {
 
--- a/src/main/java/com/mongodb/jdbc/MongoSQLStatement.java
+++ b/src/main/java/com/mongodb/jdbc/MongoSQLStatement.java
@@ -6,8 +6,7 @@ import java.sql.SQLFeatureNotSupportedException;
 import java.sql.Statement;
 
 public class MongoSQLStatement extends MongoStatement implements Statement {
-    public MongoSQLStatement(MongoConnection conn, String databaseName)
-            throws SQLException {
+    public MongoSQLStatement(MongoConnection conn, String databaseName) throws SQLException {
         super(conn, databaseName);
     }
 


src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -194,6 +194,29 @@ public void clearWarnings() throws SQLException {
     }
 
     // --------------------------JDBC 2.0-----------------------------
+    @Override
```
pmeredit, line 197: The 2.0 and 3.0 methods were moved out of the MySQLConnection class into MongoConnection super class because they all just check the direction of cursor then call `createStatement` and `prepareStatement`, so now those are all we need to implement in the subclasses.

src/main/java/com/mongodb/jdbc/MongoPreparedStatement.java
```diff
@@ -29,10 +29,10 @@
     private MongoStatement statement;
 
     public MongoPreparedStatement(
-            String sql, MongoConnection conn, String databaseName, boolean relaxed)
+            String sql, MongoStatement statement)
```
pmeredit: This change we agreed to leave out of the last PR, so I did it now. `prepareStatement` is part of the `Connection` class, so it makes sense to do now.

src/main/java/com/mongodb/jdbc/MongoSQLConnection.java
```diff
@@ -11,69 +12,31 @@
 
 public class MongoSQLConnection extends MongoConnection implements Connection {
 
-    public MongoSQLConnection(ConnectionString cs, String database, String conversionMode) {
-        super(cs, database, conversionMode);
+    public MongoSQLConnection(ConnectionString cs, String database) {
+        super(cs, database);
     }
```
pmeredit, line 14: The `conversionMode` is only relevant to `MySQL`, so we should not have it in the base class.

src/test/java/com/mongodb/jdbc/MongoSQLConnectionTest.java
```diff
@@ -0,0 +1,159 @@
+package com.mongodb.jdbc;
```
pmeredit, line 1: We need separate unit tests for MySQL and MongoSQL, unfortunately this code is very redundant

src/test/java/com/mongodb/jdbc/MySQLConnectionTest.java
```diff
@@ -0,0 +1,159 @@
+package com.mongodb.jdbc;
```
pmeredit, line 1: `MongoSQLConnectionTest` moved to `MySQLConnectionTest`

rychipman (approved):
LGTM

src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
```diff
@@ -10,6 +8,7 @@
 import java.sql.Types;
 import java.util.ArrayList;
 import java.util.regex.Pattern;
+import org.bson.*;
```
mattChiaravalloti, line 11: I may have introduced this wildcard import last PR. I thought I got rid of all of them, but must have missed this one. Do you want to expand it to just the imports we need? Your call.

mattChiaravalloti (approved):
Looks good apart from the wildcard import you may or may not choose to change 👍

src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
pmeredit: Oh, I didn't even notice this. I'll fix it! I think it did come from the last PR.

PR #46, opened 2021-10-06
SQL-511: Implement MongoSQLStatement

pmeredit:


src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -1,25 +1,32 @@
...
     // This is only used for testing, and that is why it has package level access, and the
     // tests have been moved into this package.
-    MongoResultDoc getCurrent() {
+    BsonDocument getCurrent() {
         return current;
     }
 
+    @Override
+    protected boolean checkNull(BsonValue o) {
+        // TODO, actually check if null
```
rychipman, line 26: let's correctly implement this check before merging! alternatively, if we want to leave it for a future ticket, let's throw a TODO exception like we're doing elsewhere instead of returning a hard-coded value

pmeredit: Exception sounds good, this is out of scope for this ticket.

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -0,0 +1,145 @@
...
+    // to replace lambda as input in the testExceptionAfterConnectionClosed
+    interface TestInterface {
+        void test() throws SQLException;
+    }
+
+    Column generateCol(String database, String table, String column, String bsonType) {
+        return new Column(database, table, table, column, column, bsonType);
+    }
+
+    MongoResultDoc generateMetadataDoc() {
```
mattChiaravalloti, line 83: Can we just remove this now since MongoSQL doesn't have metadata docs? Or would you prefer to revisit this as part of SQL-535?

pmeredit: In the next PR it's already removed :)

PR #47, opened 2021-10-12
SQL-513: Implement MongoSQLResultSetMetaData

pmeredit:
Ultimately, I needed to use an `ExtendedBsonType` so that we could easily represent dynamically types/unknown type column

EDIT: We moved away from this approach, now everything is in a MongoColumnInfo class

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoColumnInfo.java
+++ b/src/main/java/com/mongodb/jdbc/MongoColumnInfo.java
@@ -55,7 +55,7 @@ public interface MongoColumnInfo {
 
     public boolean isPolymorphic();
 
-    public BsonType getBsonType();
+    public BsonType getBsonTypeEnum();
 
     public String getBsonTypeName();
 
--- a/src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
+++ b/src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
@@ -118,7 +118,7 @@ public abstract class MongoResultSetMetaData implements ResultSetMetaData {
         if (ci.isPolymorphic()) {
             return true;
         }
-        BsonType t = ci.getBsonType();
+        BsonType t = ci.getBsonTypeEnum();
         switch (t) {
             case ARRAY:
             case BINARY:
@@ -153,7 +153,7 @@ public abstract class MongoResultSetMetaData implements ResultSetMetaData {
         if (ci.isPolymorphic()) {
             return true;
         }
-        BsonType t = ci.getBsonType();
+        BsonType t = ci.getBsonTypeEnum();
         switch (t) {
             case DOUBLE:
             case DECIMAL128:
@@ -188,7 +188,7 @@ public abstract class MongoResultSetMetaData implements ResultSetMetaData {
         if (ci.isPolymorphic()) {
             return unknownLength;
         }
-        BsonType t = ci.getBsonType();
+        BsonType t = ci.getBsonTypeEnum();
         switch (t) {
             case ARRAY:
                 return unknownLength;
@@ -243,7 +243,7 @@ public abstract class MongoResultSetMetaData implements ResultSetMetaData {
         if (ci.isPolymorphic()) {
             return unknownLength;
         }
-        BsonType t = ci.getBsonType();
+        BsonType t = ci.getBsonTypeEnum();
         switch (t) {
             case ARRAY:
                 return unknownLength;
@@ -297,7 +297,7 @@ public abstract class MongoResultSetMetaData implements ResultSetMetaData {
         if (ci.isPolymorphic()) {
             return unknownLength;
         }
-        BsonType t = ci.getBsonType();
+        BsonType t = ci.getBsonTypeEnum();
         switch (t) {
             case ARRAY:
             case BINARY:
--- a/src/main/java/com/mongodb/jdbc/MongoSQLColumnInfo.java
+++ b/src/main/java/com/mongodb/jdbc/MongoSQLColumnInfo.java
@@ -126,7 +126,7 @@ public class MongoSQLColumnInfo implements MongoColumnInfo {
     }
 
     @Override
-    public BsonType getBsonType() {
+    public BsonType getBsonTypeEnum() {
         return bsonType;
     }
 
--- a/src/main/java/com/mongodb/jdbc/MySQLColumnInfo.java
+++ b/src/main/java/com/mongodb/jdbc/MySQLColumnInfo.java
@@ -32,6 +32,8 @@ public class MySQLColumnInfo implements MongoColumnInfo {
         nullTypes.add("undefined");
     }
 
+    public MySQLColumnInfo() {}
+
     public MySQLColumnInfo(
             String database,
             String table,
@@ -141,7 +143,7 @@ public class MySQLColumnInfo implements MongoColumnInfo {
     }
 
     @Override
-    public BsonType getBsonType() {
+    public BsonType getBsonTypeEnum() {
         return bsonTypeEnum;
     }
 


src/main/java/com/mongodb/jdbc/MySQLResultSetMetaData.java
```diff
@@ -3,12 +3,29 @@
...
+import java.util.HashMap;
+import java.util.List;
 
 public class MySQLResultSetMetaData extends MongoResultSetMetaData implements ResultSetMetaData {
+    protected List<Column> columns;
+    protected HashMap<String, Integer> columnPositions;
 
     public MySQLResultSetMetaData(MongoResultDoc metadataDoc) {
-        super(metadataDoc);
+        columns = metadataDoc.columns;
```
pmeredit, line 14: This was just moved from the super class since it was specific to `MySQL`, as mentioned above.

src/test/java/com/mongodb/jdbc/MongoConnectionTest.java
```diff
@@ -21,7 +21,7 @@
 @ExtendWith(MockitoExtension.class)
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 @MockitoSettings(strictness = Strictness.WARN)
-class MongoConnectionTest extends MongoMock {
+class MongoConnectionTest extends MySQLMock {
```
pmeredit, line 24: This is also covered in the other outstanding PR, changed just enough to fix it here (so not split into two test classes)

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -0,0 +1,193 @@
...
+        // Mock MongoCursor
+        when(mongoCursor.hasNext()).thenReturn(false);
+    }
+
+    // to replace lambda as input in the testExceptionAfterConnectionClosed
+    interface TestInterface {
+        void test() throws SQLException;
+    }
+
+    protected static MongoJsonSchema generateMongoJsonSchema() {
```
pmeredit, line 79: Note that both the datasources and keys within them are not in lexicographical order to show that they are properly sorted in the test

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -0,0 +1,193 @@
...
+        botSchema.properties.put("str", strSchema);
+
+        schema.properties = new HashMap<String, MongoJsonSchema>();
+        schema.properties.put("foo", fooSchema);
+        schema.properties.put("", botSchema);
+        return schema;
+    }
+
+    BsonDocument generateRow() {
+        /**/
```
pmeredit, line 190: This will be filled in in the `MongoSQLResultSet` ticket

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
```diff
@@ -0,0 +1,260 @@
...
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.MockitoAnnotations;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+@ExtendWith(MockitoExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class MongoSQLResultSetMetaDataTest extends MongoSQLMock {
+    private static MongoResultSetMetaData resultSetMetaData;
+
```
pmeredit, line 20: note while it says `__bot` here to make the comment easier to read, below in the `getTable` test we actually get `""` as the result.

src/main/java/com/mongodb/jdbc/ExtendedBsonType.java
```diff
@@ -0,0 +1,61 @@
...
+    /** A BSON 64-bit integer. */
+    INT64(0x12),
+    /**
+     * A BSON Decimal128.
+     *
+     * @since 3.4
+     */
+    DECIMAL128(0x13),
+    /** This encodes an unknown, poissibly dynamically changing BSON type */
+    ANY(0x14),
```
rychipman, line 50: I don't think this is the correct abstraction to use for representing dynamic/heterogeneous types. The fact that `ExtendedBSONType` and `ANY` are leaking out of the MongoSQL-specific classes and polluting the abstract/MySQL classes seems like a significant code smell to me.

I think that we should probably be handling/representing this in `MongoSQLColumnTypeInfo`. Happy to hop on a zoom call to discuss this one in person, since it's a higher-level architectural question that might not lend itself well to github threads

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
```diff
@@ -2,57 +2,144 @@
...
+    // This gets the datasource for a given columnLabel, and is used
+    // in MongoSQLResultSet to retrieve data by label.
+    String getDatasource(String columnLabel) {
+        return columnLabels.get(columnLabel).datasource;
+    }
+
+    private static void assertObjectSchema(MongoJsonSchema schema) throws SQLException {
+        if (schema.bsonType == null
+                || !schema.bsonType.equals("object")
+                || schema.properties == null) {
```
mattChiaravalloti, line 48: Why are properties required? Isn't the following a valid object schema:
```
{
    bsonType: "object",
    additionalProperties: true
}
```
? 

Or is this invalid because missing/null `properties` is not equivalent to `properties: {}`?

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
```diff
@@ -2,57 +2,144 @@
...
+            throw new SQLException("ResultSetMetaData json schema must be object with properties");
+        }
+    }
+
+    private void processDataSource(MongoJsonSchema schema, String datasource) throws SQLException {
+        MongoJsonSchema datasourceSchema = schema.properties.get(datasource);
+        assertObjectSchema(datasourceSchema);
+
+        Object[] columns = datasourceSchema.properties.keySet().toArray();
```
mattChiaravalloti, line 57:
```suggestion
        String[] columns = datasourceSchema.properties.keySet().toArray(new String[0]);
```
We could get the columns as Strings from the start and avoid the cast in the loop.

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
```diff
@@ -2,57 +2,144 @@
...
+        Arrays.sort(columns);
+
+        for (Object column : columns) {
+            String columnAsStr = (String) column;
+            MongoJsonSchema columnSchema = datasourceSchema.properties.get(columnAsStr);
+            columnIndices.add(new NameSpace(datasource, columnAsStr));
+            int subNullability =
```
mattChiaravalloti, line 64: What does the prefix "sub" mean here? Why not just call this `nullability` or `nullable`?

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
```diff
@@ -2,57 +2,144 @@
...
+        columnLabels = new HashMap<String, DatasourceAndIndex>();
+        columnIndices = new ArrayList<NameSpace>();
+        columnTypeInfo = new ArrayList<MongoSQLColumnTypeInfo>();
+
+        Object[] datasources = schema.properties.keySet().toArray();
+        Arrays.sort(datasources);
+
+        for (Object datasource : datasources) {
+            String datasourceAsString = (String) datasource;
+            processDataSource(schema, datasourceAsString);
```
mattChiaravalloti, line 89:
```suggestion
        String[] datasources = schema.properties.keySet().toArray(new String[0]);
        Arrays.sort(datasources);

        for (string datasource : datasources) {
            processDataSource(schema, datasource);
```
same can be done here

src/main/java/com/mongodb/jdbc/MongoSQLColumnTypeInfo.java
```diff
@@ -0,0 +1,113 @@
...
+                && schema.anyOf == null
+                && schema.required == null
+                && schema.items == null
+                && schema.additionalProperties == false;
+    }
+
+    private void constructFromAnyOf(MongoJsonSchema schema, int nullable) throws SQLException {
+        if (schema.anyOf == null) {
+            throw new SQLException(
+                    "both bsonType and anyOf are null and this is not ANY, this is not a valid schema");
```
mattChiaravalloti, line 49:
```suggestion
                    "invalid schema: both bsonType and anyOf are null and this is not ANY");
```
[nit] phrasing

src/main/java/com/mongodb/jdbc/MongoSQLColumnTypeInfo.java
```diff
@@ -0,0 +1,113 @@
...
+    private boolean isAny(MongoJsonSchema schema) {
+        return schema.bsonType == null
+                && schema.properties == null
+                && schema.anyOf == null
+                && schema.required == null
+                && schema.items == null
+                && schema.additionalProperties == false;
+    }
+
+    private void constructFromAnyOf(MongoJsonSchema schema, int nullable) throws SQLException {
```
mattChiaravalloti, line 46: can we document our assumption that the schema will be simplified? specifically 1. that there will be no nested anyofs, and 2. we will not encounter an ANY in an anyof. These assumptions are baked into the code below but are not explicitly stated. as we get new team members, this may become lost knowledge.

mattChiaravalloti (changes requested):
A few thoughts for now! I figure we'll talk about Ryan's outstanding questions at standup today and I could give it another pass after that's sorted out.

src/main/java/com/mongodb/jdbc/MongoColumnInfo.java
```diff
@@ -0,0 +1,117 @@
+package com.mongodb.jdbc;
```
pmeredit, line 1: This used to be `Column`, more or less, or at least `MySQLColumnInfo` is.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -833,73 +833,6 @@ protected String getDataTypeBytesCase(String col) {
         return getTypeCase(col, typeBytes);
     }
 
-    protected MongoResultDoc getTypeInfoMetaDoc() {
```
pmeredit, line 836: These should have been moved into the `MySQLDatabaseMetaData` class before, we missed these.

src/main/java/com/mongodb/jdbc/MySQLColumnInfo.java
```diff
@@ -0,0 +1,175 @@
...
+                + ", bsonType="
+                + bsonType
+                + '}';
+    }
+
+    @Override
+    public boolean isPolymorphic() {
+        return false;
+    }
+
```
pmeredit, line 124: Since the POJO codec isn't going to call constructors, we need to compute these 3 pieces of data at runtime rather than at construction time as we do in `MongoSQL`. We're probably going to deprecate `MySQL` anyway, so I'm not too worried about it.

src/main/java/com/mongodb/jdbc/MySQLColumnInfo.java
```diff
@@ -0,0 +1,182 @@
...
+            throws SQLException {
+        this.database = database;
+        this.table = table;
+        this.tableAlias = tableAlias;
+        this.column = column;
+        this.columnAlias = columnAlias;
+        this.bsonType = bsonType;
+    }
+
+    public void init() throws SQLException {
```
pmeredit, line 51: The POJO codec is not going to call the constructor, so we use this init method to set data that isn't specified by the bson doc.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
```diff
@@ -0,0 +1,238 @@
...
+    // __bot.a
+    private static int DOUBLE_COL = 1;
+    // __bot.str
+    private static int STRING_COL = 2;
+    // foo.a
+    private static int ANY_OF_INT_STRING_COL = 3;
+    // foo.b
+    private static int INT_NULLABLE_COL = 4;
+    // foo.c
+    private static int INT_COL = 5;
```
mattChiaravalloti, line 31: Last time I reviewed this, I had to double check the `MongoSQLMock` to see why both of these return `ColumnNullable` for `isNullable`, since their names imply the former is nullable and the latter is not. This time, the same thing happened. Would you consider one or both of the following suggestions:
1. rename `INT_NULLABLE_COL` => `INT_OR_NULL_COL`
2. comment that `__bot.a`, `__bot.str`, `foo.c`, and `foo.d` _may_ be missing

I think either of those changes would help make this clear in just this file alone, without need to reference back to the mock class.

src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
```diff
@@ -5,48 +5,122 @@
...
+    @Override
+    public String getCatalogName(int column) throws SQLException {
+        return getColumnInfo(column).getDatabase();
+    }
+
+    @Override
+    public int getColumnType(int column) throws SQLException {
+        MongoColumnInfo ci = getColumnInfo(column);
+        return ci.getJDBCType();
+    }
```
mattChiaravalloti, line 97: [nit] Why are some methods implemented in one line, like `getCatalogName`, and some broken up into two lines, like `getColumnType`? Can we change them all to be one way or the other?

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
```diff
@@ -2,57 +2,106 @@
...
+    // in MongoSQLResultSet to retrieve data by label.
+    String getDatasource(String columnLabel) {
+        return columnLabels.get(columnLabel).datasource;
     }
 
-    @Override
-    public String getColumnName(int column) throws SQLException {
-        throw new SQLFeatureNotSupportedException("TODO");
+    private void assertObjectSchema(MongoJsonSchema schema) throws SQLException {
+        if (!schema.isObject() || schema.properties == null) {
```
mattChiaravalloti, line 46:
```suggestion
        if (!schema.isObject()) {
```
Should we also get rid of the `schema.properties == null` here and depend on `schema.isObject()` to capture the correct definition of "object schema"?

src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
pmeredit: Various stages of refactor ;) Sure.

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
pmeredit: I'll add a comment clarifying this, too

src/main/java/com/mongodb/jdbc/MongoColumnInfo.java
```diff
@@ -0,0 +1,126 @@
...
+        // "object"
+        // "objectId"
+        // "regex"
+        // "string"
+        // "symbol"
+        // "timestamp"
+        // "undefined"
+        //
+        // This function will not always throw an exception for an unknown type name. Type
+        // names returned from ADL are assumed correct.
```
rychipman, line 52: It seems like it would be much safer to change this function to match the exact values we expect. For example, if an "any" or "bson" somehow got passed into this function, it would silently return an incorrect value, which could be annoying to track down

src/main/java/com/mongodb/jdbc/MongoSQLColumnInfo.java
```diff
@@ -0,0 +1,171 @@
...
+            throws SQLException {
+        this.datasource = datasource;
+        this.field = field;
+        // All schemata except Any and AnyOf must have a bsonType.
+        if (schema.bsonType != null) {
+            bsonTypeName = schema.bsonType;
+            bsonType = MongoColumnInfo.getBsonTypeHelper(schema.bsonType);
+            jdbcType = getJDBCTypeForBsonType(this.bsonType);
+            isPolymorphic = false;
+            this.nullable = nullable;
```
rychipman, line 29: I think the argument to this function should be something like `boolean fieldRequired` instead -- IIUC from the implementation, it doesn't actually represent nullability in the same way that the `nullable` member does, since nullability is not just dependent on the field's required status, but also on the actual value (or, more specifically, the possible values allowed by the schema)

src/main/java/com/mongodb/jdbc/MongoSQLColumnInfo.java
```diff
@@ -0,0 +1,171 @@
...
+            return;
+        }
+        if (schema.isAny()) {
+            bsonTypeName = "bson";
+            jdbcType = Types.OTHER;
+            bsonType = BsonType.UNDEFINED;
```
rychipman, line 35: should this just be `null`? or is it not possible to have a `null` value for java enums?

rychipman (changes requested):
looks so much better after the refactor! just a few small things left for me

Issue #33, opened 2020-11-16
Passing database via connection string doesn't work


OiNutter:
With JDBC you should be able to specify the database via the connection string, not by having to pass it in the info argument. At the moment this will cause a NullPointerException as it fails the Preconditions check in MongoStatement.

rychipman:
@OiNutter could you provide an example of a connection string with database that you expect the JDBC driver to be able to handle?

mmpataki:
@rychipman I have submitted a PR, can you please review it

RaoulHofmann:
Is there a workaround for this? I need to use the driver with DataSourceProperties but can't pass in the database

osowskit:
+1 to this request. Ideally the `database` parameter can be passed in the URL parameters

jchemburkar:
This feature is now supported, as of [this](https://github.com/mongodb/mongo-jdbc-driver/commit/655fdf34c83ad00348bfdd7866599c9cdb0cf35e) commit. Similar to mongodb uris, this will come after the host / port. For example `jdbc:mongodb://<host>:<port>/<database>?authSource=<auth db>`. Please note that if the database is specified in the connection string and in the info, the latter will take precedence.

PR #49, opened 2021-10-13
ISSUE #33: Passing database via connection string doesn't work

mmpataki:
Database name specified in the connection string is not honored. With this change one can specify a db name as
   jdbc:mongodb://foo:foopass@localhost:27017/admin?database=mydatabase

### DIFF
--- a/src/test/java/com/mongodb/jdbc/MongoDriverTest.java
+++ b/src/test/java/com/mongodb/jdbc/MongoDriverTest.java
@@ -78,6 +78,7 @@ class MongoDriverTest {
     public void testDBNameInUrl() throws SQLException {
         Connection conn = DriverManager.getConnection(dbNameURL);
         conn.createStatement().close();
+        conn.close();
     }
 
     @Test


rychipman:
Hi @mmpataki, 

In the [documentation for the MongoDB URI|https://docs.mongodb.com/manual/reference/connection-string/], you'll see that the database name after the `/` is not actually a database to connect to, but rather the default auth database.

The JDBC driver allows users to specify the database to connect to via the `database` key in the `Properties` object passed into the `MongoConnection` constructor.

mmpataki:
@pmeredit @rychipman you guys had a discussion?

GavinRay97:
I am pretty sure you can do this, right?
In my app I am connecting like this:

- Connect to `sample` database, using `admin` database to check credentials

```properties
mongodb://root:password@localhost:27017/sample?authSource=admin
```

terakilobyte:
@mmpataki We've decided to match the way the drivers handle the default database in the URI, i.e. `mongodb://server/db?authSource=admin&otherProps`, see #122 .  cc @nbagnard and @pmeredit

terakilobyte:
Closing this issue as #122 resolves this. Users will be able to specify a default db in the URI on the next release.

PR #51, opened 2021-10-18
SQL-550: Add support for dialect property

mattChiaravalloti:
This PR adds support for the dialect property to the `MongoDriver` class according to the rules outlined [here in the design doc](https://docs.google.com/document/d/1kCladBO7LQXT_-Vyuv8UcOI8AMGBmpvgHKA5NFgnys8/edit#heading=h.rxmntstne1j8).

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -131,7 +131,8 @@ public class MongoDriver implements Driver {
 
     private MongoConnection createConnection(ConnectionString cs, Properties info) throws SQLException {
         // attempt to get DIALECT property, and default to "mysql" if none is present
-        switch (info.getProperty(DIALECT, MYSQL_DIALECT).toLowerCase()) {
+        String dialect = info.getProperty(DIALECT, MYSQL_DIALECT);
+        switch (dialect.toLowerCase()) {
             case MYSQL_DIALECT:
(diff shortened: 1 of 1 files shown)


src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -200,4 +202,32 @@ void testGetPropertyInfo() throws SQLException {
...
     }
+
+    @Test
+    void testDialectProperty() throws SQLException {
+        MongoDriver d = new MongoDriver();
+        Properties p = new Properties();
+        Connection c;
+
+        // dialect set to "mysql" results in MySQLConnection
+        p.setProperty("dialect", "MySQL");
```
bucaojit, line 213: Let's add one more test without this property set to confirm the default value of MySQL

bucaojit (changes requested):
Looks good, had a few small suggestions.

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
mattChiaravalloti: oh right! I had that in mind and then forgot to actually add it. The other tests are technically all covering that case since none of them set it, but I'll ensure here that without it set we return an `instanceof MySQLConnection`.

pmeredit (approved):
LGTM!

PR #52, opened 2021-10-18
SQL-551: Implement MongoSQLDatabaseMetaData methods that return empty result sets

mattChiaravalloti:
This PR implements the `MongoSQLDatabaseMetaData` methods that return empty result sets. For now, this is blocked on SQL-513 and SQL-535, but there is some progress to be reviewed. Specifically the `BsonExplicitCursor`, the methodology for returning an empty result set (a static member of that class), and the testing methodology for `{Mongo|My}SQLDatabaseMetaData`.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
+++ b/src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
@@ -182,12 +182,16 @@ public class MySQLDatabaseMetaData extends MongoDatabaseMetaData implements Data
 
         MySQLResultDoc metaDoc = new MySQLResultDoc();
         metaDoc.columns = new ArrayList<>();
-        metaDoc.columns.add(new MySQLColumnInfo("", "", "", PROCEDURE_CAT, PROCEDURE_CAT, "string"));
-        metaDoc.columns.add(new MySQLColumnInfo("", "", "", PROCEDURE_SCHEM, PROCEDURE_SCHEM, "string"));
-        metaDoc.columns.add(new MySQLColumnInfo("", "", "", PROCEDURE_NAME, PROCEDURE_NAME, "string"));
+        metaDoc.columns.add(
+                new MySQLColumnInfo("", "", "", PROCEDURE_CAT, PROCEDURE_CAT, "string"));
+        metaDoc.columns.add(
+                new MySQLColumnInfo("", "", "", PROCEDURE_SCHEM, PROCEDURE_SCHEM, "string"));
+        metaDoc.columns.add(
+                new MySQLColumnInfo("", "", "", PROCEDURE_NAME, PROCEDURE_NAME, "string"));
         metaDoc.columns.add(new MySQLColumnInfo("", "", "", REMARKS, REMARKS, "string"));
         metaDoc.columns.add(new MySQLColumnInfo("", "", "", PROCEDURE_TYPE, PROCEDURE_TYPE, "int"));
-        metaDoc.columns.add(new MySQLColumnInfo("", "", "", SPECIFIC_NAME, SPECIFIC_NAME, "string"));
+        metaDoc.columns.add(
+                new MySQLColumnInfo("", "", "", SPECIFIC_NAME, SPECIFIC_NAME, "string"));
 
         docs.add(metaDoc);
         return new MySQLResultSet(null, new MySQLExplicitCursor(docs), true);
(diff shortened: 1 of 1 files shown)


src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -363,30 +405,37 @@ void testGetFunctions() throws SQLException {
         rs = databaseMetaData.getFunctions(null, null, "%s%");
         assertEquals(countRows(rs), 0);
     }
+}
+
```
pmeredit, line 409: Does gradle correctly run these when they are in the same file? I was just using separate files to be safe.

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -6,25 +6,34 @@
...
-class MongoDatabaseMetaDataTest extends MySQLMock {
-    private static DatabaseMetaData databaseMetaData;
+abstract class MongoDatabaseMetaDataTest {
+    protected DatabaseMetaData databaseMetaData;
+
+    protected abstract DatabaseMetaData createDatabaseMetaData();
 
-    static {
-        databaseMetaData = new MySQLDatabaseMetaData(null);
+    @BeforeEach
```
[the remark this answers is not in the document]
nbagnard: This is odd, if you want to share the stacktrace offline, I can take a quick look. Here, it's probably not too much of a problem to initialize it for each test since it's not too costly. It could be hiding a potential issue with the metadata code though.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -39,7 +39,10 @@ public String getTimeDateFunctions() throws SQLException {
     @Override
     public ResultSet getProcedures(
             String catalog, String schemaPattern, String procedureNamePattern) throws SQLException {
-        throw new SQLFeatureNotSupportedException("TODO");
+        // TODO: create result set metadata that describes the columns of this result set
+        //  even though it is empty. This applies to all updated methods in this PR and is
+        //  blocked on SQL-513 and SQL-535.
```
bucaojit, line 44: Will this be implemented in this PR after SQL-513 and SQL-535 are merged?

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -6,30 +6,47 @@
...
         for (int i = 0; ; ++i) {
             if (!rs.next()) {
                 return i;
             }
         }
     }
 
-    // Most DatabaseMetaData tests require connection to an ADL
-    // cluster. These are just simple tests for things that return empty result sets.
+    void validateResultSet(ResultSet rs, int expectedNumRows, String[] expectedColumns)
```
bucaojit, line 32: Easier to read with the helper function.  That was a good idea @nbagnard

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: I plan to do that, yes. If you look at Patrick's 553 PR (#54), that's what I'm going for. At minimum, after 513 (#47) is merged, I'll be able to make the metadata docs and maybe that will be enough for now if 535 (#53) is still in pogress. Perhaps as part of 535 you could change the `return new MongoSQLResultSet(...)` lines in the DatabaseMetaData methods to whatever new constructor signature you make in that ticket. We'll have to see how everything plays out.

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -325,13 +287,93 @@ void testGetAttributes() throws SQLException {
         assertFalse(rs.next());
     }
 
+    @Test
```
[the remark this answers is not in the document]
nbagnard: Looking good :)

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -318,32 +252,92 @@ void testGetAttributes() throws SQLException {
                     "SCOPE_TABLE",
                     "SOURCE_DATA_TYPE",
                 };
-        for (int i = 0; i < columns.length; ++i) {
-            assertEquals(rsmd.getColumnName(i + 1), columns[i]);
-            assertEquals(rsmd.getColumnLabel(i + 1), columns[i]);
-        }
-        assertFalse(rs.next());
+
+        ResultSet rs = databaseMetaData.getAttributes(null, null, null, null);
```
nbagnard, line 256: I'm sorry, I totally missed this last time, but null is not always accepted and not always a synonym for no filter.
Here for example, typeNamePattern and attributeNamePattern are not supposed to support Null. It's probably Ok to say that Null = %, especially since this are empty ResultSet. But be mindful that it is not following the spec and tools will use a pattern, not Null, so it might be best to test with a pattern at least for the ones that are not empty to make sure that we behave correctly.

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -353,40 +347,37 @@ void testGetFunctions() throws SQLException {
                     "FUNCTION_TYPE",
                     "SPECIFIC_NAME",
                 };
-        for (int i = 0; i < columns.length; ++i) {
-            assertEquals(rsmd.getColumnName(i + 1), columns[i]);
-            assertEquals(rsmd.getColumnLabel(i + 1), columns[i]);
-        }
-        assertEquals(countRows(rs), 117);
+
+        ResultSet rs = databaseMetaData.getFunctions(null, null, null);
```
nbagnard, line 351: This is where I'd add databaseMetaData.getFunctions(null, null, "%") and make sure it also return 117 rows for example.

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
mattChiaravalloti: Will do. We'll need to be sure to keep this in mind for the integration test tickets as well, since those will all deal with non-empty result sets.

PR #53, opened 2021-10-19
SQL-535: Implement MongoSQLResultSet

bucaojit:
Jira: https://jira.mongodb.org/browse/SQL-535
Implemented the MongoSQLResultSet subclass functions
Added test MongoSQLResultSetTest.java based on MySQLResultSetTest
Enabled tests in MongoSQLStatementTest.java


src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -32,114 +37,670 @@ BsonDocument getCurrent() {
...
+                // This is what $convert does.
+                return o.asDateTime().getValue();
+            case DB_POINTER:
+                return handleLongConversionFailure(DB_POINTER);
+            case DECIMAL128:
+                return o.asDecimal128().longValue();
+            case DOCUMENT:
+                return handleLongConversionFailure(DOCUMENT);
+            case DOUBLE:
+                return (long) o.asDouble().getValue();
```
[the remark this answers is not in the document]
bucaojit: From discussion and standup: 
mongodb returns error if double is less than min long or greater than max long value.
Decided to handle it the way Java would cast to long.  It has the value at this point, not error.

src/main/java/com/mongodb/jdbc/MongoSQLStatement.java
```diff
@@ -31,10 +34,30 @@ public ResultSet executeQuery(String sql) throws SQLException {
                 iterable = iterable.batchSize(fetchSize);
             }
 
-            resultSet = new MongoSQLResultSet(this, iterable.cursor());
+            MongoIterable<MongoJsonSchemaResult> schemaIterable =
+                    currentDB
+                            .withCodecRegistry(MongoDriver.registry)
+                            .aggregate(
+                                    Collections.singletonList(schemaStage),
```
mattChiaravalloti, line 41: Have you run this against an adl yet? I don't think this should be an `aggregate`. the `sqlGetResultSchema` command is not an aggregation stage. You should use `runCommand` instead (specifically this version: `<TResult> TResult runCommand(Bson var1, Class<TResult> var2)`).

src/main/java/com/mongodb/jdbc/MongoSQLStatement.java
```diff
@@ -31,10 +34,30 @@ public ResultSet executeQuery(String sql) throws SQLException {
...
+                                    MongoJsonSchemaResult.class)
+                            .maxTime(maxQuerySec, TimeUnit.SECONDS);
+            MongoJsonSchemaResult schemaResult = schemaIterable.cursor().next();
+            if (schemaResult.ok != 1) {
+                throw new SQLException("Invalid schema result from server");
```
mattChiaravalloti, line 46: We may want to ask the ADL team if they return an `err` or `error` field or something that we could pass along to users to provide more details.

mattChiaravalloti: @henretty This is where we're calling `sqlGetResultSchema`. 2 questions:
1. Is it possible for this to return a result with `ok` != 1?
2. If so, does the result have any error information we can/should share with JDBC users?

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -187,7 +247,56 @@ protected static MongoJsonSchema generateMongoJsonSchema() {
...
+                "$undefined":true
+            },
+             "foo.null": null
+             "foo.vec":[
+                1,
+                2,
+                3
+            ],
+            "__bot.a":1.2,
+            "__bot.str":"a"
```
mattChiaravalloti, line 265:
```suggestion
            "__bot.a":1.2,
            "__bot.binary": <binary data>
            "__bot.str":"a"
```
update comment to include new field. I don't think we need to write out the extended json representation of the data; the fact that it is binary is the important part, not the specific value.

mattChiaravalloti (changes requested):
Just a few more notes 🤏

src/main/java/com/mongodb/jdbc/MongoSQLStatement.java
bucaojit: I will update to using runCommand().  Thanks for the suggestion of running against ADL, I was able to run the command and it reached ADL and returned an error.  This made it clearer how the errors would be propagated to the user.

src/main/java/com/mongodb/jdbc/MongoSQLStatement.java
bucaojit: When running the runCommand({sqlGetResultSchema:1, ...}) against ADL it hit an error that threw an exception at the runCommand() line, prior to this check of schemaResult.ok !=1.  There is a check done that throws the error if the response for ok is not 1, [here.](https://github.com/mongodb/mongo-java-driver/blob/53b0ff156a33d883ec6c78a64d2c9c3d14e9218c/driver-core/src/main/com/mongodb/internal/connection/InternalStreamConnection.java#L417)

So it seems that by the time we get to line 45, it has already been checked making it no longer necessary.  Does that sound correct?

mattChiaravalloti: Ah that sounds good to me. It's weird they throw an exception for that considering mongo has pretty rich/detailed error codes, but I suppose exceptions are the java-way. Given that check, there is no need for us to do the `ok != 1` check ourselves.

Do you think we should be catching and handling that exception, though?

bucaojit: It throws a runtime exception, which leads me to think the error that is thrown is meant to stop execution and show the error.  In terms of handling there isn't much to do if the schema wasn't returned, besides retry if it were due to a timeout perhaps.  
I'm leaning towards leaving it for now.

mattChiaravalloti: Makes sense. Sounds good to me 👍

mattChiaravalloti (approved):
lgtm 💯 💯

src/test/java/com/mongodb/jdbc/MongoSQLStatementTest.java
```diff
@@ -62,75 +71,71 @@ void testNoop(MongoSQLConnectionTest.TestInterface ti) throws SQLException {
...
-        //        assertThrows(
-        //                SQLException.class,
-        //                new Executable() {
-        //                    @Override
-        //                    public void execute() throws Throwable {
-        //                        rs.getInt(1);
-        //                    }
-        //                });
-        //        assertTrue(rs.isLast());
+        AtomicInteger rowCnt = new AtomicInteger();
```
nbagnard, line 74: Why does this have to be an AtomicInteger? I don't see where is the concurrency coming from, do you know?

src/test/java/com/mongodb/jdbc/MongoSQLStatementTest.java
```diff
@@ -62,75 +71,71 @@ void testNoop(MongoSQLConnectionTest.TestInterface ti) throws SQLException {
...
+        when(mongoDatabase.runCommand(any(), eq(MongoJsonSchemaResult.class)))
+                .thenReturn(generateSchema());
+        when(mongoCursor.hasNext()).thenAnswer(invocation -> rowCnt.get() < 1);
+
+        when(mongoCursor.next())
+                .thenAnswer(
+                        invocation -> {
+                            if (rowCnt.incrementAndGet() == 1) {
```
nbagnard, line 118: I am not sure I follow what is going on here. The if will increment the row count, then if rowCnt == 1 it will call generateRow() and if not, it will also call generateRow. Is this a left over from the previous version of the test?

src/main/java/com/mongodb/jdbc/MongoSQLStatement.java
```diff
@@ -31,10 +34,24 @@ public ResultSet executeQuery(String sql) throws SQLException {
                 iterable = iterable.batchSize(fetchSize);
             }
 
-            resultSet = new MongoSQLResultSet(this, iterable.cursor());
+            MongoJsonSchemaResult schemaResult =
+                    currentDB
+                            .withCodecRegistry(MongoDriver.registry)
+                            .runCommand(getSchemaCmd, MongoJsonSchemaResult.class);
+
+            MongoJsonSchema schema = schemaResult.schema.jsonSchema;
```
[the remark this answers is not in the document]
mattChiaravalloti: If the command fails, the java driver throws an exception. If it succeeds, the `schemaResult` object will definitely be non-null. The `schema` field should also never be null ([docs](https://docs.mongodb.com/datalake/reference/cli/sql/sqlgetschema/#output)). 

The `jsonSchema` field _may_ be null if the collection has no schema. I'm not sure what the java driver does when unmarshaling an object with missing fields.

If the collection has no schema, the response is
```
{ "ok" : 1, "metadata" : { }, "schema" : { } }
```
So the `schema` field would probably be initialized with default values: `version = 0`, `jsonSchema = null`? Worth investigating how the java driver handles this kind of case (we can test this without using the `sqlGetSchema` command). If the `jsonSchema` could be null, we should definitely handle that.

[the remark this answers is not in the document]
pmeredit: I'll test this, since I already have a framework for it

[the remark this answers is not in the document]
pmeredit: It maintains the `jsonSchema` field but with a null value.  Which, in retrospect, should be obvious, otherwise it would fail to deserialize to a `JsonSchemaResult` haha.

src/test/java/com/mongodb/jdbc/MongoSQLStatementTest.java
pmeredit: I'm not sure why it needs to be `Atomic` either, but I assume this decision was made because something failed. Not really worried in a test file.

src/test/java/com/mongodb/jdbc/MongoSQLStatementTest.java
pmeredit: I think this is because in the `MySQL` implementation we sent the `metadata` in the first document of the cursor, but now the metadata is coming from a separate call to ADL. I think this should be changed!

pmeredit (approved):
LGTM, but let's take out the part in the test where we are checking if the `rowCnt` is `1`

src/test/java/com/mongodb/jdbc/MongoSQLStatementTest.java
bucaojit: That's right, it came from the `MySQL` test where the first one would return a `generateMetadataDoc()`, and was left in on accident and still worked since the behavior was the same.  I fixed it now.

PR #54, opened 2021-10-21
SQL-553: Implement Function methods for MongoSQLDatabaseMetadata

pmeredit:
This does not have testing and is not complete until SQL-513, SQL-535, and SQL-551 are merged, but it's worth looking at.


src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
```diff
@@ -907,6 +907,80 @@ public ResultSet getClientInfoProperties() throws SQLException {
...
+        metaDoc.columns.add(new Column("", "", "", FUNCTION_NAME, FUNCTION_NAME, "string"));
+        metaDoc.columns.add(new Column("", "", "", REMARKS, REMARKS, "string"));
+        metaDoc.columns.add(new Column("", "", "", FUNCTION_TYPE, FUNCTION_TYPE, "int"));
+        metaDoc.columns.add(new Column("", "", "", SPECIFIC_NAME, SPECIFIC_NAME, "string"));
+        return metaDoc;
+    }
+
+    private MongoResultDoc getFunctionValuesDoc(String functionName, String remarks) {
+        MongoResultDoc doc = new MongoResultDoc();
+        doc.values = new ArrayList<>(5);
```
mattChiaravalloti, line 924:
```suggestion
        doc.values = new ArrayList<>(6);
```
same here?

src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
```diff
@@ -907,6 +907,80 @@ public ResultSet getClientInfoProperties() throws SQLException {
         return new MySQLResultSet(null, new MongoExplicitCursor(docs), true);
     }
 
+    private MongoResultDoc getFunctionMetaDoc() {
+        MongoResultDoc metaDoc = new MongoResultDoc();
+        metaDoc.columns = new ArrayList<>(5);
```
mattChiaravalloti, line 912:
```suggestion
        metaDoc.columns = new ArrayList<>(6);
```
off-by-one?

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -190,10 +208,142 @@ public ResultSet getClientInfoProperties() throws SQLException {
         throw new SQLFeatureNotSupportedException("TODO");
     }
 
+    private MongoJsonSchema getFunctionJsonSchema() {
+        MongoJsonSchema schema = new MongoJsonSchema();
+        schema.bsonType = "object";
+        schema.additionalProperties = false;
+        schema.required = new HashSet<String>();
+        schema.properties = new HashMap<String, MongoJsonSchema>();
```
mattChiaravalloti, line 216: Maybe on the `MongoJsonSchema` class we should have some additional constructor or some initialization method or some other static "constructor" method that directly returns to us this initialized, empty-object schema?

Something like
```
public MongoJsonSchema(String bsonType) {
    // if object: initialize like an object schema, else initialize with just bsonType set?
    ...
}

// or
public MongoJsonSchema(String bsonType, boolean additionalProperties) {
    ...
}

// alternatively
public static MongoJsonSchema makeEmptyObjectSchema {
    MongoJsonSchema s = new MongoJsonSchema();
    // initialize
    ...
}
```

We may be doing this type of initialization a lot.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -190,10 +208,142 @@ public ResultSet getClientInfoProperties() throws SQLException {
...
     @Override
     public ResultSet getFunctions(String catalog, String schemaPattern, String functionNamePattern)
             throws SQLException {
-        throw new SQLFeatureNotSupportedException("TODO");
+        ArrayList<BsonDocument> docs = new ArrayList<>(MongoFunction.mongoSQLFunctionNames.length);
+        MongoJsonSchema schema = getFunctionJsonSchema();
+
+        Pattern functionPatternRE = null;
+        if (functionNamePattern != null) {
+            functionPatternRE = Pattern.compile(functionNamePattern.replaceAll("%", ".*"));
```
mattChiaravalloti, line 255: I'm working with patterns in my PR too and wonder if this is sufficient. Don't we also need to replace `_` with `.`, and should we escape other java-style regex characters that otherwise likely aren't escaped in sql-style patterns?

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -190,10 +208,142 @@ public ResultSet getClientInfoProperties() throws SQLException {
...
+            String remarks = func.comment;
+            if (functionPatternRE != null && !functionPatternRE.matcher(functionName).matches()) {
+                continue;
+            }
+            BsonDocument doc = getFunctionValuesDoc(func.name, func.comment);
+            docs.add(doc);
+        }
+
+        return new MongoSQLResultSet(null, null);
+        // return MongoSQLResultSet(null, new MongoSQLExplicitCursor(docs), schema);
```
mattChiaravalloti, line 269: What is your preference for the name: my current PR for 551 has `BsonExplicitCursor` since it operates over `BsonDocument`s. Here, you have `MongoSQLExplicitCursor` which also makes sense since it will be used by the MongoSQL implementations of the driver interfaces. I'm leaning a bit more `Bson` than `MongoSQL` since this cursor type doesn't strictly need to be used with just the MongoSQL code, but will go with whichever you prefer.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
pmeredit: I like `makeEmptyObjectSchema`

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
pmeredit: Probably, just copied this from what we did in `MySQL`. You think we want to escape any escape sequence?

pmeredit: I think this gets tricky. Let's support `_` and make another ticket for possible regex characters.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
pmeredit: Oh, I'm fine with `Bson`, I just forgot to recheck your PR!

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -190,10 +208,142 @@ public ResultSet getClientInfoProperties() throws SQLException {
...
+    }
+
+    private MongoJsonSchema getFunctionColumnJsonSchema() {
+        MongoJsonSchema schema = new MongoJsonSchema();
+        schema.bsonType = "object";
+        schema.additionalProperties = false;
+        schema.required = new HashSet<String>();
+        schema.properties = new HashMap<String, MongoJsonSchema>();
+
+        schema.required.add(FUNCTION_CAT);
```
pmeredit, line 279: @mattChiaravalloti Maybe it makes sense to add a method that adds required fields with types, too?

src/main/java/com/mongodb/jdbc/MySQLDatabaseMetaData.java
pmeredit: Ah yea, this has been "wrong" for over a year!

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -4,8 +4,26 @@
...
+import org.bson.BsonNull;
+import org.bson.BsonString;
+import org.bson.BsonValue;
 
 public class MongoSQLDatabaseMetaData extends MongoDatabaseMetaData implements DatabaseMetaData {
+    private static final MongoJsonSchema stringSchema;
+    private static final MongoJsonSchema intSchema;
+
+    static {
+        stringSchema = new MongoJsonSchema();
```
bucaojit, line 22: This is a good idea, I was instantiating these individually in the methods.  I'll switch to using these variables.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -190,10 +208,142 @@ public ResultSet getClientInfoProperties() throws SQLException {
...
+            BsonDocument doc = getFunctionValuesDoc(func.name, func.comment);
+            docs.add(doc);
+        }
+
+        return new MongoSQLResultSet(null, null);
+        // return MongoSQLResultSet(null, new MongoSQLExplicitCursor(docs), schema);
+    }
+
+    private MongoJsonSchema getFunctionColumnJsonSchema() {
+        MongoJsonSchema schema = new MongoJsonSchema();
```
bucaojit, line 273: Schema creation looks like it has some similar initial lines and patterns.  
I'm wondering if a generateSchema() helper function could be written that takes in the desired columns and returns a schema.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -190,10 +208,142 @@ public ResultSet getClientInfoProperties() throws SQLException {
...
+        schema.properties.put(FUNCTION_CAT, stringSchema);
+        schema.properties.put(FUNCTION_SCHEM, stringSchema);
+        schema.properties.put(FUNCTION_NAME, stringSchema);
+        schema.properties.put(REMARKS, stringSchema);
+        schema.properties.put(FUNCTION_TYPE, intSchema);
+        schema.properties.put(SPECIFIC_NAME, stringSchema);
+        return schema;
+    }
+
+    private BsonDocument getFunctionValuesDoc(String functionName, String remarks) {
```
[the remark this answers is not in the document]
pmeredit: I think that's extra complexity that doesn't really help in this case. Now, if there were some base class functionality that used it, it would make sense, but right now it's only used in the base class. Using the same name is just to make it more clear to someone familiar with the other class.

PR #55, opened 2021-10-21
SQL-554: Implement MongoSQLDatabaseMetadata, MongoSQL dialect meta me…

bucaojit:
…thods

JIRA: https://jira.mongodb.org/browse/SQL-554
Implemented MongoSQLDatabaseMetadata, MongoSQL dialect meta methods:
getClientInfoProperties
getSQLKeywords, 
getTypeInfo, 
getTableTypes

Does not have testing included.  Similar to the PR for SQL-553 , depends on SQL-513, SQL-535, and SQL-551


src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -54,7 +94,37 @@ public ResultSet getProcedureColumns(
 
     @Override
     public ResultSet getTableTypes() throws SQLException {
-        throw new SQLFeatureNotSupportedException("TODO");
+        ArrayList<BsonDocument> docs = new ArrayList<>();
+
+        MongoJsonSchema schema = new MongoJsonSchema();
```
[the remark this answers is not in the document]
bucaojit: Sounds good, using the new MongoJsonSchema methods now.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -13,9 +24,42 @@ public MongoSQLDatabaseMetaData(MongoConnection conn) {
         super(conn);
     }
 
+    private BsonDocument createBottomBson(BsonElement... elements) {
+        BsonDocument bot = new BsonDocument(Arrays.asList(elements));
+        return new BsonDocument(BOT_NAME, bot);
+    }
+
     @Override
     public String getSQLKeywords() throws SQLException {
```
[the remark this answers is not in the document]
bucaojit: I updated the keywords list based on the linked [keywords document](https://developer.mimer.com/wp-content/uploads/standard-sql-reserved-words-summary.pdf)

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -382,7 +1116,19 @@ public ResultSet getSchemas(String catalog, String schemaPattern) throws SQLExce
...
+        schema.addRequiredScalarKeys(
+                new Pair<>(NAME, BSON_STRING_TYPE_NAME),
+                new Pair<>(MAX_LEN, BSON_STRING_TYPE_NAME),
+                new Pair<>(DEFAULT_VALUE, BSON_STRING_TYPE_NAME),
+                new Pair<>(DESCRIPTION, BSON_STRING_TYPE_NAME));
+
+        // All fields in this result set are nested under the bottom namespace.
+        MongoJsonSchema botSchema = MongoJsonSchema.createEmptyObjectSchema();
+        botSchema.properties.put(BOT_NAME, schema);
+        return new MongoSQLResultSet(null, BsonExplicitCursor.EMPTY_CURSOR, botSchema);
```
bucaojit, line 1131: @nbagnard informed me that the clientInfoProperties are not for connection settings.  Postgres for example only returns applicationName for the session, mysql returns empty set.  So removed the connection settings and passing the empty cursor instead.  
I will keep the mysql clientInfoProperties as is to not change functionality.

pmeredit (approved):
LGTM once the other suggestions are met. Oh, but do make sure that the current comment for keywords matches what we have now.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
     public <T> T unwrap(Class<T> iface) throws SQLException {
         return (T) this;
     }
+    // Helper functions for getTypeInfo
+    protected Integer getPrecision(String typeName) throws SQLException {
```
mattChiaravalloti, line 1143: [organizational nit] My initial vision for this these helper methods was to make them `static`, possibly even part of a totally separate class and `public static`. What if we create a `BsonTypeInfo` class (or some other relevant name) that has all of these as `static` methods? We can also move the `BSON_<type>_TYPE_NAME` constants to that class.

This feels like a big chunk of related fields and methods that could be factored out separately. They are pretty specific to `MongoDatabaseMetaData` subtypes, but still it could be nice to have a larger number of smaller classes than continuing to build these large monolithic classes. What do you think?

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+                return 34;
+            case BSON_DOUBLE_TYPE_NAME:
+                return 15;
+            case BSON_LONG_TYPE_NAME:
+                return 19;
+        }
+        throw new SQLException("unknown bson typeName: " + typeName);
+    }
+
+    protected Boolean getCaseSensitivity(String typeName) throws SQLException {
```
mattChiaravalloti, line 1178:
```suggestion
    protected boolean getCaseSensitivity(String typeName) throws SQLException {
```
Why return `Integer` and `Boolean` from these methods instead of their primitive counterparts?

mattChiaravalloti (changes requested):
This is looking very good. We're almost there 🏁  I just have two questions about the new type info helper methods.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+                return false;
+        }
+        throw new SQLException("unknown bson typeName: " + typeName);
+    }
+
+    protected Boolean getFixedPrecScale(String typeName) throws SQLException {
+        switch (typeName) {
+            case BSON_INT_TYPE_NAME:
+            case BSON_LONG_TYPE_NAME:
+                return true;
```
nbagnard, line 1211: I thought that fixedPrecScale was for currency/money values, which are generally related to Decimal/numeric with a fixed precision and scale. I don't know if reporting Int and Long will cause issues or not, but I don't think they are great type for representing money values either. Maybe we should just return false for every type?

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+            case BSON_BSON_TYPE_NAME:
+                return false;
+        }
+        throw new SQLException("unknown bson typeName: " + typeName);
+    }
+
+    protected Integer getMinScale(String typeName) throws SQLException {
+        switch (typeName) {
+            case BSON_DECIMAL_TYPE_NAME:
+                return 34;
```
nbagnard, line 1239: Shouldn't min scale be 0 for Decimal and Double too? For Decimal, I'm pretty sure that you can have a scale < 34 and 0 is how you would represent an int/long value.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+            case BSON_DOUBLE_TYPE_NAME:
+                return 15;
```
nbagnard, line 1241: For Double, it's a floating point decimal so I am not totally sure because it does not really apply, but certainly it can be < 15. I think 0 also makes sense.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+        }
+        throw new SQLException("unknown bson typeName: " + typeName);
+    }
+
+    protected Integer getMaxScale(String typeName) throws SQLException {
+        switch (typeName) {
+            case BSON_DECIMAL_TYPE_NAME:
+                return 34;
+            case BSON_DOUBLE_TYPE_NAME:
+                return 15;
```
nbagnard, line 1271: @pmeredit , @mattChiaravalloti : is this true? 
Also, I'm still not sure if we should report a max scale for a floating point decimal.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+
+            case BSON_BINDATA_TYPE_NAME:
+            case BSON_STRING_TYPE_NAME:
+            case BSON_INT_TYPE_NAME:
+            case BSON_BOOL_TYPE_NAME:
+            case BSON_DATE_TYPE_NAME:
```
nbagnard, line 1277: Based on this page, I think the max scale for BSON Date is 3, not 0 : https://docs.mongodb.com/manual/reference/mongodb-extended-json/#mongodb-bsontype-Date
Maybe @pmeredit or @mattChiaravalloti you can confirm/infirm?

mattChiaravalloti: Agreed that the docs do seem to say that!

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
nbagnard: Got the answer from conversation in standup : BSON_DECIMAL scale is always 34.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+            case BSON_TIMESTAMP_TYPE_NAME:
+            case BSON_UNDEFINED_TYPE_NAME:
+            case BSON_BSON_TYPE_NAME:
+                return 0;
+            case BSON_JAVASCRIPT_TYPE_NAME:
+            case BSON_JAVASCRIPTWITHSCOPE_TYPE_NAME:
+            case BSON_REGEX_TYPE_NAME:
+            case BSON_SYMBOL_TYPE_NAME:
+            case BSON_ARRAY_TYPE_NAME:
+                return unknownLength;
```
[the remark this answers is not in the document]
mattChiaravalloti: I agree that `null` is the right move here, given the documentation's description.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: You're right, primitives are more appropriate.  I'll make the change.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: Okay that makes sense.  I moved the helper functions and the `BSON_<type>_TYPE_NAME` to a new class: `BsonTypeInfo`

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: Checked with @nbagnard and good to resolve.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1103,4 +1139,188 @@ public boolean isWrapperFor(Class<?> iface) throws SQLException {
...
+            case BSON_BINDATA_TYPE_NAME:
+            case BSON_STRING_TYPE_NAME:
+            case BSON_OBJECT_TYPE_NAME:
+            case BSON_DBPOINTER_TYPE_NAME:
+            case BSON_MAXKEY_TYPE_NAME:
+            case BSON_MINKEY_TYPE_NAME:
+            case BSON_TIMESTAMP_TYPE_NAME:
+            case BSON_UNDEFINED_TYPE_NAME:
+            case BSON_BSON_TYPE_NAME:
+                return 0;
```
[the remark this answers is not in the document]
bucaojit: Okay, setting these `0` and `unknownLength` to `null` as mentioned in an earlier comment.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -84,11 +102,29 @@
     protected static final String SCOPE_SCHEMA = "SCOPE_SCHEMA";
     protected static final String SCOPE_TABLE = "SCOPE_TABLE";
     protected static final String SOURCE_DATA_TYPE = "SOURCE_DATA_TYPE";
-
     protected static final String COLUMN_USAGE = "COLUMN_USAGE";
 
+    protected static final String BSON_BINDATA_TYPE_NAME = "binData";
```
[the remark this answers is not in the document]
bucaojit: I created a class that will help with this: `BsonTypeInfo`.  I'll create a refactoring ticket to make better use of `BsonTypeInfo` in other classes besides MongoDatabaseMetadata.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: Okay, I am fine with this.  I'll make the change.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: I switched getPrecision() back to Integer in order to return `null`.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
mattChiaravalloti: awesome, thanks

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -0,0 +1,183 @@
+package com.mongodb.jdbc;
+
+import java.sql.SQLException;
+
+public class BsonTypeInfo {
+    public static final String BSON_BINDATA_TYPE_NAME = "binData";
```
mattChiaravalloti, line 6:
```suggestion
    public static final String BINDATA_TYPE_NAME = "binData";
```
[ultra-nit] Totally up to you (or anyone else who has a strong opinion): we could remove the BSON prefix from all of these since they'll be accessed using the class name (`BsonTypeInfo.XYZ_TYPE_NAME`).

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -266,9 +324,929 @@ public ResultSet getPrimaryKeys(String catalog, String schema, String table)
...
-        throw new SQLFeatureNotSupportedException("TODO");
+        BsonValue n = new BsonNull();
+        ArrayList<BsonDocument> docs = new ArrayList<>();
+        MongoJsonSchema schema = getTypeInfoJsonSchema();
+
+        docs.add(
+                createBottomBson(
+                        new BsonElement(
+                                TYPE_NAME, new BsonString(BsonTypeInfo.BSON_BINDATA_TYPE_NAME)),
+                        new BsonElement(DATA_TYPE, new BsonInt32(Types.BINARY)),
```
mattChiaravalloti, line 369: should we add a `static getJDBCType` method to `BsonTypeInfo` to get this for each type? I know I also need this info in at least one other place

mattChiaravalloti (approved):
Looks good to me! 🦅 

I have two small suggestions/requests that are up to you. The helper method I suggested would be nice to see but I'll leave that up to you. No need to retag when/if you make changes!

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
bucaojit: I think we can defer this to the refactor change Natacha is suggesting with making BsonTypeInfo into enums.  The enum would have the TYPE_NAME, DATA_TYPE specified then we can call `enumType.getJdbcType()` such as:
```
       for (BsonTypeInfo currType : BsonTypeInfo.values()) {
           if (currType.isMySqlSupported) {
               docs.add(
                       MySQLDatabaseMetaData.getTypeInfoValuesDoc(
                               currType.getTypeName(),
                               currType.getJdbcType(), //dataType
                               currType.getPrecision(), //precision
                               ... ));
           }
       }
```

PR #57, opened 2021-10-21
SQL-513: Add MongoJsonSchema and MongoJsonSchemaResult

pmeredit:
Check me on the `MongoJsonSchemaResult`, the goal there is for that to be the deserialization type for the return type of `sqlGetResultSchema`

We aren't going to return a superdocument with `$jsonSchema`, right?

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -129,7 +129,8 @@ public class MongoDriver implements Driver {
         return createConnection(p.left(), info);
     }
 
-    private MongoConnection createConnection(ConnectionString cs, Properties info) throws SQLException {
+    private MongoConnection createConnection(ConnectionString cs, Properties info)
+            throws SQLException {
         // attempt to get DIALECT property, and default to "mysql" if none is present
         String dialect = info.getProperty(DIALECT, MYSQL_DIALECT);
         switch (dialect.toLowerCase()) {
@@ -139,9 +140,11 @@ public class MongoDriver implements Driver {
             case MONGOSQL_DIALECT:
                 if (info.containsKey(CONVERSION_MODE)) {
                     throw new SQLClientInfoException(
-                            String.format("must not set '%s' if '%s' is '%s'",
+                            String.format(
+                                    "must not set '%s' if '%s' is '%s'",
                                     CONVERSION_MODE, DIALECT, MONGOSQL_DIALECT),
-                            Collections.singletonMap(CONVERSION_MODE, ClientInfoStatus.REASON_VALUE_INVALID));
+                            Collections.singletonMap(
+                                    CONVERSION_MODE, ClientInfoStatus.REASON_VALUE_INVALID));
                 }
                 return new MongoSQLConnection(cs, info.getProperty(DATABASE));
             default:
--- a/src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
+++ b/src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
@@ -46,10 +46,10 @@ public class MongoJsonSchema {
     @SafeVarargs
     public final void addRequiredScalarKeys(Pair<String, String>... scalarProperties) {
         if (properties == null) {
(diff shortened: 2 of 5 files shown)


src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -139,9 +140,11 @@ private MongoConnection createConnection(ConnectionString cs, Properties info) t
             case MONGOSQL_DIALECT:
                 if (info.containsKey(CONVERSION_MODE)) {
                     throw new SQLClientInfoException(
-                            String.format("must not set '%s' if '%s' is '%s'",
```
pmeredit, line 142: These are being done by spotless. Not sure how we are passing the spotless test without running spotless

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -0,0 +1,67 @@
...
+        return ret;
+    }
+
+    public static MongoJsonSchema createScalarSchema(String type) {
+        MongoJsonSchema ret = new MongoJsonSchema();
+        ret.bsonType = type;
+        return ret;
+    }
+
+    @SafeVarargs
```
pmeredit, line 34: This will make the creation of `MongoJsonSchema` in the `MongoSQLDatabaseMetaData` methods much cleaner.

src/main/java/com/mongodb/jdbc/MongoJsonSchemaResult.java
```diff
@@ -0,0 +1,6 @@
+package com.mongodb.jdbc;
+
+public class MongoJsonSchemaResult {
```
mattChiaravalloti, line 3: This is meant to be the result of `sqlGetResultSchema`, right? If so, the `schema` field return there is actually going to be formatted the same way as the `schema` field from `sqlGetSchema`: https://docs.mongodb.com/datalake/reference/cli/sql/sqlgetschema/#output
```
{
    ...,
    "schema": {
        "version": 1,
        "jsonSchema": ...
    }
}
```

So we need to account for that one level of nesting

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -0,0 +1,67 @@
...
+    public final void addRequiredScalarKeys(Pair<String, String>... values) {
```
mattChiaravalloti, line 35: Can you add a comment to this method explaining what the arguments are? It is clear enough from reading the code, but I wonder if that's because I have more familiarity with the codebase than a newcomer might.

I don't mind if the comment isn't javadoc style, but I'm not sure what java convention is in that department. Maybe a javadoc style comment would be best.

src/main/java/com/mongodb/jdbc/MongoJsonSchemaResult.java
mattChiaravalloti: Come to think of it, I also need to deserialize the result of `sqlGetSchema` which has the same fields, `ok` and `schema`, but also one additional field, `metadata`. Will I be able to use this class for that as well?

mattChiaravalloti: looks like an object with one field, `description`, which is a string. We probably don't need to actually parse it all the way though. Leaving `metadata` as `BsonDocument` should be fine.

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -0,0 +1,79 @@
...
+     *     bsonType
+     * @return void
+     */
+    @SafeVarargs
+    public final void addRequiredScalarKeys(Pair<String, String>... scalarProperties) {
+        if (properties == null) {
+            properties = new HashMap<String, MongoJsonSchema>();
+        }
+        if (required == null) {
+            required = new HashSet<String>();
```
mattChiaravalloti, line 52:
```suggestion
            properties = new HashMap<>();
        }
        if (required == null) {
            required = new HashSet<>();
```

PR #58, opened 2021-10-22
SQL-557: Update functions categorization

nbagnard:
Update the functions categorization based on the JDBC spec.
Started refactoring to split Mysql and MongoSql specific code since the supported functions are different.


src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -357,11 +357,11 @@ void testGetFunctions() throws SQLException {
             assertEquals(rsmd.getColumnName(i + 1), columns[i]);
             assertEquals(rsmd.getColumnLabel(i + 1), columns[i]);
         }
-        assertEquals(countRows(rs), 117);
+        assertEquals(119, countRows(rs));
```
pmeredit: Thanks for moving expected first!

pmeredit (approved):
nit: I would prefer if we named the classes `MySQLFunctions` and `MongoSQLFunctions`, but otherwise looks great!

nbagnard:
> nit: I would prefer if we named the classes `MySQLFunctions` and `MongoSQLFunctions`, but otherwise looks great!

Sure, I used to always use the parent class name as a suffix, but I agree it doesn't bring much here. Just makes the class names very long.

src/main/java/com/mongodb/jdbc/MongoSQLFunctions.java
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    public static MongoSQLFunctions getInstance() {
+        if (null == instance) {
+            instance =
+                    new MongoSQLFunctions(
+                            new MongoFunction[] {
+                                new MongoFunction(
+                                        "BIT_LENGTH",
+                                        "long",
```
bucaojit, line 17: I'd like to see these repeated strings ie `BIT_LENGTH` and `long` as `protected static final String` in the abstract class like how we have it in `MongoDatabaseMetaData.java` and used for `MongoSQLFunctions.java` and `MySQLFunctions.java`.

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -400,4 +404,12 @@ void testGetSchemas() throws SQLException {
         rs = databaseMetaData.getSchemas(null, null);
         validateResultSet(rs, 0, columns);
     }
+
+    @Test
+    @Override
+    void testGetFunctions() throws SQLException {
+        testGetFunctionsHelper("%", 16);
```
bucaojit, line 411: This is nice and concise.  I see it is testing the different function name patterns, but not super clear as to what it's testing.  Please add a brief comment on what is being tested.

src/main/java/com/mongodb/jdbc/MySQLFunctions.java
```diff
@@ -0,0 +1,730 @@
+// This is generated code. To regenerate go to the resources directory:
```
bucaojit, line 1: This comment seems old, is this file still generated code?

src/main/java/com/mongodb/jdbc/MongoFunctions.java
```diff
@@ -0,0 +1,110 @@
...
+    public String systemFunctionsString;
+
+    protected MongoFunctions(MongoFunction[] functions) {
+        this.functions = functions;
+        initCategorizedFunctionsList();
+    }
+
+    // Build the list of numeric, string, dateTime and system functions.
+    private void initCategorizedFunctionsList() {
+        StringBuilder numericFunctionBuilder = new StringBuilder();
```
bucaojit, line 55: nit: naming goes from numericFunction`s`String to numericFunctionBuilder without the 's'.  Suggest adding the 's' for consistency.

src/main/java/com/mongodb/jdbc/MongoFunctions.java
```diff
@@ -0,0 +1,110 @@
...
+                default:
+                    {
+                        currBuilder = null;
+                        // Nothing to do
+                        break;
+                    }
+            }
+
+            if (null != currBuilder) {
+                if (0 < currBuilder.length()) {
```
bucaojit, line 99: nit:  Might by stylistic, I'm used to seeing `if (currBuilder.length() > 0) {`.  Makes more sense to me, fine with either way.

bucaojit (changes requested):
Nice work.  Some comments for now.  I still have to go through the actual functions correctness.

src/main/java/com/mongodb/jdbc/MongoSQLFunctions.java
nbagnard: So I intentionally left the functions list completely separate because it seemed that they were quite different. BIT_LENGTH is not a supported MySQL function for example, so it does not make much sense to have it in the abstract class. I can refactor the common ones in the abstract class if you think that it will help readability.

src/main/java/com/mongodb/jdbc/MySQLFunctions.java
nbagnard: Yes, not sure where I got this from. Certainly does not apply here.

src/main/java/com/mongodb/jdbc/MongoSQLFunctions.java
nbagnard: So I intentionally left the functions list completely separate because it seemed that they were quite different. BIT_LENGTH is not a supported MySQL function for example, so it does not make much sense to have it in the abstract class. I can refactor the common ones in the abstract class if you think that it will help readability.

For the type names, I don't see the point of having the same ones redefined in another file. They should be shared everywhere in the driver. After you merge SQL-559, then we can make use of it here for example and likely in the result set implementation too.

bucaojit (approved):
LGTM!

Small comment on a variable.  
And the TRIM function we discussed taking additional parameters to match the syntax & semantics doc.

PR #60, opened 2021-11-02
SQL-552: Implement MongoSQLDatabaseMetaData db and table methods

mattChiaravalloti:
This PR implements the `MongoSQLDatabaseMetaData` database and table methods (`getColumns`, `getColumnPrivileges`, `getTables`, `getTablePrivileges`, and `getCatalogs`).

This ticket is also meant to cover the index methods (`getBestRowIdentifier`, `getIndexInfo`, and `getPrimaryKeys`). I am still working on those so I figured I'd put these up as a milestone since they're in a more ready state right now. I'll update this PR with the index methods soon.


src/main/java/com/mongodb/jdbc/MongoListCollectionsResult.java
```diff
@@ -0,0 +1,6 @@
+package com.mongodb.jdbc;
+
+public class MongoListCollectionsResult {
```
mattChiaravalloti, line 3: There are other fields returned by this, but these are the only relevant ones for this PR. We can extend this class as needed in the future.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -20,6 +38,11 @@ public MongoSQLDatabaseMetaData(MongoConnection conn) {
         super(conn);
     }
 
+    private BsonDocument createBottomBson(BsonElement... elements) {
+        BsonDocument bot = new BsonDocument(Arrays.asList(elements));
+        return new BsonDocument(BOT_NAME, bot);
+    }
```
mattChiaravalloti, line 44: This is introduced in #55 and I'll probably end up rebasing on top of that. Figured I'd get a head start using it here.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+                                        dbName ->
+                                                createBottomBson(
+                                                        new BsonElement(
+                                                                TABLE_CAT, new BsonString(dbName))))
+                                .collect(Collectors.toList()));
+
+        return new MongoSQLResultSet(null, c, botSchema);
+    }
+
+    /**
```
mattChiaravalloti, line 295: I know javadoc-style comments are necessary for private methods, but figured I'd go all out here to thoroughly explain what this does, why it exists, what its parameters are, what it returns, and when/why it throws a `SQLException`.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+                new BsonElement(TABLE_SCHEM, new BsonString("")),
+                new BsonElement(TABLE_NAME, new BsonString(i.collName)),
+                new BsonElement(COLUMN_NAME, new BsonString(i.columnName)),
+                new BsonElement(DATA_TYPE, new BsonInt32(info.getJDBCType())),
+                new BsonElement(TYPE_NAME, new BsonString(info.getTableAlias())),
+                new BsonElement(COLUMN_SIZE, new BsonNull()),
+                new BsonElement(BUFFER_LENGTH, new BsonInt32(0)),
+                new BsonElement(
+                        DECIMAL_DIGITS,
+                        new BsonNull()), // TODO DECIMAL_DIGITS -> Oliver's helper (may need to add new method)
```
mattChiaravalloti, line 376: In #55 , Oliver is adding utility methods for getting this information. I think I'll need to add a couple more to get these last bits of information. That PR is very close to being ready so I'll just wait on that to finish up these TODOs.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+                        : nullability == ResultSetMetaData.columnNullable ? "YES" : "";
+
+        return createBottomBson(
+                new BsonElement(TABLE_CAT, new BsonString(i.dbName)),
+                new BsonElement(TABLE_SCHEM, new BsonString("")),
+                new BsonElement(TABLE_NAME, new BsonString(i.collName)),
+                new BsonElement(COLUMN_NAME, new BsonString(i.columnName)),
+                new BsonElement(DATA_TYPE, new BsonInt32(info.getJDBCType())),
+                new BsonElement(TYPE_NAME, new BsonString(info.getTableAlias())),
+                new BsonElement(COLUMN_SIZE, new BsonNull()),
```
mattChiaravalloti, line 372: Not sure what to do here. The MySQL variant always returns `null` for this. (Well, it really returns `INFORMATION_SCHEMA.COLUMNS.CHARACTER_MAXIMUM_LENGTH`, but we always set that to `null` in the BIC/ADL catalog builder code.)

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
...
+    // Helper for creating BSON documents for the getTablePrivileges method. Intended
+    // for use with the getTablesFromDB helper method which is shared between getTables
+    // and getTablePrivileges.
+    private BsonDocument toGetTablePrivilegesDoc(String dbName, MongoListCollectionsResult res) {
+        return createBottomBson(
+                new BsonElement(TABLE_CAT, new BsonString(dbName)),
+                new BsonElement(TABLE_SCHEM, new BsonString("")),
+                new BsonElement(TABLE_NAME, new BsonString(res.name)),
+                new BsonElement(GRANTOR, new BsonNull()),
+                new BsonElement(GRANTEE, new BsonString("")),
```
mattChiaravalloti, line 158: The JDBC javadocs say `GRANTOR` may be null but does not say the same for `GRANTEE`. The design doc said to make these `null` or `""`, so I went with `null` for the former since it is valid there and empty string for the latter since `null` is not valid there.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+
+                // filter only for collections that have schemas
+                .filter(p -> p.right().ok == 1 && p.right().schema.jsonSchema.isObject())
+
+                // flatMap the column data into a single stream of BSON docs
+                .flatMap(
+                        p -> {
+                            Pair<String, String> ns = p.left();
+                            MongoJsonSchemaResult res = p.right();
+                            AtomicInteger idx = new AtomicInteger();
```
mattChiaravalloti, line 447: Used for tracking ordinal position (note that we base position on lexicographic sort order). Should I say that in a code comment?

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+        BsonExplicitCursor c = new BsonExplicitCursor(docsList);
+
+        return new MongoSQLResultSet(null, c, botSchema);
     }
 
     @Override
     public ResultSet getBestRowIdentifier(
             String catalog, String schema, String table, int scope, boolean nullable)
             throws SQLException {
+        // TODO
```
mattChiaravalloti, line 700: like I said in the PR summary -- working on these remaining 3 still. I cut out the current progress for this milestone so it didn't look too cluttered and confusing.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
nbagnard: Not sure why you need an AtomicInteger here? This is not multithreaded, is it (I don't think so otherwise you would potentially loose the sorting)?

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+                new Pair<>(IS_GRANTABLE, BSON_STRING_TYPE_NAME));
+
+        // All fields in this result set are nested under the bottom namespace.
+        MongoJsonSchema botSchema = MongoJsonSchema.createEmptyObjectSchema();
+        botSchema.properties.put(BOT_NAME, resultSchema);
+
+        // Note: JDBC has Catalogs, Schemas, and Tables: they are three levels of organization.
+        // MongoDB only has Databases (Catalogs) and Collections (Tables), so we ignore the
+        // schemaPattern argument.
+        Pattern tableNamePatternRE = Pattern.compile(tableNamePattern);
```
nbagnard, line 656: Also looks like it is missing a toJavaPattern here.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+            throw e;
+        }
+    }
+
+    // Helper class for representing all info needed to serialize column data for the
+    // getColumns and getColumnPrivileges methods. Intended for use with toGetColumnsDoc
+    // and toGetColumnPrivilegesDoc helpers.
+    private static class GetColumnsDocInfo {
+        String dbName;
+        String collName;
```
nbagnard, line 326: I understand why this is named "collName" since it is the collection name, but I must tell you that I got confused for a while at why you were applying a table name pattern to collName. I was reading colName and thought it was the column name. I guess once you are aware of this, it's easy enough to remember and check if it is collName or columnName, but would renaming this to collectionName or tableName look like a suitable option for you? For me, it will save us the trouble to have to be extra vigilant on that one and when you'll get new member in the team it will also make ramp-up easier because there will be no potential for error.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: it is not, I could just use a normal int and `++` I think

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
         throw new SQLFeatureNotSupportedException("TODO");
     }
 
+    // Helper for creating BSON documents for the getTables method. Intended for use
+    // with the getTablesFromDB helper method which is shared between getTables and
+    // getTablePrivileges.
+    private BsonDocument toGetTablesDoc(String dbName, MongoListCollectionsResult res) {
+        return createBottomBson(
+                new BsonElement(TABLE_CAT, new BsonString(dbName)),
+                new BsonElement(TABLE_SCHEM, new BsonString("")),
```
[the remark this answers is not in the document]
mattChiaravalloti: I had the same idea and we filed a separate ticket for this: https://jira.mongodb.org/browse/SQL-579

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
...
+            BiFunction<String, MongoListCollectionsResult, BsonDocument> bsonSerializer) {
+        MongoDatabase db = this.conn.getDatabase(dbName).withCodecRegistry(MongoDriver.registry);
+
+        return db.listCollections(MongoListCollectionsResult.class)
+                .into(new ArrayList<>())
+                .stream()
+                .filter(
+                        res ->
+                                tableNamePatternRE.matcher(res.name).matches()
+                                        && (types == null || types.contains(res.type)))
```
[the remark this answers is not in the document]
mattChiaravalloti: great catch, I'll be sure to make this case insensitive

src/main/java/com/mongodb/jdbc/MongoListCollectionsResult.java
pmeredit: sgtm, I assume you tested this, but at the same time, it's so simple it shouldn't be a problem.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+     * @throws SQLException If a RuntimeException that wraps a SQLException is caught.
+     */
+    private static <T> T liftSQLException(Supplier<T> f) throws SQLException {
+        try {
+            return f.get();
+        } catch (RuntimeException e) {
+            if (e.getCause() instanceof SQLException) {
+                throw (SQLException) e.getCause();
+            }
+            throw e;
```
bucaojit, line 317: Will there be a case where the `RuntimeException` has useful information and it wraps a SQLException?  I'm curious what the type of the RuntimeException is that wraps the SQLException.

bucaojit (changes requested):
Lots of great work on the database calls and sorting.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
...
+    // Helper for creating BSON documents for the getTables method. Intended for use
+    // with the getTablesFromDB helper method which is shared between getTables and
+    // getTablePrivileges.
+    private BsonDocument toGetTablesDoc(String dbName, MongoListCollectionsResult res) {
+        return createBottomBson(
+                new BsonElement(TABLE_CAT, new BsonString(dbName)),
+                new BsonElement(TABLE_SCHEM, new BsonString("")),
+                new BsonElement(TABLE_NAME, new BsonString(res.name)),
+                new BsonElement(TABLE_TYPE, new BsonString(res.type)),
+                new BsonElement(REMARKS, new BsonNull()),
```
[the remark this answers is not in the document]
pmeredit: Didn't know that existed, nice

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+     * RuntimeExceptions to appease the compiler. We want to propagate those SQLExceptions all the
+     * way out to the caller so this method is a way of recovering them.
+     *
+     * @param f The Supplier function to execute. This function may throw a RuntimeException that
+     *     wraps a SQLException.
+     * @param <T> The return type of the Supplier.
+     * @return The return value of the Supplier.
+     * @throws SQLException If a RuntimeException that wraps a SQLException is caught.
+     */
+    private static <T> T liftSQLException(Supplier<T> f) throws SQLException {
```
pmeredit, line 310: I feel like I'm reading Haskell code again :joy:

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
pmeredit:
```
The COLUMN_SIZE column specifies the column size for the given column. For numeric data, this is the maximum precision. For character data, this is the length in characters. For datetime datatypes, this is the length in characters of the String representation (assuming the maximum allowed precision of the fractional seconds component). For binary data, this is the length in bytes. For the ROWID datatype, this is the length in bytes. Null is returned for data types where the column size is not applicable.
```
We messed up in MySQL (oh well). I think we should return these based on types. I'm not sure what to do for string columns. Either null or figure out what max strings size in mongodb actually is (I suspect it's 4GB).

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+                        CHAR_OCTET_LENGTH,
+                        new BsonNull()), // TODO CHAR_OCTET_LENGTH -> Oliver's helper (may need to add new method)
+                new BsonElement(ORDINAL_POSITION, new BsonInt32(i.idx)),
+                new BsonElement(IS_NULLABLE, new BsonString(isNullable)),
+                new BsonElement(SCOPE_CATALOG, new BsonNull()),
+                new BsonElement(SCOPE_SCHEMA, new BsonNull()),
+                new BsonElement(SCOPE_TABLE, new BsonNull()),
+                new BsonElement(SOURCE_DATA_TYPE, new BsonInt32(0)),
+                new BsonElement(IS_AUTOINCREMENT, new BsonString("")),
+                new BsonElement(IS_GENERATEDCOLUMN, new BsonInt32(0)));
```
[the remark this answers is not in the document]
pmeredit: Yes, it's bizarre

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: ah this is because of the compiler error "Variable used in lambda expression should be final or effectively final." My IDE suggested using `AtomicInteger` or using a 1-element array, since making this a `final int` would not be helpful. `AtomicInteger`, while overkill for a single-threaded use case, is still useful here simply because it is an int wrapper that has the `++` functionality we need here.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
...
+
+        // All fields in this result set are nested under the bottom namespace.
+        MongoJsonSchema botSchema = MongoJsonSchema.createEmptyObjectSchema();
+        botSchema.properties.put(BOT_NAME, resultSchema);
+
+        // Note: JDBC has Catalogs, Schemas, and Tables: they are three levels of organization.
+        // MongoDB only has Databases (Catalogs) and Collections (Tables), so we ignore the
+        // schemaPattern argument.
+        Pattern tableNamePatternRE = Pattern.compile(tableNamePattern);
+        List<String> typesList = Arrays.asList(types);
```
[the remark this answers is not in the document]
mattChiaravalloti: good catch, thanks!

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: There never will be (unless some code not owned by us catches `SQLException`s and wraps them in `RuntimeExceptions`, which I suspect is not the case).

The type that wraps the `SQLException` _is_ `RuntimeException` (which is a concrete class). In a few places in this PR, we explicitly catch `SQLException`s in streams and immediately wrap them in and throw runtime exceptions. All places that happen are surrounded by this `liftSQLException` method. So the only time we'll hit this code is if we specifically intended to.

src/main/java/com/mongodb/jdbc/MongoListCollectionsResult.java
mattChiaravalloti: yup, tested it!

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
...
+
+        Stream<BsonDocument> docs;
+        if (catalog == null) {
+            // If no catalog (database) is specified, get tables for all databases.
+            docs =
+                    this.conn
```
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
...
+                                                    dbName,
+                                                    tableNamePatternRE,
+                                                    typesList,
+                                                    this::toGetTablesDoc));
+        } else {
+            docs = getTablesFromDB(catalog, tableNamePatternRE, typesList, this::toGetTablesDoc);
+        }
+
+        // Collect to a sorted list
+        List<BsonDocument> docsList =
```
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -106,11 +129,126 @@ public ResultSet getTableTypes() throws SQLException {
...
+                new Pair<>(REF_GENERATION, BSON_STRING_TYPE_NAME));
+
+        // All fields in this result set are nested under the bottom namespace.
+        MongoJsonSchema botSchema = MongoJsonSchema.createEmptyObjectSchema();
+        botSchema.properties.put(BOT_NAME, resultSchema);
+
+        // Note: JDBC has Catalogs, Schemas, and Tables: they are three levels of organization.
+        // MongoDB only has Databases (Catalogs) and Collections (Tables), so we ignore the
+        // schemaPattern argument.
+        Pattern tableNamePatternRE = Pattern.compile(tableNamePattern);
```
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+                new BsonElement(
+                        CHAR_OCTET_LENGTH,
+                        new BsonNull()), // TODO CHAR_OCTET_LENGTH -> Oliver's helper (may need to add new method)
+                new BsonElement(ORDINAL_POSITION, new BsonInt32(i.idx)),
+                new BsonElement(IS_NULLABLE, new BsonString(isNullable)),
+                new BsonElement(SCOPE_CATALOG, new BsonNull()),
+                new BsonElement(SCOPE_SCHEMA, new BsonNull()),
+                new BsonElement(SCOPE_TABLE, new BsonNull()),
+                new BsonElement(SOURCE_DATA_TYPE, new BsonInt32(0)),
+                new BsonElement(IS_AUTOINCREMENT, new BsonString("")),
```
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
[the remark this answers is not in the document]
mattChiaravalloti: Done

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -129,33 +267,437 @@ public ResultSet getSchemas() throws SQLException {
...
+                new Pair<>(TABLE_SCHEM, BSON_STRING_TYPE_NAME),
+                new Pair<>(TABLE_NAME, BSON_STRING_TYPE_NAME),
+                new Pair<>(GRANTOR, BSON_STRING_TYPE_NAME),
+                new Pair<>(GRANTEE, BSON_STRING_TYPE_NAME),
+                new Pair<>(PRIVILEGE, BSON_STRING_TYPE_NAME),
+                new Pair<>(IS_GRANTABLE, BSON_STRING_TYPE_NAME));
+
+        // All fields in this result set are nested under the bottom namespace.
+        MongoJsonSchema botSchema = MongoJsonSchema.createEmptyObjectSchema();
+        botSchema.properties.put(BOT_NAME, resultSchema);
```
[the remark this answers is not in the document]
mattChiaravalloti: Done

mattChiaravalloti:
Finally got this back into review. All previous comments have been addressed and the remaining methods have been implemented 👍

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -23,8 +25,58 @@
     public static final String SYMBOL_TYPE_NAME = "symbol";
     public static final String TIMESTAMP_TYPE_NAME = "timestamp";
     public static final String UNDEFINED_TYPE_NAME = "undefined";
+    public static final String NULL_TYPE_NAME = "null";
```
mattChiaravalloti, line 28: Null is a BSON type and was missing from this class. In addition to the new methods, I also updated the existing ones to handle this type as well. I had it behave the same way as `Undefined`.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -23,8 +25,58 @@
...
     public static final String BSON_TYPE_NAME = "bson";
 
+    public static Pair<String, Integer> getBsonTypeNameAndNullability(
```
mattChiaravalloti, line 31: This method is 90% the same as the `MongoSQLColumnInfo` constructor. The differences are subtle, but I figured for now we'd just stick with the duplication and have the refactor ticket in the scheduled queue address unification.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -175,9 +231,111 @@ public static int getNumPrecRadix(String typeName) throws SQLException {
...
+            case TIMESTAMP_TYPE_NAME:
+            case UNDEFINED_TYPE_NAME:
+            case NULL_TYPE_NAME:
+            case BSON_TYPE_NAME:
+                return null;
+        }
+        throw new SQLException("unknown bson typeName: " + typeName);
+    }
+
+    public static Integer getCharOctetLength(String typeName) throws SQLException {
```
mattChiaravalloti, line 272: These are the values we've been returning for MySQL. Not sure if they are exactly what we're supposed to be returning here...

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -116,8 +137,8 @@ public static String escapeString(String value) {
         return escaped.replace("\\", "\\\\");
     }
 
-    public static String toJavaPattern(String sqlPattern) {
-        return sqlPattern.replaceAll("%", ".*").replaceAll("_", ".");
+    public static Pattern toJavaPattern(String sqlPattern) {
+        return Pattern.compile(sqlPattern.replaceAll("%", ".*").replaceAll("_", "."));
```
mattChiaravalloti, line 141: every call to this method looked the same:
```
Pattern xyz = Pattern.compile(toJavaPattern(sqlPattern));
```
so I decided to just do the `Pattern` creation _in_ the method itself.

src/main/java/com/mongodb/jdbc/MongoSQLFunctions.java
```diff
@@ -104,16 +104,16 @@ public static MongoSQLFunctions getInstance() {
...
-                                            "TRIM",
-                                            "string",
-                                            "returns the string str with all remstr prefixes and/or suffixes removed.",
-                                            new String[] {"string", "string"}),
-                                    new MongoFunction(
-                                            "TRIM",
-                                            "string",
-                                            "returns the string str with all remstr prefixes and/or suffixes removed.",
-                                            new String[] {"string", "string", "string"}),
+                                new MongoFunction(
```
mattChiaravalloti, line 107: result of `spotlessApply`

src/main/java/com/mongodb/jdbc/SortableBsonDocument.java
```diff
@@ -0,0 +1,61 @@
...
+        this.sortSpecs = sortSpecs;
+        this.nestedDocValue = docValue;
+    }
+
+    @Override
+    public int compareTo(SortableBsonDocument o) {
+        int r = 0;
+        for (SortSpec sortSpec : this.sortSpecs) {
+            switch (sortSpec.type) {
+                case String:
```
mattChiaravalloti, line 38: Note that we need to use `getString` or `getInt32` on the `BsonDocument` to get the relevant sort field. That is the only reason we have this type enum and switch on it. The `BsonValue` class returned by the generic `get` method does not implement `Comparable`, unfortunately. Open to other suggestions on how to handle sorting but I think this is a reasonable alternative to the massive in-line `Comparator`s I had last time around.

src/main/java/com/mongodb/jdbc/SortableBsonDocument.java
```diff
@@ -0,0 +1,61 @@
...
+
+    enum ValueType {
+        String,
+        Int,
+    }
+
+    List<SortSpec> sortSpecs;
+    BsonDocument nestedDocValue;
+
+    SortableBsonDocument(List<SortSpec> sortSpecs, String key, BsonDocument docValue) {
```
mattChiaravalloti, line 26: Note that this is exclusively constructed as `new SortableBsonDocument(<sortSpecs>, BOT_NAME, botDoc)`, and since we compare values nested in the `botDoc`, I just store that value to avoid additional `this.getDocument(BOT_NAME)` calls in the `compareTo` method.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -1,38 +1,122 @@
...
+                            TABLE_NAME, SortableBsonDocument.ValueType.String),
+                    new SortableBsonDocument.SortSpec(
+                            ORDINAL_POSITION, SortableBsonDocument.ValueType.Int));
+
+    private static final List<SortableBsonDocument.SortSpec> GET_COLUMN_PRIVILEGES_SORT_SPECS =
+            Collections.singletonList(
+                    new SortableBsonDocument.SortSpec(
+                            COLUMN_NAME, SortableBsonDocument.ValueType.String));
+
+    private static final List<SortableBsonDocument.SortSpec> GET_PRIMARY_KEYS_SORT_SPECS =
```
mattChiaravalloti, line 69: This is the same as the one above (`GET_COLUMN_PRIVILEGES_SORT_SPECS`). Should we use the same one twice or construct these two separately so we can stick with the naming pattern?

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -91,18 +175,14 @@ public String getTimeDateFunctions() throws SQLException {
...
-                new Pair<>(PROCEDURE_CAT, BsonTypeInfo.STRING_TYPE_NAME),
-                new Pair<>(PROCEDURE_SCHEM, BsonTypeInfo.STRING_TYPE_NAME),
-                new Pair<>(PROCEDURE_NAME, BsonTypeInfo.STRING_TYPE_NAME),
-                new Pair<>(REMARKS, BsonTypeInfo.STRING_TYPE_NAME),
-                new Pair<>(PROCEDURE_TYPE, BsonTypeInfo.INT_TYPE_NAME),
-                new Pair<>(SPECIFIC_NAME, BsonTypeInfo.STRING_TYPE_NAME));
-
-        // All fields in this result set are nested under the bottom namespace.
-        MongoJsonSchema botSchema = MongoJsonSchema.createEmptyObjectSchema();
-        botSchema.properties.put(BOT_NAME, resultSchema);
```
mattChiaravalloti, line 105: All schema creation code followed this exact same pattern, so I extracted all of that into the `createBottomSchema` helper method.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -161,130 +237,766 @@ public ResultSet getTableTypes() throws SQLException {
...
+                .filter(tableName -> tableNamePatternRE.matcher(tableName).matches())
+
+                // map the collection names into triples of (dbName, tableName, tableSchema)
+                .map(
+                        tableName ->
+                                new Pair<>(
+                                        new Pair<>(dbName, tableName),
+                                        db.runCommand(
+                                                new BsonDocument(
+                                                        "sqlGetSchema", new BsonString(tableName)),
```
bucaojit, line 569: Is this command correct: `sqlGetSchema`?  In MongoSQLStatement the command used is `sqlGetResultSchema`.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -298,34 +1010,73 @@ public ResultSet getCrossReference(
...
+        return keys.keySet()
+                .stream()
+                .map(
+                        key ->
+                                createSortableBottomBson(
+                                        // Per JDBC spec, sort by COLUMN_NAME.
+                                        GET_PRIMARY_KEYS_SORT_SPECS,
+                                        new BsonElement(
+                                                TABLE_CAT, new BsonString(namespace.left())),
+                                        new BsonElement(TABLE_SCHEM, new BsonString("")),
```
bucaojit, line 1052: Since there isn't an actual schema, it seems that `null` would be the appropriate value here.

bucaojit (changes requested):

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -161,130 +237,766 @@ public ResultSet getTableTypes() throws SQLException {
...
+                        new Pair<>(REF_GENERATION, BsonTypeInfo.STRING_TYPE_NAME));
+
+        // Note: JDBC has Catalogs, Schemas, and Tables: they are three levels of organization.
+        // MongoDB only has Databases (Catalogs) and Collections (Tables), so we ignore the
+        // schemaPattern argument.
+        Pattern tableNamePatternRE = toJavaPattern(tableNamePattern);
+        List<String> typesList = toTableTypeList(types);
+
+        Stream<BsonDocument> docs;
+        if (catalog == null) {
```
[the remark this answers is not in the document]
mattChiaravalloti: Oh thanks for trying that. My expectation was that `catalog = ""` wouldn't need special handling in any of these methods. I figured doing `this.conn.getDatabase("").command(...).stream()...` would just naturally result in empty result set since there would never be a database with that name.

I'll update these methods to explicitly handle the empty string case.

Side question for my own understanding: where was that IllegalArgumentException thrown? was it thrown by the java driver? something in our code? somewhere else?

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: Yes, `sqlGetSchema` ([docs](https://docs.mongodb.com/datalake/reference/cli/sql/sqlgetschema/)) is the command used to get the schema for a _collection_. The command you used in the other PR is `sqlGetResultSchema` which is used to get the result set schema for a _query_. Here, we need the collection schema so we could get type information for the fields.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: I went with empty since that's what we've been doing for the MySQLDatabaseMetaData implementation. I'll switch these to null though because that is probably the more accurate result value.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
pmeredit: Ah yes, this is because Java's safety algorithm is overly conservative.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
pmeredit: Agreed. Unification should be easy, but I like it in a separate ticket.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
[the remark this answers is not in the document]
bucaojit: It is thrown in the java driver, the check for the empty string is [here](https://github.com/mongodb/mongo-java-driver/blob/a04a96c3f1afe044bc4396b81244608a95c6494d/driver-core/src/main/com/mongodb/MongoNamespace.java#L59).

The stack trace is:
```
state should be: databaseName is not empty
java.lang.IllegalArgumentException: state should be: databaseName is not empty
	at com.mongodb.assertions.Assertions.isTrueArgument(Assertions.java:101)
	at com.mongodb.MongoNamespace.checkDatabaseNameValidity(MongoNamespace.java:59)
	at com.mongodb.client.internal.MongoDatabaseImpl.<init>(MongoDatabaseImpl.java:74)
	at com.mongodb.client.internal.MongoClientImpl.getDatabase(MongoClientImpl.java:80)
	at com.mongodb.jdbc.MongoConnection.getDatabase(MongoConnection.java:97)
	at com.mongodb.jdbc.MongoSQLDatabaseMetaData.getTableDataFromDB(MongoSQLDatabaseMetaData.java:251)
	at com.mongodb.jdbc.MongoSQLDatabaseMetaData.getTableDataFromDB(MongoSQLDatabaseMetaData.java:305)
	at com.mongodb.jdbc.MongoSQLDatabaseMetaData.getTables(MongoSQLDatabaseMetaData.java:359)
```

src/main/java/com/mongodb/jdbc/SortableBsonDocument.java
```diff
@@ -0,0 +1,61 @@
...
+                                    .compareTo(o.nestedDocValue.getString(sortSpec.key));
+                    break;
+                case Int:
+                    r =
+                            this.nestedDocValue
+                                    .getInt32(sortSpec.key)
+                                    .compareTo(o.nestedDocValue.getInt32(sortSpec.key));
+                    break;
+                default:
+                    throw new IllegalArgumentException("unreachable");
```
nbagnard, line 51: The switch covers all cases for ValueType. There is no need for a default case. 
If you want/need to keep it, can you detail the error a bit more, just in case? Something like "Invalid column type for sorting" or something along this line?

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
nbagnard: Both Mysql and MongoSQL are retrieving BSON data, so it should be the same, no?
If when executing a query with MongoSQL you have extra info which tell you the size of a char column, then you can use it to fine-tune this.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -175,9 +231,111 @@ public static int getNumPrecRadix(String typeName) throws SQLException {
...
+            case DECIMAL_TYPE_NAME:
+                return 34;
+            case DOUBLE_TYPE_NAME:
+                return 15;
+            case INT_TYPE_NAME:
+            case LONG_TYPE_NAME:
+            case BINDATA_TYPE_NAME:
+            case STRING_TYPE_NAME:
+            case BOOL_TYPE_NAME:
+            case DATE_TYPE_NAME:
```
bucaojit, line 252: Date type can have 3 decimal digits: https://docs.mongodb.com/manual/reference/mongodb-extended-json/#mongodb-bsontype-Date

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
pmeredit: Actually, I expected to use the `ResultSetMetadata` here, but this is good for now.

pmeredit (approved):
This looks great!

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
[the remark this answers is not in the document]
mattChiaravalloti: Thanks!

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -161,130 +237,766 @@ public ResultSet getTableTypes() throws SQLException {
...
+            }
+
+        } catch (SQLException e) {
+            throw new RuntimeException(e);
+        }
+
+        BsonValue isNullable =
+                new BsonString(
+                        nullability == columnNoNulls
+                                ? "NO"
```
nbagnard, line 488: Note : Double-check that SQL-559 has BsonString("Yes) and BSonString("No"). If not, please add it and use for isNullable and isAutoincrement and whereever else the data is "Yes"/"No".

src/main/java/com/mongodb/jdbc/SortableBsonDocument.java
mattChiaravalloti: If I remember correctly, the compiler was giving me a complaint without the default. I'll work on updating this to either remove it or improving the message 👍

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
mattChiaravalloti: Do you mean [SQL-579](https://jira.mongodb.org/browse/SQL-579)?

mattChiaravalloti: Added a comment calling this out: https://jira.mongodb.org/browse/SQL-579?focusedCommentId=4186564&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-4186564

nbagnard (approved):

bucaojit (approved):
Amazing work!  Thanks for addressing all the comments. 💯

PR #61, opened 2021-11-16
SQL-566: Create very rough outline of script for starting local ADL against local mongod

mattChiaravalloti:
@bucaojit This is the bare-bones script. Feel free to just use it as a jumping off point!

PR #63, opened 2021-12-01
SQL-566: Create script that starts local mongod and local ADL

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-566

Based on initial commit by @mattChiaravalloti here: 
https://github.com/mongodb/mongo-jdbc-driver/pull/61/commits/a77c42a837bc50d78af575b529a6c182ba93d69c

Script will install and run mongod if not already running on port 28017.
If ADL isn't running, it will download latest version of golang, mongohouse repo, and start ADL. 

Tested on supported platforms: macos, ubuntu1804, rhel7
No support added for windows


resources/run_adl.sh
```diff
@@ -0,0 +1,158 @@
...
+  go run cmd/buildscript/build.go tools:download:mongosql
+  
+  # Set relevant environment variables
+  export GOPRIVATE=github.com/10gen
+  export MONGOHOUSE_ENVIRONMENT="local"
+  export MONGOHOUSE_MQLRUN="$(pwd)/artifacts/mqlrun"
+  export LIBRARY_PATH="$(pwd)/artifacts"
+  
+  # Load tenant config into mongodb
+  # TODO - we should probably make a different tenant config than the one that exists at
```
bucaojit, line 148: This may take some effort, and may not be necessary.  I propose I create a ticket for this and address it in a future release if needed.

resources/run_adl.sh
```diff
@@ -0,0 +1,158 @@
...
+  #  ./testdata/config/mongodb_local/tenant-config.json since that one includes a ton of extra datasources
+  #  we not only don't need but likely don't want. We'll probably want a tenant config that has a wildcard
+  #  so we can have any databases and collections in the local mongod.
+  #  - We'll probably want the test runner, as opposed to this script, to be able to say what data should get
+  #    restored.
+  go run cmd/buildscript/build.go init:mongodb-tenant 
+  
+  # Start mongohoused with appropriate config
+  go run -tags mongosql ./cmd/mongohoused/mongohoused.go --config ./testdata/config/mongodb_local/frontend-agent-backend.yaml
```
bucaojit, line 157: One thing to determine, and this may also depend on how we are running this script, is how we want the script to end.  The last line is the `go run` command so the script will has to continue running, and the output is to stdout.  Another alternative would be to use `nohup` and send the output to a log file.

resources/run_adl.sh
bucaojit: You're right.  I tried querying the underlying mongod data and it turns out I needed to update the config to add the storage/stores configuration.  I'll see what is a good way to add that config, perhaps there's a straightforward way to add to the existing config on ADL.

resources/run_adl.sh
```diff
@@ -0,0 +1,158 @@
...
+  
+  OS_LOWERCASE=$(echo $OS | tr '[:upper:]' '[:lower:]')
+  curl -O "https://dl.google.com/go/${GOVERSION}.${OS_LOWERCASE}-amd64.tar.gz"
+  mkdir $GO_DIR
+  tar -C $GO_DIR -xzf ${GOVERSION}.${OS_LOWERCASE}-amd64.tar.gz
+  export PATH=$GO_DIR/go/bin:$PATH
+  # May have been set from previous install
+  unset GOROOT
+  
+  go mod vendor
```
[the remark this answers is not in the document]
bucaojit: Sounds good.  `go mod download` worked for me.  Changing it to that.

resources/run_adl.sh
mattChiaravalloti: Yea we could possibly make use of something known as "wildcard" datasources in ADL so we could query arbitrary mongodb data. I don't know the full details off-hand but some of the docs (like [these](https://docs.mongodb.com/datalake/admin/manage-namespace-catalog-cli/)) may help, or we could ask the ADL team for guidance.

One idea is we could copy the config file we're currently using from the mongohouse repo and just include it in our repo in the resources folder. We could remove all the extra datasources we won't use and just add a wildcard mongodb source.

bucaojit: Added a wildcard collection for the tenant-config.json.  It uses the 'jq' command line tool, which I confirmed is on Ubuntu and RHEL.  I don't have access to the macOS host yet to check.
The script deletes the existing `storage` section and then adds the wildcard collection for the database: integration_test.  I confirmed that is the only database present on the local ADL instance.

resources/run_adl.sh
bucaojit: Sounds good.  Made the change to use `nohup` and sends the output to a log file.

resources/run_adl.sh
```diff
@@ -0,0 +1,158 @@
...
+    git clone $MONGOHOUSE_URI
+  fi
+  cd $MONGOHOUSE_DIR
+  git pull $MONGOHOUSE_URI
+  
+  OS_LOWERCASE=$(echo $OS | tr '[:upper:]' '[:lower:]')
+  curl -O "https://dl.google.com/go/${GOVERSION}.${OS_LOWERCASE}-amd64.tar.gz"
+  mkdir $GO_DIR
+  tar -C $GO_DIR -xzf ${GOVERSION}.${OS_LOWERCASE}-amd64.tar.gz
+  export PATH=$GO_DIR/go/bin:$PATH
```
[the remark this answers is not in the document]
bucaojit: Thanks for that, I didn't know about the pre-installed go versions.  I've switched it to use the install under `/opt/golang/go1.16`.

resources/run_adl.sh
```diff
@@ -0,0 +1,158 @@
...
+  # May have been set from previous install
+  unset GOROOT
```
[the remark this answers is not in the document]
bucaojit: This is no longer needed, I have removed the `unset`.  
The GOROOT was pointing to the older version of GO ie 1.7 and unsetting it allowed the downloaded version to be used instead.

nbagnard (approved):
LGTM!

resources/run_adl.sh
```diff
@@ -0,0 +1,195 @@
+#!/bin/bash
+
+NAME=`basename "$0"`
+if [[ $SKIP_RUN_ADL -eq 1 ]]; then
```
mattChiaravalloti, line 4: Should we document what any of these external variables--such as `SKIP_RUN_ADL`--are in the README or at least in this script?

resources/run_adl.sh
```diff
@@ -0,0 +1,195 @@
...
+  echo "Skipping $NAME"
+  exit 0
+fi
+
+ARG=`echo $1 | tr '[:upper:]' '[:lower:]'`
```
mattChiaravalloti, line 9: Should we document how this script is intended to be run? Like what this and possibly other arguments are supposed to represent?

resources/run_adl.sh
```diff
@@ -0,0 +1,195 @@
...
+    git pull $MONGOHOUSE_URI
+
+    go mod download
+
+    # Download external dependencies
+    go run cmd/buildscript/build.go tools:download:mqlrun
+    go run cmd/buildscript/build.go tools:download:mongosql
+
+    # Set relevant environment variables
+    export GOPRIVATE=github.com/10gen
```
mattChiaravalloti, line 162: I'm surprised you don't need to do this before doing the `go mod download` since that command is one that _should_ rely on this. For example, when it attempts to download the `10gen/mongoast` dependency I'm surprised it doesn't complain. If this is working, I suppose there is no need to change it.

resources/run_adl.sh
```diff
@@ -0,0 +1,195 @@
...
+      --config ./testdata/config/mongodb_local/frontend-agent-backend.yaml >> $LOGS_PATH/${MONGOHOUSED}.log &
+    echo $! > $TMP_DIR/${MONGOHOUSED}.pid
+  fi
+else
+  if [ $ARG = $STOP ]; then
+    MONGOHOUSED_PID=$(< $TMP_DIR/${MONGOHOUSED}.pid)
+    echo "Stopping $MONGOHOUSED, pid $MONGOHOUSED_PID"
+    pkill -TERM -P ${MONGOHOUSED_PID}
+  fi
+fi
```
mattChiaravalloti, line 195:
```suggestion
fi

```
[nit] ending with newline

resources/run_adl.sh
```diff
@@ -0,0 +1,195 @@
...
+  MONGO_DOWNLOAD_LINK=$MONGO_DOWNLOAD_MAC
+else
+  echo $(uname) not supported
+  exit 1
+fi
+
+check_mongod
+if [[ $? -ne 0 ]]; then
+  if [ $ARG = $START ]; then
+    # Install and start mongod
```
mattChiaravalloti, line 117: Let's consider adding `echo "Starting mongod"` like we have logging for stopping it in the else branch.

resources/run_adl.sh
```diff
@@ -0,0 +1,195 @@
...
+    MONGOD_PID=$(< $TMP_DIR/${MONGOD}.pid)
+    echo "Stopping $MONGOD, pid $MONGOD_PID"
+    kill "$(< $TMP_DIR/${MONGOD}.pid)"
+  fi
+fi
+
+check_mongohoused
+if [[ $? -ne 0 ]]; then
+  if [ $ARG = $START ]; then
+    # Install and start mongohoused
```
mattChiaravalloti, line 146: Let's consider adding` echo "Starting mongohoused"` like we have logging for stopping it in the else branch.

mattChiaravalloti (changes requested):
Super close 💯 Just a few last touch-ups then this should be good to go

resources/run_adl.sh
bucaojit: Added a description to the beginning of the script and in the README

resources/run_adl.sh
bucaojit: I looked into this a bit and I'm finding that the command that helps `go mod download` succeed (tested on a new ubuntu1804 host) is:
```git config --global url.git@github.com:.insteadOf https://github.com/```
Otherwise it complains with the error:
```
$ GOPRIVATE=github.com/10gen go mod download
go: github.com/10gen/bsonio@v0.0.0-20211202155813-2e4634d25cd1: invalid version: git fetch -f origin refs/heads/*:refs/heads/* refs/tags/*:refs/tags/* in /home/ubuntu/go/pkg/mod/cache/vcs/3a9008e6621e88540dcd7b013a52c240ba3677268aedf02ee2f397893fbb4f73: exit status 128:
	fatal: could not read Username for 'https://github.com': terminal prompts disabled
```
I'll move the export GOPRIVATE above the `go mod download` command just in case.

mattChiaravalloti: oh that makes so much sense -- thanks for looking into that and figuring out what was happening

mattChiaravalloti (approved):
lgtm!

PR #64, opened 2021-12-13
SQL-563: Add DatabaseMetaData integration tests

mattChiaravalloti:
This PR adds integration tests for the DatabaseMetaData methods. ~The tests are effectively just stubs right now, though they do cover interesting cases. Once #62 is merged, the test-generating code should bump these stubs out into full-fledged tests.~ I was hesitant to add more data to the testdata set for 2 reasons: 1. SQL-561 is still WIP so I figured I should wait for that to be settled, and 2. I'm not convinced we need many tests for very large datasets -- especially not for these methods. Specifically, these database metadata methods should behave the same regardless of scale, and any changes that come with scale (such as reading from a large cursor), should be completely blackboxed for us by the java driver and therefore are not interesting to test from the jdbc perspective. Let me know if you disagree.

This is updated with mostly generated tests and skip reasons where appropriate (I think we're actually skipping most tests 😅  but that's ok! we'll get to the bugs soon enough)


resources/integration_test/testdata/adl_db_config.json
```diff
@@ -0,0 +1,38 @@
+[
```
mattChiaravalloti, line 1: I decided to make the ADL storage config DATABASES field a separate file so we can more easily manipulate it when needed.

resources/integration_test/testdata/dbmd_variable_result_set_data.yml
```diff
@@ -0,0 +1,20 @@
+dataset:
+  - db: integration_test
+    collection: foo
+    docs:
+      - {"_id": 0, "a": 1}
+      - {"_id": 1, "a": 2}
+    nonuniqueIndexes:
+      - {"a": 1}
```
mattChiaravalloti, line 8: Added a way to define nonuniqueIndexes. Not sure if we want/need the ability to create general indexes with options. This is really just useful for one test so I'm ok with this special-purpose field. LMK if you feel differently!

resources/integration_test/tests/dbmd_variable_result_sets_function_methods.yml
```diff
@@ -0,0 +1,182 @@
...
+    meta_function: [getFunctions, integration_test, null, '%']
+    expected_result:
+      - [def, BIT_LENGTH, null, 1, returns length of string in bits, BIT_LENGTH]
+      - [def, CAST, null, 1, converts the provided expression into a value of the
+                               specified type., CAST]
+      - [def, CHAR_LENGTH, null, 1, returns length of string, CHAR_LENGTH]
+      - [def, COALESCE, null, 1, 'returns the first non-null value in the list, or
+          null if there are no non-null values.', COALESCE]
+      - [def, CURRENT_TIMESTAMP, null, 1, returns the current date and time., CURRENT_TIMESTAMP]
+      - [def, CURRENT_TIMESTAMP, null, 1, returns the current date and time., CURRENT_TIMESTAMP]
```
mattChiaravalloti, line 15: Overloaded functions are returned multiple times. This seems like valid behavior as this is what we do in the (released) MySQL version of the driver and nothing in the docs says it is incorrect.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
```diff
@@ -70,7 +70,7 @@ private void generateSchema() {
     private void readDataFiles(String dataDirectory) throws IOException {
         File folder = new File(dataDirectory);
         for (final File fileEntry : Objects.requireNonNull(folder.listFiles())) {
-            if (fileEntry.isFile()) {
+            if (fileEntry.isFile() && fileEntry.getName().endsWith(".yml")) {
```
mattChiaravalloti, line 73: Since I decided to put the db_config in the `testdata` directory

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -1030,7 +1030,7 @@ public ResultSet getCrossReference(
                                                 TABLE_NAME, new BsonString(namespace.right())),
                                         new BsonElement(COLUMN_NAME, new BsonString(key)),
                                         new BsonElement(
-                                                KEY_SEQ, new BsonInt32(pos.getAndIncrement())),
+                                                KEY_SEQ, new BsonInt32(pos.incrementAndGet())),
```
mattChiaravalloti, line 1033: I did this drive-by fix instead of filing a skip-reason ticket since it is so small and the index methods already have so much blocking them.

bucaojit (commented):
Good work covering the database metadata functions and cases.

In response to the description, I think that is fine to not have large datasets as well.  It's good that collections were added for the variety of name filtering and the unique and non-unique index.  

The test harness is limited to testing the functions that return a ResultSet.  There are a few methods in MongoSQLDatabaseMetaData that return a String such as getSQLKeywords(), getNumericFunctions(), getStringFunctions(), getSystemFunctions(), getTimeDateFunctions().  It would be nice to write a separate test case that checks the returns of these for completeness.

mattChiaravalloti:
> The test harness is limited to testing the functions that return a ResultSet. There are a few methods in MongoSQLDatabaseMetaData that return a String such as getSQLKeywords(), getNumericFunctions(), getStringFunctions(), getSystemFunctions(), getTimeDateFunctions(). It would be nice to write a separate test case that checks the returns of these for completeness.

Yea I ran into this since I had written stub yml tests for all methods. I removed these and then forgot to add unit tests for them. I'l add unit tests for them now 👍

bucaojit (approved):
Looks good!  Thanks for making the suggested changes.

PR #65, opened 2021-12-28
SQL-602: Refactor shared BSON information into single class

mattChiaravalloti:
This PR creates the `BsonTypeInfo` enum (as proposed [here](https://docs.google.com/document/d/1U4xASLn4Z3VskUCMZ7CYAXGouUQF7ucAiT23J17LQSo/edit)) and replaces all references to BSON information (such as type names, precisions, scales, etc) with references to variants of this new enum as a single source of information.

Note I did not implement the suggestion of using a flag to indicate whether the context is MongoSQL or MySQL. The refactor primarily focused on updating the MongoSQL* classes. The MySQL classes were updated but only to use the bson type names from the enum. Let me know if you think it is worth extending this further to fully update the MySQL classes as well. I'm not sure that would be worth the effort for a minimal win since, as I often come back to, the MySQL dialect will not be around in this iteration of the driver for long.


src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -1,342 +1,230 @@
...
+                            BSON_DBPOINTER.bsonName,
+                            BSON_JAVASCRIPT.bsonName,
+                            BSON_SYMBOL.bsonName,
+                            BSON_JAVASCRIPTWITHSCOPE.bsonName,
+                            BSON_INT.bsonName,
+                            BSON_TIMESTAMP.bsonName,
+                            BSON_LONG.bsonName,
+                            BSON_DECIMAL.bsonName,
+                            BSON_MINKEY.bsonName,
+                            BSON_MAXKEY.bsonName));
```
bucaojit, line 68: Missing BSON_BSON type

```suggestion
                            BSON_BSON.bsonName,
                            BSON_MAXKEY.bsonName));
```

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -1,342 +1,230 @@
...
-                return Types.DOUBLE;
+        // All type names can be determined uniquely off a combination of first letter
+        // and length except for "minKey" vs "maxKey" and "string" vs "symbol".
+        switch (typeName.charAt(0)) {
+            case 'a':
+                return BSON_ARRAY;
+            case 'b':
+                switch (typeName.length()) {
+                    case 4:
+                        return BSON_BOOL;
```
bucaojit, line 164: Adding the BSON type to the switch. 

```suggestion
                    case 4:
                        switch (typeName.charAt(1)) {
                            case 'o':
                                return BSON_BOOL;
                            case 's':
                                return BSON_BSON;
```

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -1,342 +1,230 @@
...
+
+public enum BsonTypeInfo {
+    BSON_DOUBLE("double", BsonType.DOUBLE, Types.DOUBLE, false, 15, 15, 2, 15, 15, 8),
+    BSON_STRING("string", BsonType.STRING, Types.LONGNVARCHAR, true, 0, 0, 0, null, null, null),
+    BSON_OBJECT("object", BsonType.DOCUMENT, Types.OTHER, false, 0, 0, 0, null, null, null),
+    BSON_ARRAY("array", BsonType.ARRAY, Types.OTHER, false, 0, 0, 0, null, null, null),
+    BSON_BINDATA("binData", BsonType.BINARY, Types.BINARY, false, 0, 0, 0, null, null, null),
+    BSON_UNDEFINED("undefined", BsonType.UNDEFINED, Types.OTHER, false, 0, 0, 0, null, null, null),
+    BSON_OBJECTID("objectId", BsonType.OBJECT_ID, Types.OTHER, false, 0, 0, 0, 24, null, null),
+    BSON_BOOL("boolean", BsonType.BOOLEAN, Types.BIT, false, 0, 0, 0, 1, null, 1),
```
bucaojit, line 18: Should this be `bool` instead of `boolean`?  The previous type name was `bool`, and a test is currently failing with:
```
FAILURE: java.sql.SQLException: Unknown bson type name: "bool" (java.sql.SQLException)
```

src/test/java/com/mongodb/jdbc/BsonTypeInfoTest.java
```diff
@@ -0,0 +1,36 @@
...
+        assertEquals(BsonTypeInfo.BSON_MAXKEY, BsonTypeInfo.getBsonTypeInfoByName("maxKey"));
+        assertEquals(BsonTypeInfo.BSON_MINKEY, BsonTypeInfo.getBsonTypeInfoByName("minKey"));
+        assertEquals(BsonTypeInfo.BSON_NULL, BsonTypeInfo.getBsonTypeInfoByName("null"));
+        assertEquals(BsonTypeInfo.BSON_OBJECT, BsonTypeInfo.getBsonTypeInfoByName("object"));
+        assertEquals(BsonTypeInfo.BSON_OBJECTID, BsonTypeInfo.getBsonTypeInfoByName("objectId"));
+        assertEquals(BsonTypeInfo.BSON_REGEX, BsonTypeInfo.getBsonTypeInfoByName("regex"));
+        assertEquals(BsonTypeInfo.BSON_STRING, BsonTypeInfo.getBsonTypeInfoByName("string"));
+        assertEquals(BsonTypeInfo.BSON_SYMBOL, BsonTypeInfo.getBsonTypeInfoByName("symbol"));
+        assertEquals(BsonTypeInfo.BSON_TIMESTAMP, BsonTypeInfo.getBsonTypeInfoByName("timestamp"));
+        assertEquals(BsonTypeInfo.BSON_UNDEFINED, BsonTypeInfo.getBsonTypeInfoByName("undefined"));
```
bucaojit, line 34: Adding BSON_BSON test
```suggestion
        assertEquals(BsonTypeInfo.BSON_UNDEFINED, BsonTypeInfo.getBsonTypeInfoByName("undefined"));
        assertEquals(BsonTypeInfo.BSON_BSON, BsonTypeInfo.getBsonTypeInfoByName("bson"));
```

src/test/java/com/mongodb/jdbc/BsonTypeInfoTest.java
```diff
@@ -0,0 +1,36 @@
+package com.mongodb.jdbc;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.sql.SQLException;
+import org.junit.jupiter.api.Test;
+
+public class BsonTypeInfoTest {
+    @Test
+    void testGetBsonTypeInfoByName() throws SQLException {
```
bucaojit, line 10: A test with an unexpected bsonName would be good to check that it throws an exception.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -1,342 +1,230 @@
...
+                        return BSON_MAXKEY;
+                    case 'i':
+                        return BSON_MINKEY;
+                }
+                break;
+            case 'n':
+                return BSON_NULL;
+            case 'o':
+                switch (typeName.length()) {
+                    case 6: // "object"
```
bucaojit, line 205: Unneeded comment
```suggestion
                    case 6:
```

bucaojit (changes requested):
Great work with the refactor!  It really does make things clearer and helps having the bson type information in one place.

I agree with the comment on focusing on MongoSQL* classes as the MySQL dialect won't be around for long so no need to put effort into changes that will eventually be removed.  

There are tests failing however, I see a few failing in the MySQL code and looks related to the `bool` type that I commented on.

A few comments for now as I go through the rest of the changes.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -1,342 +1,230 @@
...
-                return Types.LONGVARCHAR;
-            case DECIMAL_TYPE_NAME:
-                return Types.DECIMAL;
-            case DOUBLE_TYPE_NAME:
-                return Types.DOUBLE;
+        // All type names can be determined uniquely off a combination of first letter
+        // and length except for "minKey" vs "maxKey" and "string" vs "symbol".
+        switch (typeName.charAt(0)) {
+            case 'a':
+                return BSON_ARRAY;
```
nbagnard, line 160: Just a note that this method might be a source of errors later on.
I suppose this is what we are going for this because it should be faster than a switch, correct?

If somehow I end up passing the typeName 'Apricot' then it will return BSON_ARRAY. I assume that since we control the calls to the method, it should not be an issue, but when adding new types, it can be easy to forget to update this function and then the new type might end up as an unexpected BSON type or a SQLException.

I am fine to keep it as is if you considered this and think that our tests will protect us well enough and this should not happen.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
mattChiaravalloti: The "bson" type isn't a real bson type so it was intentionally omitted from this list! I'll add a comment to the code explaining that.

This set is specifically used for the `getBsonTypeInfoByName` method which should fail on unknown type names. The interesting thing to note here is that we have this extra "bson type" called "bson" that is meant to represent "any". It is a construct that only exists in our codebase.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
mattChiaravalloti: Good catch! will update

src/test/java/com/mongodb/jdbc/BsonTypeInfoTest.java
mattChiaravalloti: Good idea, will add!

nbagnard (commented):
This looks pretty good 👍 
I have only a couple of things that you might want to look into and agress with all of Oliver's comments.
I am not sure about catching the SQLException and throwing a runtime exception unconditionally, because if not used with streams it can created havoc. Maybe we can talk about it with Oliver and see if we could find something.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
mattChiaravalloti: This method was just moved from one file to this one. It is meant to be used only with names provided in ADL collection schemas. Since those schemas are json schema validators, it is impossible for them to have invalid names; if they had invalid type names, ADL would have failed to create them (I suppose technically ADL could be buggy, but I don't think we should write code that handles ADL returning invalid information).

> I suppose this is what we are going for this because it should be faster than a switch, correct?

Yes, this is for speed! This is a common pattern I've seen in a couple repos on different teams that deal with bson parsing.

> but when adding new types, it can be easy to forget to update this function and then the new type might end up as an unexpected BSON type or a SQLException.

This is true in theory, but the introduction of a new BSON type would certainly result in a downstream ticket notification so we'd be aware of it before it is released and we'd be able to update this method appropriately.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -431,66 +433,40 @@ public ResultSet getCatalogs() throws SQLException {
...
             this.columnName = columnName;
-            this.columnSchema = columnSchema;
-            this.isRequired = isRequired;
             this.idx = idx;
+
+            try {
+                this.columnBsonTypeInfo = columnSchema.getBsonTypeInfo();
+                this.nullability = parentSchema.getColumnNullability(columnName);
+            } catch (SQLException e) {
+                throw new RuntimeException(e);
```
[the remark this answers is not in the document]
mattChiaravalloti: I will move this try-catch block to the stream code itself, instead of having it be part of this constructor.

For clarification though: do you want to reconsider the whole try-catch/`liftSQLException` pattern that is used throughout the file, or did you just want to address this particular try-catch block in this one method? If it's the former, I'll file a follow-up ticket. If it's the latter, which I'm assuming it is, I'll definitely make the change to move the try-catch to the stream code, leaving this constructor free to throw a SQLException.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
nbagnard: Sounds good to me 👍

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -76,4 +78,126 @@ public boolean isAny() {
...
+        if (columnSchema.anyOf == null) {
+            throw new SQLException(
+                    "invalid schema: both bsonType and anyOf are null and this is not ANY");
+        }
+
+        for (MongoJsonSchema anyOfSchema : columnSchema.anyOf) {
+            if (anyOfSchema.bsonType == null) {
+                // Schemata returned by MongoSQL must be simplified. Having nested anyOf is invalid.
+                throw new SQLException(
+                        "invalid schema: anyOf subschema must have bsonType field; nested anyOf must be simplified");
```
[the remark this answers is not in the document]
nbagnard: Yes, makes sense :) Thank you !

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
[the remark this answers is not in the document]
nbagnard: Sorry, I got fooled by the GitHub view and thought this block was in the getColumns method 🙃
Moving this try/catch block will just help with this one and is not of much value. I was looking at the broader pattern.
The try-catch/liftSQLException is our only solution for streams and it is fine. I am just concerned about missing methods which require using this pattern. 
Maybe the easiest thing we could do would be to mark methods which are throwing RuntimeException with a specific naming convention and adding a very clear comment in the method javadoc. I'll create a investigation task for this so that we can discuss and see what everyone in the team think about this.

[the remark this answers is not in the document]
mattChiaravalloti: One idea I have is to have every overridden method in this class use `liftSQLException` and an auxiliary helper method. This would look like:
```
...

@Override
public ResultSet getColumns(...) throws SQLException {
    return liftSQLException(() -> getColumnsAux(...));
}

private ResultSet getColumnsAux(...) {
    ...
    try {
       ...
   } catch (SQLException e) {
       throw new RuntimeException(e);
   }
   ...
}

@Override
public ResultSet getTables(...) throws SQLException {
    return liftSQLException(() -> getTablesAux(...));
}

private ResultSet getTablesAux(...) {
    ...
}

...
```

Benefits of this pattern:
- we don't need to rewrite all of the methods to avoid using streams
- we can freely catch-and-wrap any SQLExceptions in any `*Aux` methods or their helpers since we'll guarantee every public method uses `liftSQLException`
- consistency across all public methods

Drawbacks of this pattern:
- it will require some documentation in code to explain the pattern
- it will still require a refactor to move the current public method bodies into private `Aux` methods and updating the public method bodies to return `liftSQLException(() -> Aux(...))`
- there may be some performance impact having every public method dispatch an anonymous function that runs in a try-catch block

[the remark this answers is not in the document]
mattChiaravalloti: I'm sure there are also plenty of other alternatives

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
mattChiaravalloti: no changes required; see comment above

src/test/java/com/mongodb/jdbc/BsonTypeInfoTest.java
mattChiaravalloti: no test needed since this should throw an exception; see comment above. (I also added an invalid test case already, as requested in a different comment!)

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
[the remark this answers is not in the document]
mattChiaravalloti: filed [SQL-649](https://jira.mongodb.org/browse/SQL-649)!

mattChiaravalloti:
This is ready for another look. Note that some BSON_BSON-related comments resulted in no code changes, only new code comments! See the relevant threads on this PR for details.

nbagnard (approved):
Perfect! All good to go from my side :)

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -1,342 +1,235 @@
...
 import java.sql.SQLException;
 import java.sql.Types;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
+import org.bson.BsonType;
+
+public enum BsonTypeInfo {
+    BSON_DOUBLE("double", BsonType.DOUBLE, Types.DOUBLE, false, 15, 15, 2, 15, 15, 8),
+    BSON_STRING("string", BsonType.STRING, Types.LONGNVARCHAR, true, 0, 0, 0, null, null, null),
```
bucaojit, line 12: One of the tests are failing, it's expecting the type to be LONGVARCHAR.
```suggestion
    BSON_STRING("string", BsonType.STRING, Types.LONGVARCHAR, true, 0, 0, 0, null, null, null),
```

bucaojit (approved):
Looks good!  Thanks for making the suggested changes. 

One minor change with the string type needed to pass the test and it looks good to go.

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
mattChiaravalloti: oh boy great catch, thank you

PR #67, opened 2022-01-11
SQL-555: MongoSQL JDBC Integration tests with ADL for non-DatabaseMet…

bucaojit:
…aData methods

JIRA: https://jira.mongodb.org/browse/SQL-555

Adding queries to be run by the mongosql JDBC integration test.

Added change to expected formatVersion for mongosql. 
Added rows to dataset to return different rows for the different joins.
Many of the queries have a skip reason regarding the get namespace call returning an error: SQL-659 and SQL-658

### DIFF
--- a/resources/integration_test/testdata/integration.yml
+++ b/resources/integration_test/testdata/integration.yml
@@ -70,17 +70,18 @@ dataset:
     }
 
   - db: integration_test
-    collection: null_field
+    collection: null_and_missing
     docs:
-      - { _id: 0, "a": "a" }
+      - { _id: 0, "a": !!org.bson.BsonTimestamp 1}
       - { _id: 1, "a": null }
+      - { _id: 2 }
 
   - db: integration_test
-    collection: a_column_order
+    collection: a_non_lexicographic_field_order
     docs:
       - { _id: 4, "c": 7, "b": 6, "a":5, "A":1, "B": 2, "C": 3 }
 
   - db: integration_test
-    collection: b_column_order
+    collection: b_non_lexicographic_field_order
     docs:
-      - { _id: 11, "C": 10, "B": 9, "A": 8, "a": 12, "b": 13, "c": 14 }
+      - { "a": 9, _id: 8 }
--- a/resources/integration_test/testdata/integration.yml
+++ b/resources/integration_test/testdata/integration.yml
@@ -2,11 +2,11 @@ dataset:
   - db: integration_test
     collection: class
     docs:
-      - {_id: 0, studentid: 10329,  name: "John", enrolled: true}
-      - {_id: 1, studentid: 342,  name: "Jane", enrolled: false}
-      - {_id: 2, studentid: 303,  name: "Mike", enrolled: true}
-      - {_id: 3, studentid: 204323,  name: "Mary", enrolled: false}
-      - {_id: 4, studentid: 10,  name: "Pete", enrolled: false}
+      - {_id: 0, studentid: 10329,  name: "John", enrolled: true, startdate: 2000-01-01 }
+      - {_id: 1, studentid: 342,  name: "Jane", enrolled: false, startdate: 2000-02-01}
+      - {_id: 2, studentid: 303,  name: "Mike", enrolled: true, startdate: 2000-01-01}
+      - {_id: 3, studentid: 204323,  name: "Mary", enrolled: false, startdate: 2000-03-01}
+      - {_id: 4, studentid: 10,  name: "Pete", enrolled: false, startdate: 2000-01-01}
 
   - db: integration_test
     collection: grades
@@ -20,3 +20,67 @@ dataset:
       - {_id: 6, studentid: 342, testid: 5,  score: 80.1}
       - {_id: 7, studentid: 204323, testid: 5, score: 83.3}
       - {_id: 8, studentid: 200, testid: 5, score: 78.5}
+
+  - db: integration_test
+    collection: anyof_collection
+    docs:
+      - { _id: 0, a: 3 }
+      - { _id: 1, a: 3000000000 }
+      - { _id: 2, a: 4.5 }
+    schema: {
+              "bsonType":[
+                  "object"
+              ],
+              "properties":{
+                "a":{
+                  "bsonType":[
+                      "long",
+                      "double",
+                      "int"
+                  ]
+                },
+                "_id":{
+                  "bsonType":[
+                      "int"
+                  ]
+                }
+              }
+            }
+
+  - db: integration_test
+    collection: any_collection
+    docs:
+      - { _id: 0, b: 3 }
+      - { _id: 1, b: "b" }
+      - { _id: 2, b: 4.5 }
+    schema: {
+      "bsonType":[
+          "object"
+      ],
+      "required":[
+          "b"
+      ],
+      "properties":{
+        "_id":{
+          "bsonType":[
+              "int"
+          ]
+        }
+      }
+    }
+
+  - db: integration_test
+    collection: null_field
+    docs:
+      - { _id: 0, "a": "a" }
+      - { _id: 1, "a": null }
+
+  - db: integration_test
+    collection: a_column_order
+    docs:
+      - { _id: 4, "c": 7, "b": 6, "a":5, "A":1, "B": 2, "C": 3 }
+
+  - db: integration_test
+    collection: b_column_order
+    docs:
+      - { _id: 11, "C": 10, "B": 9, "A": 8, "a": 12, "b": 13, "c": 14 }
--- a/resources/integration_test/testdata/integration.yml
+++ b/resources/integration_test/testdata/integration.yml
@@ -6,6 +6,7 @@ dataset:
       - {_id: 1, studentid: 342,  name: "Jane", enrolled: false}
       - {_id: 2, studentid: 303,  name: "Mike", enrolled: true}
       - {_id: 3, studentid: 204323,  name: "Mary", enrolled: false}
+      - {_id: 4, studentid: 10,  name: "Pete", enrolled: false}
 
   - db: integration_test
     collection: grades
@@ -18,3 +19,4 @@ dataset:
       - {_id: 5, studentid: 10329, testid: 5,  score: 74.4}
       - {_id: 6, studentid: 342, testid: 5,  score: 80.1}
       - {_id: 7, studentid: 204323, testid: 5, score: 83.3}
+      - {_id: 8, studentid: 200, testid: 5, score: 78.5}
--- a/resources/integration_test/tests/query.yaml
+++ b/resources/integration_test/tests/query.yaml
@@ -1,0 +1,255 @@
+tests:
+  - description: select_star_from_grades_ordered
+    sql: SELECT * FROM grades ORDER BY _id
+    skip_reason: SQL-659
+
+  - description: select_star_from_grades_unordered
+    sql: SELECT * FROM grades
+    expected_result:
+      - [0, 84.5, 303, 3]
+      - [1, 97.4, 10329, 3]
+      - [2, 89.3, 342, 3]
+      - [3, 91.9, 204323, 3]
+      - [4, 87.5, 303, 5]
+      - [5, 74.4, 10329, 5]
+      - [6, 80.1, 342, 5]
+      - [7, 83.3, 204323, 5]
+      - [8, 78.5, 200, 5]
+    row_count: 9
+    ordered: false
+    expected_sql_type: [INTEGER, DOUBLE, INTEGER, INTEGER]
+    expected_catalog_name: ['', '', '', '']
+    expected_column_class_name: [int, double, int, int]
+    expected_column_label: [_id, score, studentid, testid]
+    expected_column_display_size: [10, 15, 10, 10]
+    expected_precision: [10, 15, 10, 10]
+    expected_scale: [0, 15, 0, 0]
+    expected_schema_name: ['', '', '', '']
+    expected_is_auto_increment: [false, false, false, false]
+    expected_is_case_sensitive: [false, false, false, false]
+    expected_is_currency: [false, false, false, false]
+    expected_is_definitely_writable: [false, false, false, false]
+    expected_is_nullable: [columnNullable, columnNullable, columnNullable, columnNullable]
+    expected_is_read_only: [true, true, true, true]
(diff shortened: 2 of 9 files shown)


resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,293 @@
...
+    expected_is_case_sensitive: [false, false, false, false]
+    expected_is_currency: [false, false, false, false]
+    expected_is_definitely_writable: [false, false, false, false]
+    expected_is_nullable: [columnNullable, columnNullable, columnNullable, columnNullable]
+    expected_is_read_only: [true, true, true, true]
+    expected_is_searchable: [true, true, true, true]
+    expected_is_signed: [true, true, true, true]
+    expected_is_writable: [false, false, false, false]
+
+  - description: select_star_from_class_ordered
```
mattChiaravalloti, line 41: What makes this test case distinct from the first test? What is this case going to cover that the first test case does not?

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,293 @@
...
+    db: integration_test
+    sql: SELECT * FROM class ORDER BY _id;
+    ordered: true
+    skip_reason: SQL-659
+
+  - description: select_star_from_class_unordered
```
mattChiaravalloti, line 47: Same question here: how is this distinct from the second test case? Is it that the return types are different?

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,293 @@
...
+    db: integration_test
+    sql: SELECT grades.score AS score FROM grades ORDER BY score ASC
+    ordered: true
+    skip_reason: SQL-659
+
+  - description: descending
+    db: integration_test
+    sql: SELECT grades.score AS score FROM grades ORDER BY score DESC
+    ordered: true
+    skip_reason: SQL-659
```
mattChiaravalloti, line 292: At first I wasn't sure if either of these tested any interesting JDBC-specific behavior. For the `ascending` test, I still think that since `ASC` is the default sort order. However, I suppose the `descending` test is interesting since it shows the driver doesn't reorder the result sets. Maybe we should drop the `ascending` test and rename the `descending` test `driver_does_not_alter_result_set_row_order`?

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,293 @@
...
+    db: integration_test
+    sql: SELECT avg(score) FROM grades where testid=5
+    ordered: false
+    skip_reason: SQL-659
+
+  - description: max
+    db: integration_test
+    sql: SELECT max(score) FROM grades
+    ordered: false
+    skip_reason: SQL-659
```
mattChiaravalloti, line 274: What JDBC-specific feature are we testing by having separate cases for these aggregation functions?

mattChiaravalloti (changes requested):
In general I'm wondering what the purpose of these tests are: are we trying to test JDBC specific behavior like how it handles polymorphic columns and other "interesting" column types (like object, array, etc), or are we just testing SQL features like the various different clauses? I feel like we should have a few more tests for the former and possibly a few less tests for the latter. Perhaps we should discuss more at standup/triage.

resources/integration_test/tests/query.yaml
bucaojit: I removed the extra select query tests and some aggregate function tests.
My initial thinking was to write tests for general SQL features to validate the expected data was present and run queries a user might run initially.  After discussing offline with Matt, the direction is to have queries that test the different aspects of JDBC such as column types.

resources/integration_test/tests/query.yaml
bucaojit: Removed

resources/integration_test/testdata/integration.yml
```diff
@@ -18,3 +19,68 @@ dataset:
...
       - {_id: 7, studentid: 204323, testid: 5, score: 83.3}
+      - {_id: 8, studentid: 200, testid: 5, score: 78.5}
+
+  - db: integration_test
+    collection: anyof_collection
+    docs:
+      - { _id: 0, a: 3 }
+      - { _id: 1, a: 3000000000 }
+      - { _id: 2, a: 4.5 }
+    schema: {
```
bucaojit, line 30: Added a way to set the schema to help with the any and anyof columns

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
```diff
@@ -94,6 +96,38 @@ public void dropCollections() {
...
+        schema.put(
+                "jsonSchema",
+                doc.toBsonDocument(BsonDocument.class, MongoClient.getDefaultCodecRegistry()));
+        try (MongoClient mongoClient = new MongoClient(adlUri)) {
+            MongoDatabase db = mongoClient.getDatabase(database);
+            db.runCommand(command);
+        }
+    }
+
+    private void generateSchemaOne(String database, String collection) {
```
bucaojit, line 116: I'll update the SQL-632 PR to use this generateSchemaOne method instead of generating for the database so that the set schemas don't get overwritten.

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
+tests:
+  - description: select_star_from_grades_ordered
+    db: integration_test
+    sql: SELECT * FROM grades ORDER BY _id
+    ordered: true
+    skip_reason: SQL-659
+
+  - description: select_star_from_class_unordered
```
mattChiaravalloti, line 8:
```suggestion
  - description: select_star_unordered
```
Let's name based on behavior instead of specific test features (in this case, that just means dropping the collection name)

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
+tests:
+  - description: select_star_from_grades_ordered
```
mattChiaravalloti, line 2:
```suggestion
  - description: select_star_ordered
```
See related comment on the following test.

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+    expected_column_label: [ _id, enrolled, name, startdate, studentid ]
+    expected_column_display_size: [ 10, 1, 0, 24, 10 ]
+    expected_precision: [ 10, 1, 0, 24, 10 ]
+    expected_scale: [ 0, 0, 0, 0, 0 ]
+    expected_schema_name: [ '', '', '', '', '' ]
+    expected_is_auto_increment: [ false, false, false, false, false ]
+    expected_is_case_sensitive: [ false, false, true, false, false ]
+    expected_is_currency: [ false, false, false, false, false ]
+    expected_is_definitely_writable: [ false, false, false, false, false ]
+    expected_is_nullable: [ columnNullable, columnNullable, columnNullable, columnNullable,
```
mattChiaravalloti, line 32: I'm surprised to see us report these as possibly nullable. I suppose this is the result of the fields being possibly missing. Just commenting on this -- no action needed for now!

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+                            columnNullable ]
+    expected_is_read_only: [ true, true, true, true, true ]
+    expected_is_searchable: [ true, true, true, true, true ]
+    expected_is_signed: [ true, false, false, false, true ]
+    expected_is_writable: [ false, false, false, false, false ]
+
+  - description: select_as_confirm_column_label
```
mattChiaravalloti, line 39: excellent idea!

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+    db: integration_test
+    sql: SELECT enrolled AS e, studentid AS s, name AS n from class
```
mattChiaravalloti, line 41:
```suggestion
    sql: SELECT enrolled AS e, studentid AS s, name AS n FROM class
```
[super-nit] looks like we've typically CAPS the SQL keywords but we missed "from" -> "FROM" here!

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+    ordered: false
+    skip_reason: SQL-658
+
+  - description: driver_does_not_alter_result_set_row_order
+    db: integration_test
+    sql: SELECT grades.score AS score FROM grades ORDER BY score DESC
+    ordered: true
+    skip_reason: SQL-659
+
+  - description: select_star_from_anyof
```
mattChiaravalloti, line 93: What query are we anticipating for this one?

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+    ordered: false
+    skip_reason: SQL-659
+
+  - description: subquery
+    db: integration_test
+    sql: SELECT studentid FROM (SELECT * FROM class WHERE enrolled = TRUE) AS foo
+    ordered: false
+    skip_reason: SQL-659
+
+  - description: count_and_group_by
```
mattChiaravalloti, line 81: I'm still not convinced this test is demonstrating any JDBC-specific behavior

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+    ordered: true
+    skip_reason: SQL-659
+
+  - description: select_star_from_anyof
+    db: integration_test
+    skip_reason: SQL-668
+
+  - description: select_star_from_any
+    db: integration_test
+    sql: SELECT * from any
```
mattChiaravalloti, line 99: I don't think we have an `any` collection above. We have `any_collection` and `anyof_collection`. Did you intend to use `any_collection` here or did you intend to create an `any` data set?

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+    skip_reason: SQL-696
+
+  - description: select_field_from_any
+    db: integration_test
+    sql: SELECT b from any_collection
+    expected_result:
+      - - !!org.bson.BsonInt32  3
+      - - !!org.bson.BsonString  b
+      - - !!org.bson.BsonDouble  4.5
```
mattChiaravalloti, line 108: This is super interesting! Is this working as expected? 😮

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+  - description: select_star_from_anyof
+    db: integration_test
+    skip_reason: SQL-668
+
+  - description: select_star_from_any
+    db: integration_test
+    sql: SELECT * from any
+    skip_reason: SQL-696
+
+  - description: select_field_from_any
```
mattChiaravalloti, line 102:
```suggestion
  - description: select_polymorphic_field
```
Alternative name suggestion

resources/integration_test/tests/query.yaml
```diff
@@ -0,0 +1,197 @@
...
+    db: integration_test
+    sql: SELECT nonexist FROM class
+    skip_reason: SQL-698
+
+  - description: select_null_field
+    db: integration_test
+    sql: SELECT a FROM null_field
+    expected_result:
+      - [a]
+      - [null]
```
mattChiaravalloti, line 140: We could consider combining these into one test:
```
# data.yml
...
    - db: db
      collection: null_and_missing
      data:
          - { _id: 0, a: "a" }
          - { _id: 1, a: null }
          - { _id: 2 }
...

# tests/query.yml
...
    - description: select_null_and_missing_field
      sql: SELECT a FROM null_and_missing
      expected_result:
        - [a]
        - [null]
        - [null]
...
```
This test would need the SQL-698 skip-reason then, but that's ok

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
mattChiaravalloti: If that means we won't be using `generateSchema()` anymore, we could refactor this signature to be `generateSchema(String database, String collection)` (basically just a rename).

mattChiaravalloti (changes requested):
Great updates! There are some minor nits and suggestions here and there but this is on a good track now with the updated tests.

resources/integration_test/testdata/integration.yml
```diff
@@ -18,3 +19,68 @@ dataset:
...
+      - { _id: 0, "a": "a" }
+      - { _id: 1, "a": null }
+
+  - db: integration_test
+    collection: a_column_order
+    docs:
+      - { _id: 4, "c": 7, "b": 6, "a":5, "A":1, "B": 2, "C": 3 }
+
+  - db: integration_test
+    collection: b_column_order
```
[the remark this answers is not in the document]
bucaojit: I added the table joins to test that the ordering is based on the datasource as well as the column name, `a_column_order.*` columns are sorted before `b_column_order.*`.  I think one change I can make is the second collection doesn't need that many columns.  Seems like a useful test, what do you think?

I'll change the collection names to the suggested name.

resources/integration_test/tests/query.yaml
bucaojit: I've removed it.

resources/integration_test/tests/query.yaml
bucaojit: Thanks for catching that!  Yes, it was meant to be `any_collection`. 👍

resources/integration_test/tests/query.yaml
bucaojit: Yes, it is interesting.  The type tags work for certain types, so far I can tell SnakeYaml can serialize types when it has a constructor that takes a parameter and the type can be casted to a string.  For a handful of them I wasn't able to get it to work, using the extended JSON for that did work and was a great suggestion.

resources/integration_test/tests/query.yaml
bucaojit: Ah thanks.  I missed adding the query, I'll add `sql: SELECT * FROM anyof_collection`

resources/integration_test/tests/query.yaml
bucaojit: Sounds good, combined the tests.

mattChiaravalloti (approved):
looks good to me! 💯 

Just need to get through this `skip_reason` bugs and we'll be that much closer to a partially operable jdbc driver 🎊 🔌 🚗

PR #68, opened 2022-01-12
SQL-563: Update integration tests and data loader to use db defined per test

mattChiaravalloti:
This PR updates the `TestEntry` model to include the `db` keyword and updates the integration test and data loader codepaths to use that test-defined db when establishing a connection.


bucaojit (approved):
Looks good!
Thanks for catching that.

mattChiaravalloti:
The errors on evergreen seem to be transient (502s from some external dependencies). I reran the tasks but they may still fail. I'll give it another rerun in an hour and then investigate further.

mattChiaravalloti:
confirmed transient. going to merge this so we can start branching off of it.

PR #69, opened 2022-01-19
SQL-632: Adding Types.OTHER support for JDBC integration test

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-632


src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
```diff
@@ -55,4 +80,68 @@ public void loadTestConfigs() throws IOException {
...
+        row.put("_id", objectID);
+        row.put("a", dbPointer);
+        row.put("b", javaScript);
+        row.put("c", new BsonMaxKey());
+        row.put("d", new BsonMinKey());
+        row.put("e", regularExpression);
+        row.put("f", symbol);
+        row.put("g", timestamp);
+        /*
+        BsonArray array = new BsonArray();
```
bucaojit, line 107: Troubleshooting the BsonArray type still, it is returning the error:
"failed getting result set schema: failed to convert BSON document to json_schema::Schema: data did not match any variant of untagged enum BsonType"

src/main/java/com/mongodb/jdbc/MongoStatement.java
```diff
@@ -19,7 +19,6 @@
     protected int fetchSize = 0;
     protected int maxQuerySec = 0;
     protected String currentDBName;
-    protected final BsonInt32 formatVersion = new BsonInt32(2);
```
bucaojit, line 22: Added change from the SQL-555 PR for the format version as it's needed to pass the evergreen tests.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
```diff
@@ -54,7 +54,7 @@ public DataLoader(String dataDirectory) throws IOException {
         readDataFiles(dataDirectory);
     }
 
-    private void generateSchema() {
+    public void generateSchema() {
```
bucaojit, line 57: Some changes to the dataloader class to load data and generate schema from the test case.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
bucaojit: Added skip reason for this: SQL-697
The problem looks related to how the arrays schema has an items field that mongosql-rs is expecting to be a single schema but is an array.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/models/TestEntry.java
```diff
@@ -13,6 +13,7 @@
     public Boolean ordered;
     public List<Object> expected_result;
     public List<String> expected_sql_type;
+    public List<String> expected_bson_type;
```
mattChiaravalloti, line 16: should we generate this row for all the already committed integration tests? Note that several of them have been manually edited so we shouldn't replace them wholesale. We should just generate and then add _this field_ to the tests.

mattChiaravalloti: It may also be fine to _not_ add this field to the existing tests since they don't specifically contain types that would lead to interesting results here.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
```diff
@@ -55,4 +77,70 @@ public void loadTestConfigs() throws IOException {
         }
         return dynamicTests;
     }
+
+    @Test
+    public void testTypesOther() throws SQLException, IOException, IllegalAccessException {
+        DataLoader loader = new DataLoader(TEST_DATA_OTHER_DIRECTORY);
+        loader.dropCollections();
```
mattChiaravalloti, line 84: Why do we need to drop collections at the start of this test?

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
```diff
@@ -55,4 +77,70 @@ public void loadTestConfigs() throws IOException {
...
+
+        Map<String, Object> row = new LinkedHashMap<>();
+        ObjectId objID = new ObjectId("000000000000000000000001");
+        BsonObjectId objectID = new BsonObjectId(objID);
+        BsonDbPointer dbPointer = new BsonDbPointer("namespace", objID);
+        BsonJavaScript javaScript = new BsonJavaScript("function(){ }");
+        BsonRegularExpression regularExpression = new BsonRegularExpression("a(bc)*");
+        BsonSymbol symbol = new BsonSymbol("symbol");
+        BsonTimestamp timestamp = new BsonTimestamp(100);
+
```
mattChiaravalloti, line 94: We also use `Types.OTHER` for the bson types `object`, `undefined`, `symbol`, and `javascriptWithScope`. I know the latter 3 are deprecated but I believe they should still be available in the driver so we may still want to test them. At minimum, we should test an object.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
```diff
@@ -55,4 +77,70 @@ public void loadTestConfigs() throws IOException {
...
+        row.put("_id", objectID);
+        row.put("a", dbPointer);
```
mattChiaravalloti, line 96: perhaps more useful names could make this easier at a glance? "dbPointer", "javaScript", and so on

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
```diff
@@ -55,4 +77,70 @@ public void loadTestConfigs() throws IOException {
...
+        // Skip Reason - SQL-697
+        // Getting result set schema throwing error: data did not match any variant of untagged enum BsonType
+        BsonArray array = new BsonArray();
+        array.add(new BsonInt32(1));
+        array.add(new BsonInt32(2));
+        array.add(new BsonInt32(3));
+        row.put("h", array);
+         */
+
+        loader.loadTestRow("integration_test", "types_other", row);
```
mattChiaravalloti, line 113: Why aren't we writing this data directly in the yaml file? We may be able to write the data using extended JSON syntax and then have the java driver interpret the extended json info into appropriate bson types.

mattChiaravalloti (changes requested):
Generally looking good! I think we should look into getting the data to live in the yaml file so we could use the same test style for this stuff as we do for all other integration tests.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
bucaojit: I added that in case the types_other collection already exists from running this test previously on an existing local ADL, in this case it drops only the types_other collection as right now that's the only collection in the TEST_DATA_OTHER_DIRECTORY directory.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
bucaojit: Sounds good.  `symbol` is there currently.  I missed `javascriptWithScope`, I'll add that.

I'm having trouble with `object`, I am inserting a `BasicBSONObject` and I get back a `BsonDocument` so the comparison fails.  I'll see if I can cast it somehow.
The `undefined` type is throwing this error: `'failed getting result set schema: failed to add JSON schema for collection integration_test.types_other to the catalog: is not a valid BSON type`.  I'll check in the mongosql-rs if there's a bug.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
bucaojit: Some of the types were not able to be serialized using the type tags.  BsonObjectID, BsonDbPointer, BsonMaxKey, were some of the types I was having issues with.  Snakeyaml was throwing an exception that either the object could not be casted to a string, or there wasn't a suitable constructor.  I did not try the extended JSON syntax, I will try it out.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
mattChiaravalloti: - Ah, I missed the `symbol`. 
- For object, I think putting in a BsonDocument as the input type is fine since that is also logically an object
- Oh yea, we have a bug ticket in mongosql-rs to include the undefined type ([SQL-395](https://jira.mongodb.org/browse/SQL-395))

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
mattChiaravalloti: for extended json syntax, what I'm picturing is roughly:
```
# data.yml
...
    - db: db1
      collection: other_types
      data:
          - { oid: { $oid: "..." }, symbol: { $symbol: "..." }, ... }
...

# dataloader.java
...
    Document d = Document.parse(new Document(row).toJson());
    collection.insertOne(d);
...
```

I haven't tested this yet.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
mattChiaravalloti: Ok I see. This may become a non-issue if we manage to work out the ideas in other comments (re: using extended json to write this data in yaml data files like the other data).

bucaojit: I've removed the dataloader calls here in testTypesOther().  As you mention, it was worked out by using the extended json to load the data.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
bucaojit: Added `object` and `javascriptWithScope`.  For `undefined` I added the bug ticket in a comment.

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
bucaojit: Added change to process extended json if the docsJson field is not null.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/models/TestEntry.java
bucaojit: I added the additional `expected_bson_type` field.  There were `string`, `int`, `bool`, `long` types.

resources/integration_test/testdata/integration.yml
```diff
@@ -18,3 +18,25 @@ dataset:
...
+        objectId: { "$oid": "000000000000000000000001" },
+        regularExpression: {"$regularExpression": {pattern: "a(bc)*", "options": ""}},
+        symbol: { "$symbol": "symbol" },
+        timestamp: {"$timestamp": {"t":100, "i":0}}
+
+        # Skip reason: SQL-395
+        # undefined: {"$undefined":true},
+        # Skip reason: SQL-697
+        # array: [1,2,3]
+      }
```
mattChiaravalloti, line 42:
```suggestion
      }

```
[nit] newlines at end of files

src/integration-test/java/com/mongodb/jdbc/integration/testharness/models/TestDataEntry.java
```diff
@@ -7,5 +7,6 @@
     public String db;
     public String collection;
     public List<Map<String, Object>> docs;
+    public List<Map<String, Object>> docsJson;
```
mattChiaravalloti, line 10: Maybe we should document in the yaml, in this file, or in a README what these fields are and how they are used. Just a sentence each saying "`docs` uses standard JSON to represent collection data. This is useful for simple types such as int and string. `docsJson` uses extended JSON to represent collection data. This is useful for complex types such as binData and objectid."

Also another idea: maybe we should name this `docsExtJson`?

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
```diff
@@ -55,4 +78,91 @@ public void loadTestConfigs() throws IOException {
         }
         return dynamicTests;
     }
+
+    @Test
+    public void testTypesOther() throws SQLException, IOException, IllegalAccessException {
```
mattChiaravalloti, line 83: Now that we can load this type of data using a yaml file in the testdata directory, why do we have a manual test written out? Can't this test be written as a yaml test in the tests directory?

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
```diff
@@ -99,6 +99,21 @@ public void dropCollections() {
         }
     }
 
+    public void loadTestRow(String db, String collection, Map<String, Object> row) {
```
mattChiaravalloti, line 102: I don't see this being used anywhere anymore -- is it?

mattChiaravalloti (changes requested):
Super close -- lmk what you think about having a yaml test for the types.other stuff vs. a test fully written in java. I could see it going either way!

src/integration-test/java/com/mongodb/jdbc/integration/testharness/models/TestDataEntry.java
bucaojit: Added a section to the README with this information and renamed field to `docsExtJson`

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
bucaojit: Removed the method

src/integration-test/java/com/mongodb/jdbc/integration/MongoSQLIntegrationTest.java
bucaojit: I made the changes to process the document from the yaml file.  I was having trouble with getting the row to fit in document form.  I set the key as the index of the column and add it to the List<Object> to compare with the actual row.  That seems to work fine.

mattChiaravalloti (changes requested):
just a few more small typos and nits to address!

resources/integration_test/tests/dbmd_variable_result_sets_function_methods.yml
```diff
@@ -37,6 +37,7 @@ tests:
     row_count: 19
     expected_sql_type: [LONGVARCHAR, LONGVARCHAR, LONGVARCHAR, LONGVARCHAR, INTEGER,
                         LONGVARCHAR]
+    expected_bson_type: [string, string, string, int, string, string]
```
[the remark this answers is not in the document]
bucaojit: I see what happened.  After the change was merged that orders the metadata resultset columns, SQL-703, I regenerated the tests but missed the ones with a skip_reason since the generator also skips them.  I have fixed these now.
Thanks for catching that.

resources/integration_test/tests/query.yaml
```diff
@@ -180,4 +184,58 @@ tests:
     expected_is_writable: [false, false, false, false, false, false, false, false,
                            false]
 
+  - description: query_types_other
+    db: integration_test
+    sql: SELECT * FROM types_other
+    expected_result_extended_json: [ {
+      "0": 0,
```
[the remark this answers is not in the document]
bucaojit: There may be another way to do this.  I wanted to iterate through and add the items to a List to fit the format for compareRow() and this helps.

mattChiaravalloti (approved):
looks good 👍

PR #70, opened 2022-01-28
SQL-703: Order database metadata resultset columns following the column order in the spec

nbagnard:


src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
```diff
@@ -20,7 +20,7 @@
 
     static {
         try {
-            resultSetMetaData = new MongoSQLResultSetMetaData(generateMongoJsonSchema());
+            resultSetMetaData = new MongoSQLResultSetMetaData(generateMongoJsonSchema(), true);
```
mattChiaravalloti, line 23: How did you determine `true` vs `false` here? Should we test both via unit testing?

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -52,7 +52,7 @@
     static {
         try {
             schema = generateMongoJsonSchema();
-            resultSetMetaData = new MongoSQLResultSetMetaData(schema);
+            resultSetMetaData = new MongoSQLResultSetMetaData(schema, false);
```
mattChiaravalloti, line 55: How did you determine `false` vs `true` here? For this one, I don't see us using `resultSetMetaData` anywhere in the tests, so should we remove it? Or is its construction implicitly testing something?

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
```diff
@@ -50,12 +50,15 @@ private void assertDatasourceSchema(MongoJsonSchema schema) throws SQLException
         }
     }
 
-    private void processDataSource(MongoJsonSchema schema, String datasource) throws SQLException {
+    private void processDataSource(MongoJsonSchema schema, String datasource, boolean isDatabaseMetadataResultSet) throws SQLException {
```
mattChiaravalloti, line 53:
```suggestion
    private void processDataSource(MongoJsonSchema schema, String datasource, boolean sortFieldsAlphabetically) throws SQLException {
```
[bike-shedding-nit] Should we give a more general-purpose name to this parameter and use it as a positive indicator instead of a negative one (meaning we could say `if (param)` instead of `if (!param)`).

We're using this parameter to determine whether or not columns should be sorted alphabetically. Although the only use case right now is to prevent result set metadata for `DatabaseMetaData` methods from being sorted, there may be other use cases in the future. For example, maybe we'll eventually allow users to specify column order on their queries.

There may never be a use case like that, so this is probably just name bike-shedding. Up to you if you want to consider this!

mattChiaravalloti (changes requested):
Glad to see this one knocked 🔥  Just a few prelim thoughts for now. I still need to look through the tests more thoroughly, but that'll have to wait until next week!

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
nbagnard: Maybe I just misunderstood what this is for. I checked the name and the functions and thought it was for validation resultset metadata for database metadata. I'll double-check on Monday.

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
nbagnard: Sounds like a good idea to me. Will make the change.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
```diff
@@ -380,8 +396,9 @@ private static void validateRowCount(
 
     private static boolean compareRow(List<Object> expectedRow, ResultSet actualRow)
             throws SQLException {
+        boolean result = true;
         ResultSetMetaData rsMetadata = actualRow.getMetaData();
-        assertEquals(expectedRow.size(), rsMetadata.getColumnCount());
+        assertEquals(expectedRow.size(), rsMetadata.getColumnCount(), "Either the yml test specification is missing entries for 'expected_result' or the data in the DB is not matching the test");
```
bucaojit, line 401: Nit: This assert is regarding a column count mismatch between the expected row and the metadata.  
I'd reword it to something related to the column count, maybe: "Column count mismatch between 'expected_result' in the yml test specification and ResultSetMetaData in the DB"

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
```diff
@@ -20,7 +20,7 @@
 
     static {
         try {
-            resultSetMetaData = new MongoSQLResultSetMetaData(generateMongoJsonSchema());
```
bucaojit, line 23: It may be useful to add a unit test here to confirm the expected sort order with isDatabaseMetadataResultSet set and unset.  It's similar to the integration tests, but more direct as opposed to going through the DatabaseMetaData or Statement calls.

bucaojit (changes requested):

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
nbagnard: You are right, both classes are related to validating resultset from "queries", not from database metdata result sets. Used "true" for both.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
```diff
@@ -209,117 +209,133 @@ private static void validateResultSetMetadata(TestEntry test, ResultSetMetaData
             throws SQLException, IllegalAccessException {
         int columnCount = rsMetaData.getColumnCount();
         if (test.expected_sql_type != null) {
-            assertEquals(test.expected_sql_type.size(), columnCount);
+            assertEquals(test.expected_sql_type.size(), columnCount, "Either the yml test specification is missing entries for 'expected_sql_type' or the data in the DB is not matching the test");
```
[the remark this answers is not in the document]
nbagnard: np

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
nbagnard: I'm sorry but I don't quite see what you are proposing here. Can you provide more details?
MongoDatabaseMetaDataTest tests Database metadata resultsets and will verify that the order is not alphabetical but follow the order of the spec. This class will verify that result sets columns for SQL statements are sorted alphabetically. Since, the decision to set or unset is not "our" but is based on the result set type, it seems enough and less confusing to only test the expected behaviour.
Or are you asking for a new test to check that the result set metadata columns are sorted alphabetically? I just checked and it does not seem to be here. This would be a good test to add in this file.

bucaojit: Yes, the latter.  A test that confirms the sorting of the columns that we get from generateMongoJsonSchema() are what we are expecting.

rychipman:
I actually going to swap myself out for @pmeredit as a reviewer -- I'm realizing that my plate is a lot more full than I realized, and I don't want to become a blocker! Sorry for the bait-and-switch!

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
```diff
@@ -209,117 +209,133 @@ private static void validateResultSetMetadata(TestEntry test, ResultSetMetaData
...
-                assertEquals(sqlType, rsMetaData.getColumnType(i + 1));
+                assertEquals(sqlType, rsMetaData.getColumnType(i + 1), "Invalid getColumnType result for column " + i+1);
             }
         }
         if (test.expected_catalog_name != null) {
-            assertEquals(test.expected_catalog_name.size(), columnCount);
+            assertEquals(test.expected_catalog_name.size(), columnCount, "Either the yml test specification is missing entries for 'expected_catalog_name' or the data in the DB is not matching the test");
             for (int i = 0; i < columnCount; i++) {
-                assertEquals(test.expected_catalog_name.get(i), rsMetaData.getCatalogName(i + 1));
+                assertEquals(test.expected_catalog_name.get(i), rsMetaData.getCatalogName(i + 1),"Invalid getCatalogName result for column " + i+1);
```
mattChiaravalloti, line 221:
```suggestion
                assertEquals(test.expected_catalog_name.get(i), rsMetaData.getCatalogName(i + 1), "Invalid getCatalogName result for column " + i+1);
```
I really appreciate all of these test messages -- great idea! [nit] on this one: spacing

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
```diff
@@ -50,12 +50,15 @@ private void assertDatasourceSchema(MongoJsonSchema schema) throws SQLException
...
     }
 
-    private void processDataSource(MongoJsonSchema schema, String datasource) throws SQLException {
+    private void processDataSource(MongoJsonSchema schema, String datasource, boolean sortFieldsAlphabetically) throws SQLException {
         MongoJsonSchema datasourceSchema = schema.properties.get(datasource);
         assertDatasourceSchema(datasourceSchema);
 
         String[] fields = datasourceSchema.properties.keySet().toArray(new String[0]);
-        Arrays.sort(fields);
+        // [SQL-703] Only sort the result columns alphabetically for non database metadata resultsets.
```
mattChiaravalloti, line 58:
```suggestion
        // Only sort the result columns alphabetically for non database metadata resultsets.
```
2 things about this comment:
1. If we include it, I'm not sure we should/need to include the ticket number.
2. I'm not sure we need to include here in-line. Perhaps as a method-level comment explaining what the parameter is, and examples of why/when it is used would be better?

mattChiaravalloti (approved):
Looks good besides 2 small nits/thoughts: 💯 Whatever you decide for those will be okay with me.

It must have been such a headache going through all those generated tests and making the updates 😓 thanks for picking this one up 🙏

src/main/java/com/mongodb/jdbc/MongoSQLResultSetMetaData.java
nbagnard: ah, I think the issue here is that the comment would be better MongoSQLResultset. It shows why there is a flag and what value is used when. I like to flag small changes with big implications with the JIRA id, but I think with the updated comment it is not be necessary to keep it.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
nbagnard: Yes, I think it makes sense. Even though, technically, this is tested because we have tests checking the database metadata order and tests checking SQL statement result sets order.
Added a unit test for this.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
nbagnard: Splotless moved it down the next line. Not an issue anymore :)

bucaojit (approved):
Looks good!  
Thanks for adding the test

pmeredit (approved):
LGTM!

nbagnard:
evergreen merge

PR #71, opened 2022-02-03
SQL-672: If type is 'Collection', translate to 'Table'

nbagnard:


src/main/java/com/mongodb/jdbc/MongoListCollectionsResult.java
```diff
@@ -1,6 +1,16 @@
 package com.mongodb.jdbc;
 
 public class MongoListCollectionsResult {
+    public static final String TABLE = "table";
```
bucaojit, line 4: In the spec I see the table keyword is usually uppercase, is there a reason why we are setting it as lowercase here?

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
```diff
@@ -553,11 +578,15 @@ private static boolean compareRow(List<Object> expectedRow, ResultSet actualRow)
...
                     case Types.OTHER:
                         if (expectedRow.get(i) != actualRow.getObject(i + 1)) {
+                            System.err.println("Expected " + expected + " but is " + actual);
                             return false;
                         }
                      */
                 case Types.OTHER:
+                    Object expected_obj = expectedRow.get(i);
+                    Object actual_obj = expectedRow.get(i);
                     if (!expectedRow.get(i).equals(actualRow.getObject(i + 1))) {
```
bucaojit, line 588: To follow your previous edits, and change to get the actual_obj:
```suggestion
                    Object actual_obj = actualRow.getObject(i + 1);
                    if (!expected_obj.equals(actual_obj)) {
```

resources/integration_test/tests/dbmd_empty_result_sets.yml
```diff
@@ -5,6 +5,7 @@ tests:
     meta_function: [getProcedures, integration_test, null, '%']
     expected_result: []
     row_count: 0
+    ordered: true
```
bucaojit, line 8: nit: Up to you whether to remove, these tests return 0 rows and the `ordered` field would have no effect and wouldn't be needed.

bucaojit (changes requested):
Great work on updating the tests.  And thanks again for the log messages.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
nbagnard: oh, thank you for spotting this. Definitely a mistake, I forgot to update after asking you if this was supposed to be this way or not.

resources/integration_test/tests/dbmd_empty_result_sets.yml
nbagnard: Yes, I know, but I want to keep the marker, so that the day we add procedures support, even if it is very unlikely we'll ever do, then things are "ready"

src/main/java/com/mongodb/jdbc/MongoListCollectionsResult.java
bucaojit: Sounds good, thanks for the explanation

bucaojit (approved):
LGTM!

src/main/java/com/mongodb/jdbc/MongoListCollectionsResult.java
```diff
@@ -1,6 +1,16 @@
...
+    public static final String COLLECTION = "collection";
+
     public String name;
     public String type;
+
+    public void setType(String type) {
+        // If mongodb type is collection, map it as TABLE.
+        // Otherwise, keep the type as is.
+        this.type = type.equalsIgnoreCase(COLLECTION) ? TABLE : type;
```
rychipman, line 13: Given that this class is supposed to represent a collection as described by MongoDB's `listCollections`, it feels a little funny for `type` to take the value `TABLE`, which isn't a concept that exists in MongoDB. I'd suggest the following as an alternative, but happy to discuss further if this isn't idiomatic/desirable for some reason (as I'm far from a java/jdbc expert):

- Leave the `type` value as it is returned from the `listCollections` command
- Add a method (perhaps `getTableType` or similar) that returns the JDBC table type for this collection

rychipman (changes requested):
one question/suggestion from me!

src/main/java/com/mongodb/jdbc/MongoListCollectionsResult.java
nbagnard: I agree it's weird and initially we wanted to report 'collection' as is proper. We have to report `table` for `collection` because Tableau thinks that every DB must at least have tables, and it is very likely not the only BI tool to make this assumption 🙃 
I did this here because it is the safest place. It will make sure that there is no place where someone forget to translate.
What I can do is make `type` private and only provide access to `tableType`, where I would do the translation. Or I can keep MongoListCollection as it was before and create MongoListTables where type is the 'jdbc' type. 
While I understand that it might look funny to a Mongo addict, I don't really see the value though of having an extra variable or getter or child class, it is more likely to make things more confusing.

nbagnard: As discussed in stand-up, renamed the class to `MongoListTablesResult` to avoid confusion and show that we are exposing the JDBC concept here.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
```diff
@@ -553,11 +578,15 @@ private static boolean compareRow(List<Object> expectedRow, ResultSet actualRow)
                     /*
```
varsha804, line 578: I'm very unfamiliar with the jdbc rep so I'm most likely missing something, but should this commented section be removed given lines 585-589? If not, [super nit] should we at least modify the commented code so it uses the
```
Object expected_obj = expectedRow.get(i);
Object actual_obj = actualRow.getObject(i + 1);
```
pattern the other sections use?

src/main/java/com/mongodb/jdbc/MongoListTablesResult.java
```diff
@@ -0,0 +1,16 @@
...
+
+public class MongoListTablesResult {
+    public static final String TABLE = "table";
+    public static final String COLLECTION = "collection";
+
+    public String name;
+    public String type;
+
+    public void setType(String type) {
+        // If mongodb type is collection, map it as TABLE.
```
varsha804, line 11: [nit]
```suggestion
        // If mongodb type is COLLECTION, map it as TABLE.
```

varsha804 (changes requested):
Just a couple nits!

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
nbagnard: ahah, don't worry, I had a similar question. So, this is/will be addressed by Oliver in one of his PR that's why I left it here.

src/main/java/com/mongodb/jdbc/MongoListTablesResult.java
nbagnard: Done :)

varsha804 (approved):
lgtm!

PR #72, opened 2022-02-03
SQL-667: getFunctionColumns fails with a NPE

nbagnard:


src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -866,7 +866,7 @@ public static Integer typeScale(String typeName) {
             case "decimal":
                 return 34;
         }
```
pmeredit, line 868: Since we can no longer return null here, I think we might as well return `int` instead of `Integer`. Saves us some GC.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -891,7 +891,7 @@ public static Integer typeBytes(String typeName) {
             case "bool":
                 return 1;
         }
-        return null;
```
pmeredit, line 894: Same

pmeredit (approved):
Looks great. Just one small request, feel free not to retag me :)

bucaojit (approved):
LGTM

pmeredit:
@nbagnard I noticed we'll need to overload `bsonInt32` for my idea to work, but I think it's worth it.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
nbagnard: Sometime, null is actually a valid value. It will handled correctly when SQL-710 is implemented.
I'm leaving the return type as an Object for now, just so that we won't forget that it can be null.
But I update the method callings these helpers to use bsonint32 (which can handle a null value) instead of new BsonInt32 (which can't handle a null value).

PR #73, opened 2022-02-05
SQL-668: bsontype can be an array or a single String.

nbagnard:
This PR handles decoding schema where bsontype is either a Set<String> or String.
It also implements a logic for transforming sets of types in bsontype in a list of anyOf and flattening anyOfs.

The assumptions I made for this are:
- `{
    "bsonType": "bool",
    "additionalProperties": false
}`
is not equals to 
`{
    "bsonType": "bool",
    "additionalProperties": true
}`
so the same type can be twice in the anyOf list if the other fields values are different.
- If bsonType is a set of types, there will be no duplicates. For example, no `bsontype: [string, object, bool, object]`


src/main/java/com/mongodb/jdbc/JsonSchema.java
```diff
@@ -0,0 +1,64 @@
...
+import org.bson.BsonValue;
+import org.bson.codecs.pojo.annotations.BsonIgnore;
+
+import static java.lang.String.format;
+
+// Simple POJO for deserializing jsonschema.
+public class JsonSchema {
+    private BsonValue polymorphicBsonType;
+    @BsonIgnore
+    public Set<String> bsonType;
```
rychipman, line 22:
```suggestion
    public BsonValue bsonType;
```
I like the idea of having this `JsonSchema` class (for use as a deserialization target) as something separate from `MongoJsonSchema` (for using everywhere else in the API). I'd suggest that we go even further towards just making this class just a "dumb" deserialization target:
- keep the raw BsonValue for the `bsonType` field
- put whatever array/string polymophism-handling logic we need in the function that converts this to a `MongoJsonSchema`

rychipman (changes requested):
only one relatively minor request from me

pmeredit:
> This PR handles decoding schema where bsontype is either a Set or String. It also implements a logic for transforming sets of types in bsontype in a list of anyOf and flattening anyOfs.
> 
> The assumptions I made for this are:
> 
> * `{ "bsonType": "bool", "additionalProperties": false }`
>   is not equals to
>   `{ "bsonType": "bool", "additionalProperties": true }`
>   so the same type can be twice in the anyOf list if the other fields values are different.

It's still correct to have the same type twice, so I think this is a good simplifying assumption.

> * If bsonType is a set of types, there will be no duplicates. For example, no `bsontype: [string, object, bool, object]`

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -4,22 +4,196 @@
...
+            for (JsonSchema baseAnyOf : baseSchema.anyOf) {
+                result.anyOf.add(toSimplifiedMongoJsonSchema(baseAnyOf));
+            }
+        }
+        result.required = baseSchema.required;
+        result.items = toSimplifiedMongoJsonSchema(baseSchema.items);
+        result.additionalProperties = baseSchema.additionalProperties;
+
+        if (baseSchema.bsonType!= null) {
+            Set<String> bsonTypes = polymorphicBsonTypeToStringSet(baseSchema.bsonType);
```
pmeredit, line 72: I love this method name.

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -4,22 +4,196 @@
...
+            //  If there are many types in the set and it can not be reduced to one type after eliminating any Null
+            //  type in the list, the types will be inserted in the list  of anyOf to be handled as polymorphic type.
```
pmeredit, line 74:
```suggestion
            //  type in the list, the types will be inserted in the list of anyOf to be handled as polymorphic type.
```

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -4,22 +4,196 @@
...
+                }
+            }
+
+            // We'll need to add the types from bsontype set to the anyOf list.
+            // If the list is null, create a new one.
+            if (null == result.anyOf) {
+                result.anyOf = new HashSet<MongoJsonSchema>();
+            }
+
+            // Push down each bsontype into its own anyOf schema
```
pmeredit, line 102: Could you add an example here in the comment to clarify what this mean?

pmeredit (changes requested):
I still need to go over the latter half of this. The first half looks good, just a few small requests.

pmeredit:
As a small comment, could we avoid removing new lines from the end of all the files? It messes up the PR. If that is something we want to do, we should do it in a separate PR. But I assume this was some IDE setting :)

src/main/java/com/mongodb/jdbc/JsonSchema.java
```diff
@@ -0,0 +1,36 @@
...
+    public JsonSchema items;
+    public boolean additionalProperties;
+
+    @Override
+    public boolean equals(Object obj) {
+        if (!(obj instanceof MongoJsonSchema)) {
+            return false;
+        }
+        MongoJsonSchema other = (MongoJsonSchema) obj;
+        return Objects.equals(bsonType, other.bsonType)
```
pmeredit, line 24: Is `equals` defined for `BsonValue`?

pmeredit: Yes, it is!

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -4,22 +4,216 @@
...
+             * "y": {
+             *     "bsonType": ["string", "int"]
+             *  }
+             *  will become
+             *  "y": {
+             *      "anyOf": [
+             *          {"bsonType": "string"},
+             *          {"bsonType": "int"}
+             *      ]
+             *  }
```
pmeredit, line 114: perfect!

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -4,22 +4,216 @@
...
+                }
+            }
+
+            // We'll need to add the types from bsontype set to the anyOf list.
+            // If the list is null, create a new one.
+            if (null == result.anyOf) {
+                result.anyOf = new HashSet<MongoJsonSchema>();
+            }
+
+            /** Push down each bsontype into its own anyOf schema.
```
pmeredit, line 103:
```suggestion
            /** Push down each bsontype from a bsontype array into a separate bsonType in an anyOf schema.
```

I think my problem with this comment before is "its own anyOf" which seems like we would have one `anyOf` per `bsonType`.

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -4,22 +4,216 @@
...
+        if (null == ioSchema) {
+            return null;
+        }
+
+        MongoJsonSchema result = ioSchema;
+        if (result.anyOf != null && !result.anyOf.isEmpty()) {
+            result.anyOf =
+                    result.anyOf
+                            .stream()
+                            .flatMap(
```
pmeredit, line 191: <3 `flatMap`. Combinators just make code so much easier to follow once you learn them.

pmeredit (approved):
LGTM! Feel free to apply the comment suggestion or not.

rychipman (approved):
LGTM

PR #74, opened 2022-02-08
SQL-698: Return Null when the bsonvalue is Null

nbagnard:


src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -214,21 +223,27 @@ protected Object getObject(BsonValue o, int columnType) throws SQLException {
     @Override
     public Object getObject(int columnIndex) throws SQLException {
         BsonValue out = getBsonValue(columnIndex);
+        // If the value is Null, no need to try to convert to a Java Object
```
pmeredit, line 226: This check is unnecessary since we are calling `getObject(value, columnType)` which also checks it.

pmeredit (approved):
LGTM, just a small suggestion

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
nbagnard: Yes, it's only here to saves us a call rsMetaData.getColumnType which is useless if the data is null.

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -78,6 +78,13 @@ protected BsonValue getBsonValue(String columnLabel) throws SQLException {
 
     @Override
     protected Object getObject(BsonValue o, int columnType) throws SQLException {
+        // The type of the Java object will be the default Java object type corresponding to the column's SQL type (see
+        // "TABLE B-3 Mapping from JDBC Types to Java Object Types" in the JDBC specification), following the mapping
+        // for built-in types specified in the JDBC specification.
+        // If the value is an SQL NULL, the driver returns a Java null.
+        if (checkNull(o)) {
```
bucaojit, line 85: The BsonUndefined type returns true for checkNull() for some reason.  With that being the case, it doesn't seem like we would want to return null here and in the other getObject() methods.  We may need to get more context on how to handle this.

bucaojit (changes requested):

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
nbagnard: Yes, I just noticed the failure in the unit tests. I am not sure what is the behavior we want to go for.
I can see why we'd want getObject to return a value, but at the same time, calling getTime (for example) on the same cell will return Null. It seems strange to sometime return Null, sometime not, for the same cell value. A cell should either be SQL NULL or not.

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
nbagnard: I went back and checked the code for getColumnType. It's calling a few getters on already set variables, so should not be too expensive. Seems reasonable to leave checkNull for `getObject(value, columnType)` as you pointed out :)

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
nbagnard: Changed the test to expect Null for an "undefined" value as discussed this morning.

bucaojit (approved):
Looks good!

PR #75, opened 2022-02-08
SQL-721: Handle null value in getTables() for tableNamePattern

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-721

When tableNamePattern is null then table name will not be used to narrow the search.  Added test for getTables with tableNamePattern as `null`.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
@@ -138,7 +138,9 @@ public abstract class MongoDatabaseMetaData implements DatabaseMetaData {
     }
 
     public static Pattern toJavaPattern(String sqlPattern) {
-        return Pattern.compile(sqlPattern.replaceAll("%", ".*").replaceAll("_", "."));
+        return sqlPattern == null
+                ? null
+                : Pattern.compile(sqlPattern.replaceAll("%", ".*").replaceAll("_", "."));
     }
 
     // Actual max size is 16777216, we reserve 216 for other bits of encoding,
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -138,7 +138,9 @@ public static String escapeString(String value) {
     }
 
     public static Pattern toJavaPattern(String sqlPattern) {
-        return Pattern.compile(sqlPattern.replaceAll("%", ".*").replaceAll("_", "."));
+        return sqlPattern == null
+                ? null
```
pmeredit, line 142: [nit] I'd slightly prefer creating a ".*" pattern here and then getting rid of the null check on 309 because I think `null` is equivalent to ".*" in all of the meta functions.

pmeredit (approved):
LGTM, consider the comment, but do whatever you feel is best.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
nbagnard: @pmeredit : Creating a ".*" pattern makes things look easier, but in term of performance I think Oliver's solution is better. No need to compile the pattern and check if each table name matches if sqlPattern is null, we know that there is no need for filtering.
@bucaojit : I think we should also protect ourselves in getColumnsFromDB for the possibility of both table name and column name to be null. It looks like Tableau does not use null when calling getColums, but if someone does, we will also crash with a NPE.

nbagnard (changes requested):
We should also protect ourselves in getColumnsFromDB for the possibility of both table name and column name to be null.

nbagnard (approved):
LGTM! 👍

PR #76, opened 2022-02-09
Helpers for running manual checks

nbagnard:
Helpers to quickly test something out and print the Resultset (and metadata if wanted) in a format similar to :
```
+----------------------|----------------------|----------------------|----------------------+
| _id                  | a                    | b                    | xyz                  |
+----------------------|----------------------|----------------------|----------------------+
| 0                    | 1                    | true                 | hello                |
| 1                    | 2                    | null                 | null                 |
+----------------------|----------------------|----------------------|----------------------+
```

### DIFF
--- a/demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
+++ b/demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
@@ -4,6 +4,7 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 
 public class PrintUtils {
@@ -18,7 +19,7 @@ public class PrintUtils {
             StringBuilder sb = new StringBuilder();
             List<String> colNames = new ArrayList<String>();
             colNames.add("TABLE_CAT");
-            colNames.add("TABLE_SCHEM");
+            colNames.add("TABLE_SCHEMA");
             colNames.add("TABLE_NAME");
             colNames.add("COLUMN_NAME");
             colNames.add("DATA_TYPE");
@@ -112,15 +113,17 @@ public class PrintUtils {
     private static void printRsContents(StringBuilder sb, int columnCount, ResultSet rs, int[] maxColsWidth) throws SQLException
     {
         String data = null;
+        String cellValue;
         while (rs.next()) {
             String[] row = new String[columnCount];
+
             boolean hasMoreData;
             do {
                 hasMoreData = false;
                 for (int i = 0; i < columnCount; i++) {
                     if (row[i] == null) {
-                        row[i] = rs.getString(i + 1);
-                        if (row[i] == null ) {row[i] = NULL_STR;}
+                        cellValue = rs.getString(i + 1);
+                        row[i] = (cellValue == null ? NULL_STR : cellValue);
                     }
                     int cutIndex = Math.min(row[i].length(), maxColsWidth[i]);
                     data = row[i].substring(0, cutIndex);
@@ -144,7 +147,7 @@ public class PrintUtils {
     private static int[] printRsHeader(StringBuilder sb, int columnCount, ResultSetMetaData rsMeta) throws SQLException
     {
         String[] col_names = new String[columnCount];
-        int[] maxColWidth = new int[columnCount];;
+        int[] maxColWidth = new int[columnCount];
         for (int i = 0; i < columnCount; i++)
         {
             maxColWidth[i] = MAX_COL_WIDTH;
--- a/demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
+++ b/demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
@@ -93,7 +93,7 @@ public class PrintUtils {
             StringBuilder sb = new StringBuilder();
 
             int[] maxColsWidth = printRsHeader(sb, columnCount, rsMeta);
-            printRsContent(sb, columnCount, rs, maxColsWidth);
+            printRsContents(sb, columnCount, rs, maxColsWidth);
 
             System.out.println(sb.toString());
         }
@@ -109,7 +109,7 @@ public class PrintUtils {
         return "| %-" + maxColWidth + "s ";
     }
 
-    private static void printRsContent(StringBuilder sb, int columnCount, ResultSet rs, int[] maxColsWidth) throws SQLException
+    private static void printRsContents(StringBuilder sb, int columnCount, ResultSet rs, int[] maxColsWidth) throws SQLException
     {
         String data = null;
         while (rs.next()) {
--- a/demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
+++ b/demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
@@ -12,8 +12,7 @@ public class PrintUtils {
     private static int MAX_RS_META_COL_WIDTH = 12;
     public static int MAX_COL_WIDTH = 20;
 
-    public static void printResultSetMetadata(ResultSetMetaData rsMeta)  throws SQLException
-    {
+    public static void printResultSetMetadata(ResultSetMetaData rsMeta)  throws SQLException {
         try {
             int columnCount = rsMeta.getColumnCount();
             StringBuilder sb = new StringBuilder();
@@ -31,8 +30,7 @@ public class PrintUtils {
             colNames.add("LABEL");
 
             printRowSeparator(sb, colNames.size(), MAX_RS_META_COL_WIDTH);
-            for (int i = 0; i < colNames.size(); i++)
-            {
+            for (int i = 0; i < colNames.size(); i++) {
                 sb.append(String.format(getFormat(MAX_RS_META_COL_WIDTH), colNames.get(i)));
             }
             sb.append("|\n");
@@ -75,43 +73,35 @@ public class PrintUtils {
 
             System.out.println(sb.toString());
         }
-        catch (Exception e)
-        {
+        catch (Exception e) {
             e.printStackTrace();
             throw e;
         }
     }
 
-    public static void printResultSet(ResultSet rs) throws SQLException
-    {
+    public static void printResultSet(ResultSet rs) throws SQLException {
         try {
             // Get the metadata of this rs.
             ResultSetMetaData rsMeta;
             rsMeta = rs.getMetaData();
-
             int columnCount = rsMeta.getColumnCount();
-
             StringBuilder sb = new StringBuilder();
-
             int[] maxColsWidth = printRsHeader(sb, columnCount, rsMeta);
             printRsContents(sb, columnCount, rs, maxColsWidth);
 
             System.out.println(sb.toString());
         }
-        catch (Exception e)
-        {
+        catch (Exception e) {
             e.printStackTrace();
             throw e;
         }
     }
 
-    private static String getFormat(int maxColWidth) throws SQLException
-    {
+    private static String getFormat(int maxColWidth) {
         return "| %-" + maxColWidth + "s ";
     }
 
-    private static void printRsContents(StringBuilder sb, int columnCount, ResultSet rs, int[] maxColsWidth) throws SQLException
-    {
+    private static void printRsContents(StringBuilder sb, int columnCount, ResultSet rs, int[] maxColsWidth) throws SQLException {
         String data = null;
         String cellValue;
         while (rs.next()) {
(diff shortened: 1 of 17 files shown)


demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
```diff
@@ -0,0 +1,188 @@
...
+        while (rs.next()) {
+            String[] row = new String[columnCount];
+            boolean hasMoreData;
+            do {
+                hasMoreData = false;
+                for (int i = 0; i < columnCount; i++) {
+                    if (row[i] == null) {
+                        row[i] = rs.getString(i + 1);
+                        if (row[i] == null ) {row[i] = NULL_STR;}
+                    }
```
varsha804, line 124: I think we can use a ternary here
```suggestion
                    if (row[i] == null) {
                        String rowValue = rs.getString(i + 1);
                        row[i] = rowValue == null ? NULL_STR : rowValue;
                    }
```

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    @AfterEach
+    private void execDeconnection() throws Exception {
+        m_conn.close();
+    }
+
+    @Test
+    void testConnection() throws Exception {
+
```
varsha804, line 32: Extra newline?
```suggestion
```

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+    void testGetCatalogs() throws Exception
+    {
+        ResultSet rs = m_dbMeta.getCatalogs();
+        printRsInfo(rs);
+    }
+
+    @Test
+    void testGetTables() throws Exception
+    {
+        //ResultSet rs = m_dbMeta.getTables(null, null, null, null);
```
varsha804, line 61: What's the purpose of this comment and the ones below it?

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+        printRsInfo(rs);
+    }
+
+    @Test
+    void testDBMeta() throws Exception {
+        System.out
+                .println("-----------------------------------------------------------------------");
+        System.out
+                .println("-----------------------------------------------------------------------");
+        System.out.println("API databases matadata test call:");
```
varsha804, line 250:
```suggestion
        System.out.println("API databases metadata test call:");
```

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+        System.out.println("Driver Name: " + m_dbMeta.getDriverName());
+        System.out.println("Driver Version: " + m_dbMeta.getDriverVersion());
+        System.out.println("Driver Major Version: "
+                + m_dbMeta.getDriverMajorVersion());
+        System.out.println("Driver Minor Version: "
+                + m_dbMeta.getDriverMinorVersion());
+        System.out.println("Server product Name: "
```
varsha804, line 257: [nit] Can we make the capitalization more consistent across all of the strings in this code section?

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+
+    @Test
+    void testSetCatalog() throws SQLException
+    {
+        m_conn.setCatalog("test_set_catalog");
+        System.out.println(m_conn.getCatalog());
+    }
+
+    @Test
+    // Note : Cancel is not supported yet
```
varsha804, line 312: Is there a JIRA ticket we can link here?

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+        }
+        if (execThread.isAlive())
+        {
+            try
+            {
+                // Still need to wait because the m_isExecuting flag is set
+                // before
+                // statement.executeQuery is called. Need to wait until an
+                // executor
+                // object is created.
```
varsha804, line 381: The newline placement looks a little odd here
```suggestion
                // Still need to wait because the m_isExecuting flag is set
                // before statement.executeQuery is called. Need to wait until
                // an executor object is created.
```

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+            e.printStackTrace();
+            throw e;
+        }
+    }
+
+    private void printRsInfo(ResultSet rs) throws Exception
+    {
+        try
+        {
+            //PrintUtils.printResultSetMetadata(rs.getMetaData());
```
varsha804, line 430: Similar to my comment in `TestUtils.java`, but was this accidentally commented out?

varsha804 (changes requested):
Mostly just a bunch of nits!

demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
```diff
@@ -0,0 +1,188 @@
...
+                sb.append("|\n");
+            } while (hasMoreData);
+        }
+        printRowSeparator(sb, columnCount, maxColsWidth);
+    }
+
+    private static int[] printRsHeader(StringBuilder sb, int columnCount, ResultSetMetaData rsMeta) throws SQLException
+    {
+        String[] col_names = new String[columnCount];
+        int[] maxColWidth = new int[columnCount];;
```
bucaojit, line 147: extra semicolon
```suggestion
        int[] maxColWidth = new int[columnCount];
```

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+        else
+        {
+            System.out.println("The execution finished before cancel");
+            queryExecutor.getResultSet().close();
+        }
+
+        System.out.println("-----------------------------------------------------------------------");
+    }
+
+    private void runExecuteQuery(String query, boolean printRs) throws Exception
```
bucaojit, line 407: I think it would be really helpful if there was a command run through gradle to easily run queries such as:
```
./gradlew runQuery "select * from foo"
```
And see the results and metadata printed.  Totally up to you if you want to implement, seems doable with the logic in this function.

bucaojit (changes requested):
Awesome!  Thanks for implementing this, it really helps with seeing the query results and metadata.

It looks like in order to run the tests, the `exclude` in the build.gradle has to be commented/removed.  Let's add a brief description in the readme on how to run these helpers.

demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
varsha804: I figured a ternary would be better because it seems a little odd to me that we're checking `if (row[i] == null)` twice within two lines of one another. Ternaries are also idiomatic in Java, right? If you really think it makes things hard to read, then we can keep it as is, but I do think the current code looks a little awkward.

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
varsha804: Ah, I see. Can you add an explanatory comment for readability?

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
varsha804: I think it should be enough to just add a `TODO` here so we can easily search for cases that need more investigation

demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
```diff
@@ -0,0 +1,188 @@
...
+    public static int MAX_COL_WIDTH = 20;
+
+    public static void printResultSetMetadata(ResultSetMetaData rsMeta)  throws SQLException
+    {
+        try {
+            int columnCount = rsMeta.getColumnCount();
+            StringBuilder sb = new StringBuilder();
+            List<String> colNames = new ArrayList<String>();
+            colNames.add("TABLE_CAT");
+            colNames.add("TABLE_SCHEM");
```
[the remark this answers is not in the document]
varsha804: looks like this still needs to be updated!

varsha804 (changes requested):
looking good, just a couple nits remaining!

demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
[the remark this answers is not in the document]
mattChiaravalloti: I guess it depends on the use case, but some result sets have the column `TABLE_SCHEM` (no `A` at the end) ([example](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-)).

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+    void testConnection() throws Exception {
+        System.out.println(m_dbMeta.getDriverName());
+        System.out.println(m_dbMeta.getDriverVersion());
+
+        System.out.println("Connected !");
+    }
+
+    @Test
+    void testExecute() throws Exception
+    {
```
mattChiaravalloti, line 40: This may be addressed by running the formatter, but it looks like our braces are inconsistent in this file/PR. In the other files, the braces are in-line (i.e. ` public void method() {`), but for this method and most that follow, the braces are on their own lines (i.e.
```
public void method()
{
```
). Can we stick with the former convention for the sake of stylistic consistency? Ideally, just running the formatter will do this for us by enforcing one convention over the other. I'm indifferent which direction it goes, but think consistency is useful.

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
mattChiaravalloti, line 407:
```suggestion
    // Run executeQuery and print the result set metadata. Optionally print the result set data by arguing true as the second parameter.
    private void runExecuteQuery(String query, boolean printRs) throws Exception
```
A briefly comment explaining what this does would be helpful. Especially if we follow-up on Oliver's idea!

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
```diff
@@ -0,0 +1,439 @@
...
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public abstract class TestUtils {
```
mattChiaravalloti, line 12: I'm unsure the purpose of this file. The `PrintUtils` class alone is very helpful going forward as it gives building blocks for anyone to easily print result sets and metadata during development/testing. This file, however, has unclear utility. The "test" methods aren't actually testing anything -- they are just printing data  for various `DatabaseMetaData` methods.

Whenever there is a specific issue someone wants to investigate, they'll need to alter the "tests" here so they could print what they need to print. Are these meant to just be reference methods that people alter as needed? What is the expected future use of this file and the test methods in it?

mattChiaravalloti (changes requested):

mattChiaravalloti:
I likely missed a lot of context surrounding this PR but am curious what we intend to do with this code. See my last thread for my specific concerns. Also, is there a dedicated ticket associated with this or is it a spin-off from another PR/ticket, or is it just untracked side work?

demo/src/test/java/com/mongodb/jdbc/utils/PrintUtils.java
nbagnard: You seem to really like to have a ternary here. If this makes you happy, then it's Ok for me too. I can live with it. It's not a big deal either way. For me, reading the if statement is easier for me than the "duplicated" rowValue and having to backtrack on the ==. Plus, the simple if is also actually what we want to do => only replace the value of row[i] if it's null.
But ternary operator are wonderful for replacing simple if/else. Things like `max = (n1 > n2) ? n1 : n2`

demo/src/test/java/com/mongodb/jdbc/utils/TestUtils.java
nbagnard: I also decided to rename the helpers from testXX to tryXX in hope that it differentiates them from actual unit tests.

mattChiaravalloti (approved):
looks good pending our idea of a .gitignore followup! 💯

bucaojit (approved):
Looks good, this will be a useful tool for testing.  Thanks for adding it!

varsha804 (approved):
lgtm!

PR #77, opened 2022-02-09
SQL-669 : Report correct nullability for database metadata result sets

nbagnard:


src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -48,16 +66,18 @@ public static MongoJsonSchema createScalarSchema(String type) {
      * @return void
      */
     @SafeVarargs
-    public final void addRequiredScalarKeys(Pair<String, String>... scalarProperties) {
+    public final void addRequiredScalarKeys(ScalarProperties... scalarProperties) {
```
bucaojit, line 69: The comments describing this method need an update with the parameter change.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -604,30 +604,30 @@ public ResultSet getColumns(
...
-                        new Pair<>(CHAR_OCTET_LENGTH, BSON_INT.getBsonName()),
-                        new Pair<>(ORDINAL_POSITION, BSON_INT.getBsonName()),
-                        new Pair<>(IS_NULLABLE, BSON_STRING.getBsonName()),
-                        new Pair<>(SCOPE_CATALOG, BSON_STRING.getBsonName()),
-                        new Pair<>(SCOPE_SCHEMA, BSON_STRING.getBsonName()),
-                        new Pair<>(SCOPE_TABLE, BSON_STRING.getBsonName()),
-                        new Pair<>(SOURCE_DATA_TYPE, BSON_INT.getBsonName()),
-                        new Pair<>(IS_AUTOINCREMENT, BSON_STRING.getBsonName()),
-                        new Pair<>(IS_GENERATEDCOLUMN, BSON_STRING.getBsonName()));
+                        new MongoJsonSchema.ScalarProperties(TABLE_CAT, BSON_STRING),
```
bucaojit, line 607: Looks like setting of nullability was missed for these properties in getColumns

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -924,14 +925,15 @@ public ResultSet getVersionColumns(String catalog, String schema, String table)
...
                 createBottomSchema(
-                        new Pair<>(SCOPE, BSON_STRING.getBsonName()),
-                        new Pair<>(COLUMN_NAME, BSON_STRING.getBsonName()),
-                        new Pair<>(DATA_TYPE, BSON_INT.getBsonName()),
-                        new Pair<>(TYPE_NAME, BSON_STRING.getBsonName()),
-                        new Pair<>(COLUMN_SIZE, BSON_INT.getBsonName()),
-                        new Pair<>(BUFFER_LENGTH, BSON_INT.getBsonName()),
-                        new Pair<>(DECIMAL_DIGITS, BSON_INT.getBsonName()),
-                        new Pair<>(PSEUDO_COLUMN, BSON_INT.getBsonName()));
+                        new MongoJsonSchema.ScalarProperties(SCOPE, BSON_STRING, false),
```
bucaojit, line 928: In this document, the SCOPE is a type `short` and doesn't state the column is nullable:
https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getVersionColumns-java.lang.String-java.lang.String-java.lang.String-

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -1866,24 +1869,23 @@ public ResultSet getFunctions(String catalog, String schemaPattern, String funct
...
+                        new MongoJsonSchema.ScalarProperties(FUNCTION_CAT, BSON_STRING, false),
+                        new MongoJsonSchema.ScalarProperties(FUNCTION_SCHEM, BSON_STRING, false),
+                        new MongoJsonSchema.ScalarProperties(FUNCTION_NAME, BSON_STRING),
+                        new MongoJsonSchema.ScalarProperties(COLUMN_NAME, BSON_STRING),
+                        new MongoJsonSchema.ScalarProperties(COLUMN_TYPE, BSON_INT),
+                        new MongoJsonSchema.ScalarProperties(DATA_TYPE, BSON_INT),
+                        new MongoJsonSchema.ScalarProperties(TYPE_NAME, BSON_STRING),
+                        new MongoJsonSchema.ScalarProperties(PRECISION, BSON_INT),
+                        new MongoJsonSchema.ScalarProperties(LENGTH, BSON_INT),
+                        new MongoJsonSchema.ScalarProperties(SCALE, BSON_INT),
```
bucaojit, line 1881: SCALE can return [null](https://docs.oracle.com/javase/8/docs/api/java/sql/DatabaseMetaData.html#getFunctionColumns-java.lang.String-java.lang.String-java.lang.String-java.lang.String-)

bucaojit (changes requested):
Adding the ScalarProperties class really improves the readability 👍 

A few comments on the nullability columns.

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
nbagnard: That's very true, the javadoc and method name are totally out-of-date :D
Done.

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
nbagnard: I did indeed, sorry for that.
I fixed that now, but since we can't test it until SQL-668 is merged, I think it's better wait and update the integration tests when SQL-668 is done. I'll create a JIRA to track that. Not only is_nullable but all metadata validation must be added to the integration tests.

bucaojit (approved):
LGTM!  Thanks for making the suggested changes.

pmeredit (approved):
Nice, very good refactor with `addScalar`!

PR #78, opened 2022-02-11
SQL-664 : Download and install local ADL under local_adl directory

nbagnard:


resources/run_adl.sh
```diff
@@ -184,6 +190,7 @@ if [[ $? -ne 0 ]]; then
     export LIBRARY_PATH="$(pwd)/artifacts"
 
     go mod download
+    go mod vendor
```
bucaojit, line 193: The vendor step isn't needed here.  Check in your environment if you have `GOFLAGS="-mod=vendor"` in your environment.  If you unset that then the script will work without the `go mod vendor` line.

bucaojit (approved):
Looks good, one comment about the `go mod vendor` step.

resources/run_adl.sh
nbagnard: oh yes, that's right.
Removed :)

nbagnard:
@bucaojit Do you mind giving it a quick last check. I merged the change from the previous PR for using a local Mongohouse that was merged last week. I just want to make sure I did not miss something.

resources/run_adl.sh
```diff
@@ -224,6 +226,7 @@ if [[ $? -ne 0 ]]; then
     STORES='{ "name" : "localmongo", "provider" : "mongodb", "uri" : "mongodb://localhost:%s" }'
     STORES=$(printf "$STORES" "${MONGOD_PORT}")
     DATABASES=$(cat $DB_CONFIG_PATH)
+    TENANT_CONFIG="$MONGOHOUSE_DIR/testdata/config/mongodb_local/tenant-config.json"
```
bucaojit, line 229: This TENANT_CONFIG variable is also used in the else statement, line 270.  
The logic to set the variable will either need to be added there.

bucaojit (changes requested):

bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #81, opened 2022-02-23
SQL-710: Use BsonTypeInfo and remove standalone type infos values

nbagnard:


src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -76,12 +76,27 @@
...
     private final int maxScale;
+    // Radix (typically either 10 or 2) for numeric data.
     private final int numPrecRadix;
+    // The column/data size for the given type.
+    //  - For numeric data, this is the maximum precision.
+    //  - For character data, this is the maximum length in characters.
+    //  - For datetime data types, this is the length in characters of the String representation (assuming the maximum
+    //  allowed precision of the fractional seconds component).
+    //  - For binary data, this is the maximum length in bytes.
+    //  - For the ROWID datatype, this is the length in bytes.
```
mattChiaravalloti, line 91: What is the ROWID BSON type? Do you mean ObjectId?

src/main/java/com/mongodb/jdbc/MongoSQLDatabaseMetaData.java
```diff
@@ -1892,29 +1885,34 @@ private BsonDocument getFunctionColumnValuesDoc(
             int i,
             String argName,
             String argType,
-            boolean isReturnColumn) {
+            boolean isReturnColumn)
+            throws SQLException {
+
+        BsonTypeInfo bsonTypeInfo =
+                argType == null ? BSON_NULL : BsonTypeInfo.getBsonTypeInfoByName(argType);
```
mattChiaravalloti, line 1892: Should we be using the new helper method from the parent class, `getFunctionParameterValues`? This seems to mostly be the same code, if not exactly the same code.

mattChiaravalloti (changes requested):
Largely looks great! 🔥  Just two small comments before lgtm-ing

bucaojit (changes requested):
Nice work, a few small comments.

src/main/java/com/mongodb/jdbc/MongoSQLFunctions.java
```diff
@@ -35,90 +35,114 @@ public static MongoSQLFunctions getInstance() {
...
                                         "LOWER",
-                                        "string",
+                                        BsonTypeInfo.BSON_STRING.getBsonName(),
                                         "returns the provided string with all characters changed to lowercase.",
-                                        new String[] {"string"}),
+                                        new String[] {BsonTypeInfo.BSON_STRING.getBsonName()}),
                                 new MongoFunction(
                                         NULLIF,
-                                        null,
+                                        BsonTypeInfo.BSON_NULL.getBsonName(),
```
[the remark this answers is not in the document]
nbagnard: ah, good catch. 
null is for parameters or return type which are not "set" and can be anything.
BSON_NULL is specifically for a Null value.
Now, unfortunately as you notice in the end, the differentiation disappears and null is mapped to java.sql.Types 0 ('null').
I'm wondering if I should default null to OTHER instead of NULL in the resultset 🤔 None of the options are totally perfect. We would need a type which maps "ANY" and there is nothing like that in java.sql.Types. OTHER seems to be the closest though.
WDYT?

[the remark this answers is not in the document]
bucaojit: Yes, OTHER does seem like a closer return type to use.  Sounds good.

mattChiaravalloti (approved):

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -406,7 +406,7 @@ void testGetSchemas() throws SQLException {
     @Override
     void testGetFunctions() throws SQLException {
         // All function(s)
-        testGetFunctionsHelper("%", 19);
+        testGetFunctionsHelper("%", 15);
         // All function(s) with a 'S'
         testGetFunctionsHelper("%S%", 9);
```
bucaojit, line 411:
```suggestion
        testGetFunctionsHelper("%S%", 6);
```
Small update to the functions that have an `S`

bucaojit (approved):
Looks Good!

One small change in the test to account for the removed functions.

PR #82, opened 2022-03-07
SQL-739: Add logging capabilities

nbagnard:
Adding logging capabilities.
Below are examples of the produced logs.

- Public method entries
```
[2022-03-07 16:17:47.491] [FINER] [c-1] com.mongodb.jdbc.MongoSQLConnection: >> createStatement()  
[2022-03-07 16:17:47.498] [FINER] [c-1] [stmt-1] com.mongodb.jdbc.MongoSQLStatement: >> executeQuery(***)  
[2022-03-07 16:17:47.939] [FINER] [c-1] com.mongodb.jdbc.MongoSQLConnection: >> getMetaData()  
[2022-03-07 16:17:47.953] [FINER] [c-1] com.mongodb.jdbc.MongoSQLConnection: >> createStatement()  
[2022-03-07 16:17:47.954] [FINER] [c-1] [stmt-2] com.mongodb.jdbc.MongoSQLStatement: >> executeQuery(***)
[2022-03-07 16:17:48.085] [FINER] [c-1] [stmt-2] com.mongodb.jdbc.MongoResultSet: >> next()  
[2022-03-07 16:17:48.085] [FINER] [c-1] [stmt-2] com.mongodb.jdbc.MongoResultSet: >> getString(1)
```

- Error
```
[2022-03-07 16:17:48.427] [SEVERE] [c-3] [stmt-2] com.mongodb.jdbc.MongoSQLStatement: Error in MongoSQLStatement.executeQuery(..) 
com.mongodb.MongoCommandException: Command failed with error 1 (InternalError): 'failed getting namespaces: parse error: Unrecognized token `Another` found at 0:7
Expected one of SELECT, correlationID = 16da3fb6de1eaa583c68306f' on server localhost:27017. The full response is {"ok": 0, "errmsg": "failed getting namespaces: parse error: Unrecognized token `Another` found at 0:7\nExpected one of SELECT, correlationID = 16da3fb6de1eaa583c68306f", "code": 1, "codeName": "InternalError"}
	at com.mongodb.internal.connection.ProtocolHelper.getCommandFailureException(ProtocolHelper.java:175)
	at com.mongodb.internal.connection.InternalStreamConnection.receiveCommandMessageResponse(InternalStreamConnection.java:302)
	at com.mongodb.internal.connection.InternalStreamConnection.sendAndReceive(InternalStreamConnection.java:258)
```


pmeredit:
So, I feel like it's late to be asking this sort of thing, but would it be too hard to change to json formatted logs? It's what most of the company is doing at this point.

build.gradle
```diff
@@ -51,24 +52,52 @@ allprojects {
...
-
-
         // Test
         testImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: junitJupiterVersion
         testRuntimeOnly  group: 'org.junit.jupiter', name: 'junit-jupiter-engine', version: junitJupiterVersion
         testCompile  group: 'org.mockito', name: 'mockito-core', version: mockitoVersion
         testCompile  group: 'org.mockito', name: 'mockito-junit-jupiter', version: mockitoVersion
     }
 
+    ant.lifecycleLogLevel = AntBuilder.AntMessagePriority.VERBOSE
```
bucaojit, line 82: When I run a build of this PR I'm getting a TON of messages that seem to be related to ant, I suspect it's related to this line being VERBOSE?  If so can we change this, the log lines I'm seeing look like the following:
```
[ant:iajc] Join point 'method-execution(java.lang.Object com.mongodb.jdbc.MongoConnection.unwrap(java.lang.Class))' in Type 'com.mongodb.jdbc.MongoConnection' (MongoConnection.java:517) advised by afterThrowing advice from 'com.mongodb.jdbc.logging.LoggingAspect' (LoggingAspect.aj:72) [with runtime test]
```

bucaojit (changes requested):
Great work adding the logging capabilities!

A few comments for now.

build.gradle
bucaojit: That works, thanks!

src/integration-test/java/com/mongodb/jdbc/integration/MongoIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Creates a new connection.
+     *
+     * @param extraProps Extra properties on top of the default ones that the class implementating
+     *     it is using.
+     * @return The connection.
+     * @throws SQLException If the connection can not be created.
+     */
+    public abstract Connection getBasicConnection(Properties extraProps) throws SQLException;
```
pmeredit, line 29: Why not just make all of these `Connection`s `MongoConnection`s so that we don't need the down casts in the `cleanup`? It will require casting in the overriden child classes, but it's 2 casts versus 3 :)

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -235,4 +244,225 @@ void testDialectProperty() throws SQLException {
...
+    @Test
+    void testInvalidCustomLogDir() throws Exception {
+        // Set the custom log dir to an invalid path
+        // Connection is not successful.
+        Properties props = new Properties();
+        File specialLogDir = new File(new File("."), "ThisIsNotAValidPath");
+        props.setProperty(MongoDriver.LOG_DIR, specialLogDir.getAbsolutePath());
+        try {
+            MongoConnection conn = createConnectionAndVerifyLogFileExists(props);
+            fail("Expected to fail because the logging director does not exist.");
```
bucaojit, line 379:
```suggestion
            fail("Expected to fail because the logging directory does not exist.");
```

src/integration-test/java/com/mongodb/jdbc/integration/MongoIntegrationTest.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /**
+     * Connect with the given logging level.
+     *
+     * @param logLevel The log level or null if not logging.
+     * @return the connection.
+     * @throws SQLException If an error occurs during the connection process.
+     */
+    private Connection connect(Level logLevel) throws SQLException {
```
bucaojit, line 92: If I understand this test correctly, with the latest changes we now have to set a LogDir  for the logger to use to create the log file in.   Here without a directory it is not creating the connection_*.log file to test that concurrent file creation works.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -133,10 +148,39 @@ private MongoConnection createConnection(ConnectionString cs, Properties info)
             throws SQLException {
         // attempt to get DIALECT property, and default to "mysql" if none is present
         String dialect = info.getProperty(DIALECT, MYSQL_DIALECT);
+        // Default log level is OFF
+        String logLevelVal = info.getProperty(LOG_LEVEL, Level.OFF.getName());
```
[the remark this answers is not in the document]
pmeredit: In the rest of the code base we use `toLowerCase`. It's a small thing, but I'd rather be consistent (I don't care which one we choose!).

src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
```diff
@@ -186,14 +205,14 @@ public boolean isSigned(int column) throws SQLException {
     public int getColumnDisplaySize(int column) throws SQLException {
         MongoColumnInfo ci = getColumnInfo(column);
         if (ci.isPolymorphic()) {
-            return unknownLength;
+            return UNKNOWN_LENGTH;
```
pmeredit, line 208: Good change, I don't know why I made this lowercase.

src/main/java/com/mongodb/jdbc/logging/LoggingAspect.aj
```diff
@@ -0,0 +1,73 @@
...
+import java.util.Properties;
+import java.util.logging.Level;
+import org.aspectj.lang.JoinPoint;
+import org.bson.BsonValue;
+
+public aspect LoggingAspect perthis(execution(com.mongodb.jdbc.*.new(..)))
+{
+    private MongoLogger logger = null;
+
+    pointcut setLogger(MongoLogger logger): set(MongoLogger *.*) && args(logger) && !within(LoggingAspect) ;
```
pmeredit, line 15: Nice on the `!within`. Accidentally instrumenting the instrumentation code is a common footgun that bites people. Never want to see infinite instrumentation overflows ;)

src/main/java/com/mongodb/jdbc/logging/MongoLoggerUtils.java
```diff
@@ -0,0 +1,95 @@
...
+
+public class MongoLoggerUtils {
+    private static ConcurrentHashMap<Integer, Logger> loggerPerConnection =
+            new ConcurrentHashMap<>();
+
+    public static Logger initConnectionLogger(Integer connection_id, Level logLevel, File logDir) {
+        Logger parentConnectionLogger = loggerPerConnection.get(connection_id);
+        if (null != parentConnectionLogger) {
+            // There is already a parent connection logger, there is a problem
+            throw new MongoException("Parent connection logger already initialed. Ignored.");
```
pmeredit, line 19: I'm not sure if this should be a `MongoException`, but I can't think of anything better, either. I do slightly worry that this will look like a mongo java driver exception rather than a mongo jdbc exception. Maybe we should create a `MongoJDBCException` `RuntimeException`?

pmeredit (approved):
This looks great! Thanks for taking the time to figure out how to integrate AspectJ into gradle; it really makes the code cleaner and more reliable.

I just have a few small comments, however you feel like addressing them is fine by me.

bucaojit (changes requested):
It would be good to add a section in the README to describe the new properties: LogLevel, LogDir.  The acceptable log levels, and the behavior of LogDir and defaults.

src/integration-test/java/com/mongodb/jdbc/integration/MongoIntegrationTest.java
nbagnard: Oh, you are right. I made the changes for the unit tests but not for integration tests.
Fixed. It will log in the current directory.

src/main/java/com/mongodb/jdbc/MongoDriver.java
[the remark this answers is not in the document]
nbagnard: Unfortunately, we can not use toLowerCase because log levels in java.utils.logging.Level are upper case. Using toLowerCase will make the parse method fail. I think here because the toUpperCase is applied to the properties values, it kind of makes sense and does not require changing what we do elsewhere.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -155,10 +172,39 @@ private MongoConnection createConnection(ConnectionString cs, Properties info)
...
+                    "Invalid "
+                            + LOG_LEVEL
+                            + " property value : "
+                            + logLevelVal
+                            + ". Valid values are : "
+                            + LEVELS
+                            + ".");
+        }
+        String logDirVal = info.getProperty(LOG_DIR);
+        File logDir = (logDirVal == null) ? null : new File(logDirVal);
```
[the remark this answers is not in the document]
nbagnard: I'm a little bit on the fence on this, despite writing the behavior in the design doc, because if the directory does not exist and the user doesn't have "access" to stdout (because the driver is used in a tool which does not redirect stdout for example), then the fact that the driver won't log in the provided directory will be lost, possibly confusing users and making things for everyone hard to debug.

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
nbagnard: Oups, thank you for finding this one :)

src/main/java/com/mongodb/jdbc/MongoDriver.java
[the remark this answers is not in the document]
nbagnard: @bucaojit Let's discuss this tomorrow during stand-up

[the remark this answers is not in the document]
bucaojit: Yes, that makes sense not to default to stdout if the directory does not exist as the user did set the property and would expect the logs to be there.  What I think would be nice to be added is a way to specify "print to stdout" using the LogDir property versus the lack of the property being what sets the behavior.

rychipman (changes requested):
a couple questions about how we're plumbing the loggers

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -31,19 +35,30 @@
...
 public abstract class MongoConnection implements Connection {
     protected MongoClient mongoClient;
     protected String currentDB;
     protected String url;
     protected String user;
     protected boolean isClosed;
+    protected MongoLogger logger;
+    protected int connectionId;
+    private static AtomicInteger connectionCounter = new AtomicInteger();
+    private AtomicInteger stmtCounter = new AtomicInteger();
```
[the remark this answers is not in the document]
nbagnard: This is because each Connection has its own set of statements so I thought it was making more sense to use a separate counter.
Otherwise, if connection 1 creates 2 statements and then connection 2 creates 1 statement instead of [c-1][s-1], [c-1][s-2], [c-2][s-1] we will have [c-1][s-1], [c-1][s-2], [c-2][s-3] which can be a little harder to track in the application code.

src/main/java/com/mongodb/jdbc/MongoDriver.java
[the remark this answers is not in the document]
nbagnard: I made the change for this.  If no value specified for LogDir or LogDir=console, it will print to the console.

src/main/java/com/mongodb/jdbc/logging/MongoLoggerUtils.java
```diff
@@ -0,0 +1,83 @@
...
+import com.mongodb.jdbc.MongoConnection;
+import java.io.File;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.logging.*;
+
+public class MongoLoggerUtils {
+    private static ConcurrentHashMap<Integer, Logger> loggerPerConnection =
+            new ConcurrentHashMap<>();
```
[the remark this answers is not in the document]
nbagnard: I removed MongoLoggerUtils altogether. This would indeed leak and it was old code from when I wanted an higher level parent logger if logging from MongoDriver.

bucaojit (approved):
LGTM

rychipman (approved):
LGTM

nbagnard:
evergreen merge

PR #83, opened 2022-03-08
SQL-589: Use validateConn after connecting

pmeredit:
This patch just changes the visibility of `validateConn` to package level and calls it after creating a connection.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -347,15 +347,6 @@ public abstract class MongoConnection implements Connection {
                 Thread.currentThread().getStackTrace()[1].toString());
     }
 
-    private void validateConn() throws SQLException {
-        Statement statement = createStatement();
-        boolean resultExists = statement.execute("SELECT 1");
-        if (!resultExists) {
-            // no resultSet returned
-            throw new SQLException("Connection error");
-        }
-    }
-
     class ConnValidation implements Callable<Object> {
         @Override
         public Object call() throws SQLException {
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -2,6 +2,7 @@ package com.mongodb.jdbc;
 
 import com.google.common.base.Preconditions;
 import com.mongodb.ConnectionString;
+import com.mongodb.MongoClientSettings;
 import com.mongodb.MongoDriverInformation;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.MongoClients;
@@ -36,6 +37,7 @@ import org.bson.BsonInt32;
 import org.bson.Document;
 
 public abstract class MongoConnection implements Connection {
+    private MongoClientSettings mongoClientSettings;
     protected MongoClient mongoClient;
     protected String currentDB;
     protected String url;
@@ -56,9 +58,10 @@ public abstract class MongoConnection implements Connection {
                                 .append(MongoDriver.MINOR_VERSION)
                                 .toString();
 
+        this.mongoClientSettings = MongoClientSettings.builder().applyConnectionString(cs).build();
         mongoClient =
                 MongoClients.create(
-                        cs,
+                        mongoClientSettings,
                         MongoDriverInformation.builder()
                                 .driverName(MongoDriver.NAME)
                                 .driverVersion(version)
@@ -70,6 +73,10 @@ public abstract class MongoConnection implements Connection {
         if (isClosed()) {
             throw new SQLException("Connection is closed.");
         }
+    }
+
+    protected int getDefaultConnectionValidationTimeoutSeconds() {
+        return this.mongoClientSettings.getSocketSettings().getConnectTimeout(TimeUnit.SECONDS);
     }
 
     String getURL() {
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -347,9 +347,9 @@ public abstract class MongoConnection implements Connection {
                 Thread.currentThread().getStackTrace()[1].toString());
     }
 
-    class ConnValidation implements Callable<Object> {
+    class ConnValidation implements Callable<Void> {
         @Override
-        public Object call() throws SQLException {
+        public Void call() throws SQLException {
             Statement statement = createStatement();
             boolean resultExists = statement.execute("SELECT 1 from DUAL");
             if (!resultExists) {
@@ -373,7 +373,7 @@ public abstract class MongoConnection implements Connection {
         // to set the timeout adhoc on the calls, we use Executor to run a blocked call with timeout.
         ExecutorService executor = Executors.newCachedThreadPool();
 
-        Future<Object> future = executor.submit(new ConnValidation());
+        Future<Void> future = executor.submit(new ConnValidation());
         try {
             if (timeout > 0) {
                 future.get(timeout, TimeUnit.SECONDS);
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -105,7 +105,7 @@ public class MongoDriver implements Driver {
         return conn;
     }
 
-    public MongoConnection createUnvalidatedConnection(String url, Properties info)
+    protected MongoConnection createUnvalidatedConnection(String url, Properties info)
         throws SQLException {
         return getUnvalidatedConnectionAndTimeout(url, info).left();
     }
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -99,7 +99,7 @@ public class MongoDriver implements Driver {
 
     @Override
     public Connection connect(String url, Properties info) throws SQLException {
-        Pair<MongoConnection, Integer> p = createUnvalidatedConnectionAndTimeout(url, info);
+        Pair<MongoConnection, Integer> p = getUnvalidatedConnectionAndTimeout(url, info);
         Connection conn = p.left();
         conn.isValid(p.right());
         return conn;
@@ -107,10 +107,10 @@ public class MongoDriver implements Driver {
 
     public MongoConnection createUnvalidatedConnection(String url, Properties info)
         throws SQLException {
-        return createUnvalidatedConnectionAndTimeout(url, info).left();
+        return getUnvalidatedConnectionAndTimeout(url, info).left();
     }
 
-    private Pair<MongoConnection, Integer> createUnvalidatedConnectionAndTimeout(String url, Properties info)
+    private Pair<MongoConnection, Integer> getUnvalidatedConnectionAndTimeout(String url, Properties info)
         throws SQLException {
         if (!acceptsURL(url)) {
             return null;
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -101,7 +101,12 @@ public class MongoDriver implements Driver {
     public Connection connect(String url, Properties info) throws SQLException {
         Pair<MongoConnection, Integer> p = getUnvalidatedConnectionAndTimeout(url, info);
         Connection conn = p.left();
-        conn.isValid(p.right());
+        // the jdbc spec requires that null be returned if a Driver cannot handle the specified URL
+        // (cases where multiple jdbc drivers are present and the program is checking which driver
+        // to use), so it is possible for conn to be null at this point.
+        if (conn != null) {
+            conn.isValid(p.right());
+        }
         return conn;
     }
 
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -99,18 +99,50 @@ public class MongoDriver implements Driver {
 
     @Override
     public Connection connect(String url, Properties info) throws SQLException {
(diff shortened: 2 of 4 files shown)


src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -162,9 +162,9 @@ public Integer getDecimalDigits() {
     }
 
     /**
-     *  CHAR_OCTET_LENGTH is the maximum length of binary and character based data in bytes.
```
pmeredit, line 165: spotless did this

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -347,7 +347,7 @@ public SQLXML createSQLXML() throws SQLException {
                 Thread.currentThread().getStackTrace()[1].toString());
     }
 
-    private void validateConn() throws SQLException {
```
pmeredit, line 350: little known fact, no privacy modifier = "package public" in Java.

pmeredit:
This is causing a unit test failure. Looking into it :)

pmeredit:
I'm actually not sure we should do this anymore. I had to add a property to turn off connection validation to make the unit tests work. Some integration tests are still failing.

pmeredit:
We're going to discuss this in standup

nbagnard (changes requested):

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -126,7 +127,11 @@ public Connection connect(String url, Properties info) throws SQLException {
                     "unexpected driver property info prompt returned: "
                             + String.join(", ", propertyNames));
         }
-        return createConnection(p.left(), info);
+        MongoConnection ret = createConnection(p.left(), info);
+        if (!info.getProperty(VALIDATE, "").toLowerCase().equals("false")) {
```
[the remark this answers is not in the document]
bucaojit: I believe the property is still needed for the unit tests as the local ADL will not be running for them.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -347,7 +347,7 @@ public SQLXML createSQLXML() throws SQLException {
...
+    void validateConn() throws SQLException {
```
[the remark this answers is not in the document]
pmeredit: We could definitely file a ticket to change `isValid` to not use a `Statement`.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -99,18 +99,51 @@
...
+
         // reuse the code getPropertyInfo to make sure the URI is properly set wrt the passed
         // Properties info value.
         Pair<ConnectionString, DriverPropertyInfo[]> p = getConnectionString(url, info);
+        // ensure that the ConnectionString and Properties are consistent.
+        reconcileProperties(p.right(), info);
+
+        ConnectionString cs = p.left();
+        Connection ret = createDialectConnection(cs, info);
+        System.out.println(cs);
```
bucaojit, line 130: This may have been left in from debugging.  The cs at this point may have the user/pass so we don't want to print it.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -99,18 +99,51 @@
...
+        // use a timeout of 5s if no timeout is specified in the URL.
+        int timeout = 5;
+        try {
+            // mongo connect timeoutes are in MS, so divide by 1000.
```
bucaojit, line 134:
```suggestion
            // mongo connect timeouts are in MS, so divide by 1000.
```
typo

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -99,18 +99,51 @@
...
+
+    public MongoConnection createUnvalidatedConnection(String url, Properties info)
+        throws SQLException {
+        return getUnvalidatedConnectionAndTimeout(url, info).left();
+    }
+
+    private Pair<MongoConnection, Integer> getUnvalidatedConnectionAndTimeout(String url, Properties info)
+        throws SQLException {
         if (!acceptsURL(url)) {
             return null;
```
bucaojit, line 116: I think it would be better to throw an exception here, if we return `null` then the calling functions would have to check for `null` or will throw NPE.

bucaojit (changes requested):
Looks good.  Glad you were able to find a solution that didn't need the extra property. 

A few suggestions.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -99,18 +99,51 @@
...
         // Properties info value.
         Pair<ConnectionString, DriverPropertyInfo[]> p = getConnectionString(url, info);
+        // ensure that the ConnectionString and Properties are consistent.
+        reconcileProperties(p.right(), info);
+
+        ConnectionString cs = p.left();
+        Connection ret = createDialectConnection(cs, info);
+        System.out.println(cs);
+        // use a timeout of 5s if no timeout is specified in the URL.
+        int timeout = 5;
```
nbagnard, line 132: From the error message we were getting when ADL did not start, it seems that the Java driver has a default timeout of 30000ms. 
Why not keeping it as is?

I understand that it might for faster fail when testing, but it would save us from having to document the differences between the Java driver default value and the JDBC driver ones if we were not changing these values.
If we want tests to have a shorter timeout, I'm sure that we can set it in the default connection url(s).

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -126,10 +159,9 @@ public Connection connect(String url, Properties info) throws SQLException {
                     "unexpected driver property info prompt returned: "
                             + String.join(", ", propertyNames));
         }
-        return createConnection(p.left(), info);
     }
 
-    private MongoConnection createConnection(ConnectionString cs, Properties info)
+    private MongoConnection createDialectConnection(ConnectionString cs, Properties info)
```
nbagnard, line 164: Can we keep the name "createConnection"?
There will be new settings extracted and passed down to create a MongoConnection for the logging.

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -61,16 +61,16 @@ public void execute() throws Throwable {
     void testDBURL() throws SQLException {
         MongoDriver d = new MongoDriver();
         // Should not return null or throw, even with null properties.
-        assertNotNull(d.connect(authDBURL, null));
+        assertNotNull(d.createUnvalidatedConnection(authDBURL, null));
```
nbagnard, line 64: I am wondering if having this in MongoSQLMock would not work : `when(mongoConnection.isValid(anyInt())).thenReturn(true);`?
This way, no need for the special method `createUnvalidatedConnection`  just for unit test.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -99,18 +99,51 @@
 
     @Override
     public Connection connect(String url, Properties info) throws SQLException {
+        Pair<MongoConnection, Integer> p = getUnvalidatedConnectionAndTimeout(url, info);
+        Connection conn = p.left();
+        conn.isValid(p.right());
```
nbagnard, line 104: Since you are making related changes, could you change ConnValidation.call return type to Void instead of Object?
This makes things totally clear this way that we don't expect a returned value.

nbagnard (changes requested):

src/main/java/com/mongodb/jdbc/MongoDriver.java
pmeredit: No

https://docs.oracle.com/javase/7/docs/api/java/sql/Driver.html#connect(java.lang.String,%20java.util.Properties)

```Attempts to make a database connection to the given URL. The driver should return "null" if it realizes it is the wrong kind of driver to connect to the given URL. This will be common, as when the JDBC driver manager is asked to connect to a given URL it passes the URL to each loaded driver in turn.```

File: src/main/java/com/mongodb/jdbc/MongoDriver.java (hunk 5 above)
Comment (pmeredit) [reply src/main/java/com/mongodb/jdbc/MongoDriver.java]: Definitely not!

File: src/main/java/com/mongodb/jdbc/MongoDriver.java (hunk 8 above)
Comment (pmeredit) [reply src/main/java/com/mongodb/jdbc/MongoDriver.java]: 30 is extremely long, but that makes sense.

File: src/test/java/com/mongodb/jdbc/MongoDriverTest.java (hunk 10 above)
Comment (pmeredit) [reply src/test/java/com/mongodb/jdbc/MongoDriverTest.java:64]: This doesn't work. Invalid use of matchers. I think the problem is that we use `MongoSQLMock` in several tests.

I also think mocking should only be used for things that can't be easily handled otherwise. Adding one protected method is not a big deal.

File: src/main/java/com/mongodb/jdbc/MongoVersionedJsonSchema.java [hunk 12]
```diff
@@ -21,7 +21,6 @@ public MongoVersionedJsonSchema(
             @BsonProperty("version") final int version,
             @BsonProperty("jsonSchema") JsonSchema schema) {
         this.version = version;
-        System.out.println(schema);
```
pmeredit, line 24: I found another debug statement we missed... I think.

@nbagnard tell me if this was actually intended to be here!

src/main/java/com/mongodb/jdbc/MongoDriver.java
bucaojit: Oh okay, good catch find that in the docs.  Then we can return the `null` after line 102 in this file if null, so that lines 103 and on don't fail with the NPE.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -99,18 +99,50 @@
...
         // reuse the code getPropertyInfo to make sure the URI is properly set wrt the passed
         // Properties info value.
         Pair<ConnectionString, DriverPropertyInfo[]> p = getConnectionString(url, info);
+        // ensure that the ConnectionString and Properties are consistent.
+        reconcileProperties(p.right(), info);
+
+        ConnectionString cs = p.left();
+        Connection ret = createConnection(cs, info);
+        // use a timeout of 30s, if no timeout is specified in the URL.
+        int timeout = 30;
```
nbagnard, line 131: This is not really what I had in mind when saying that we should keep the same default value if nothing is specified in the connection string.

I see that you kind of had to do this because cs.getConnectTimeout returns Null.

I think I found a way to have MongoConnection provide the settings used by the underlying Java driver, including the default values:
In MongoConnection (new code inside *** ... ***):
```
public abstract class MongoConnection implements Connection {
    **private MongoClientSettings mongoClientSettings;**
    protected MongoClient mongoClient;
    ...

    public MongoConnection(ConnectionString cs, String database, Level logLevel, File logDir) {
        Preconditions.checkNotNull(cs);
         ...
        **mongoClientSettings = MongoClientSettings.builder().applyConnectionString(cs).build();**
        mongoClient =
                MongoClients.create(
                        **mongoClientSettings**,
                        MongoDriverInformation.builder()
                                .driverName(MongoDriver.NAME)
                                .driverVersion(version)
                                .build());
         ...
    }
         
    **protected int getDefaultConnectionValidationTimeout()
    {
        return mongoClientSettings.getSocketSettings().getConnectTimeout(TimeUnit.MILLISECONDS);
    }**                       
```

Then, **getDefaultConnectionValidationTimeout** can be used directly from MongoConnection and getUnvalidatedConnectionAndTimeout does not need to return the timeout.

I would even go as far as creating a `protected IsValid()` method in MongoConnection which would use the default connectTimeout from `mongoClientSettings` making this transparent to MongoDriver.

nbagnard (changes requested):

src/main/java/com/mongodb/jdbc/MongoDriver.java
pmeredit: Oh, yes, duh!

src/main/java/com/mongodb/jdbc/MongoDriver.java
pmeredit: Yep, it's a really bad api that it throws NPE if there is no option related to the given method.

src/main/java/com/mongodb/jdbc/MongoConnection.java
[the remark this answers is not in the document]
nbagnard: Sounds good to me, let's just do that 👍

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -99,18 +99,55 @@
 
     @Override
     public Connection connect(String url, Properties info) throws SQLException {
+        Pair<MongoConnection, Integer> p = getUnvalidatedConnectionAndTimeout(url, info);
+        Connection conn = p.left();
```
bucaojit, line 103:
```
        Pair<MongoConnection, Integer> p = getUnvalidatedConnectionAndTimeout(url, info);
        if (p == null) {
            return null;
         }
        Connection conn = p.left();
```
`p` is returned as null for a wrong URL , p.left() would still throw NPE

pmeredit: Wow, I did that too fast.

pmeredit: (I'm changing it to return `new Pair<>(null, 0)`)

pmeredit: Actually, after looking at @nbagnard's suggestion, I was able to clean thus up even more so we don't have to deal with `Pair` at all :)

bucaojit (approved):
LGTM!

nbagnard (approved):
LGTM :)

PR #85, opened 2022-03-10
SQL-761: Disable Mysql dialect related tests

nbagnard:
Disabled TDVT and MySQLIntegrationTest but kept the code in for now.
Will remove the code when Mysql dialect is removed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,7 +25,7 @@ sourceSets {
             compileClasspath += main.output
             runtimeClasspath += main.output
             srcDir file('src/integration-test/java')
-            exclude '**/MySQLIntegrationTest'
+            exclude '**/MySQLIntegrationTest.java'
         }
         resources.srcDir file('src/integration-test/resources')
     }
--- a/build.gradle
+++ b/build.gradle
@@ -25,6 +25,7 @@ sourceSets {
             compileClasspath += main.output
             runtimeClasspath += main.output
             srcDir file('src/integration-test/java')
+            exclude '**/MySQLIntegrationTest'
         }
         resources.srcDir file('src/integration-test/resources')
     }
--- a/.evg.yml
+++ b/.evg.yml
@@ -9,19 +9,14 @@ post:
   - func: "upload jar file"
   - func: "upload unit test classes"
   - func: "upload integration test classes"
-  - func: "upload tdvt test classes"
   - func: "upload unit test packages"
   - func: "upload integration test packages"
-  - func: "upload tdvt test packages"
   - func: "upload unit test js"
   - func: "upload integration test js"
-  - func: "upload tdvt test js"
   - func: "upload unit test css "
   - func: "upload integration test css "
-  - func: "upload tdvt test css "
   - func: "upload unit test index"
   - func: "upload integration test index"
-  - func: "upload tdvt test index"
   - func: "upload xunit test results"
 
 buildvariants:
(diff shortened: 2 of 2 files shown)


bucaojit (approved):
LGTM. 

Thanks for making the change, it's good we no longer have to worry about the shared ADL.  Further cleanup such as the entry in the readme and environment variables the shared ADL used can be handled later when the MySQL dialect is removed.

build.gradle
```diff
@@ -25,6 +25,7 @@ sourceSets {
             compileClasspath += main.output
             runtimeClasspath += main.output
             srcDir file('src/integration-test/java')
+            exclude '**/MySQLIntegrationTest.java'
```
rychipman, line 28: there's some TDVT-specific configuration elsewhere in `build.gradle` too -- should we remove that as well?

rychipman (changes requested):
just one question!

build.gradle
nbagnard: I left everything as is for TDVT and only removed the part in the evergreen script which was running TDVT.
Just in case we want to run again or we decided to try it with MongoSQL dialect at some point and also because I thought a thorough clean-up would be done when removing Mysql dialect.
I had to disable build the unit test in order to stop the build from executing automatically which is why there is the change in the gradle file and not the evergreen script for the unit tests.
I can certainly clean-up and remove more things right now as opposed to when we removed mysql if you feel like it makes more sense. But I'd rather wait until we remove the dialect if that's Ok with you. So that we can still easily check that it still works until we definitely get rid of it.

rychipman: Waiting SGTM!

rychipman (approved):
LGTM

henretty:
@bucaojit @nbagnard I'm still seeing the TDVT test running when using commit `96b86cf`. Am i missing something or did anything change in how the test needs to be executed? Below is the output of `./gradlew clean build`:

```
> Task :compileTdvtTestJava
> Task :processTdvtTestResources NO-SOURCE
> Task :tdvtTestClasses

> Task :tdvtTest FAILED

com.mongodb.jdbc.tdvt.TDVTTest > testLOGICAL_CALCS3 FAILED
    java.lang.ExceptionInInitializerError at NativeConstructorAccessorImpl.java:-2
        Caused by: java.lang.RuntimeException at TDVTTest.java:23
            Caused by: java.lang.NullPointerException at TDVTTest.java:29

com.mongodb.jdbc.tdvt.TDVTTest > testLOGICAL_CALCS4 FAILED
    java.lang.NoClassDefFoundError at NativeConstructorAccessorImpl.java:-2

com.mongodb.jdbc.tdvt.TDVTTest > testLOGICAL_CALCS8 FAILED
    java.lang.NoClassDefFoundError at NativeConstructorAccessorImpl.java:-2
```

henretty:
It looks like this only happens if the MySQL test env vars are unset. To repro:

`unset ADL_TEST_HOST ADL_TEST_USER ADL_TEST_PWD ADL_TEST_AUTH_DB`
`./gradlew clean build`

bucaojit:
@henretty @nbagnard This [check.dependsOn ](https://github.com/mongodb/mongo-jdbc-driver/blob/46052fff94fbb665a5083f2f60906b98a313c490/build.gradle#L128)may be why the tdvt test is still running.  Checking gradle docs.

henretty:
Thanks @bucaojit , it looks like @nbagnard 's latest commit has got everything working on the ADL side! Thank you all!

PR #86, opened 2022-03-10
SQL-762: Updating product and driver name

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-762


nbagnard (approved):
LGTM! If you feel like it, can you put the driver name value in MongoDriver and reference it here?
It's just a selfish ask because if I want to log the driver name and version at connection time, I'll have to move the info in MongoDriver.
But, I can just as well get the latest code and move it after you merge if you want to push a quick, simple merge.

PR #87, opened 2022-03-10
SQL-765: Run spotlessCheck in an evergreen task

rychipman:
Wanted to familiarize myself a bit more with Gradle/groovy before working on SQL-504, so decided to do this one as a quick win

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -25,6 +25,13 @@ post:
   - func: "upload xunit test results"
 
 buildvariants:
+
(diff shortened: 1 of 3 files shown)


nbagnard (approved):
LGTM! I'm happy to learn that I can use <pre> to keep formatting as is. This was on my todo list of things to look into because it was starting to drive me crazy to have to remove the changes every times I ran Spotless :)

PR #92, opened 2022-03-14
SQL-775: Handle all JsonSchema polymorphic fields

nbagnard:
Transform "items" field values to a set of MongoJsonSchema and "additionalProperties" to a boolean value.

@rychipman, @bucaojit :  It seems that "items" is only used to know if isAny should return true or false. And, even there, the only information used is whether "items" is null or not.
I am not sure if we need to convert "items" to a set of MongoJsonSchema or if we could not simply convert it to a boolean value (tracking if items is null or not) and use the boolean value in isAny.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
+++ b/src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
@@ -4,6 +4,7 @@ import static com.mongodb.jdbc.BsonTypeInfo.*;
 
 import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
@@ -15,12 +16,19 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.builder.ToStringBuilder;
 import org.apache.commons.lang3.builder.ToStringStyle;
+import org.bson.BsonArray;
 import org.bson.BsonInvalidOperationException;
 import org.bson.BsonType;
 import org.bson.BsonValue;
+import org.bson.codecs.Codec;
+import org.bson.codecs.DecoderContext;
 import org.bson.codecs.pojo.annotations.BsonIgnore;
+import org.bson.json.JsonReader;
 
 public class MongoJsonSchema {
+    private static final Codec<JsonSchema> JSON_SCHEMA_CODEC =
+            MongoDriver.registry.get(JsonSchema.class);
+
     public static class ScalarProperties {
         protected String name;
         protected boolean isRequired = true;
@@ -42,7 +50,7 @@ public class MongoJsonSchema {
     public Map<String, MongoJsonSchema> properties;
     public Set<MongoJsonSchema> anyOf;
     public Set<String> required;
-    public MongoJsonSchema items;
+    public Set<MongoJsonSchema> items;
     public boolean additionalProperties;
 
     /**
@@ -81,8 +89,11 @@ public class MongoJsonSchema {
             }
         }
         result.required = baseSchema.required;
-        result.items = toSimplifiedMongoJsonSchema(baseSchema.items);
-        result.additionalProperties = baseSchema.additionalProperties;
+        if (baseSchema.items != null) {
+            result.items = polymorphicItemsToMongoJsonSchemaSet(baseSchema.items);
+        }
+        result.additionalProperties =
+                toMongoJsonSchemaAdditionalProperties(baseSchema.additionalProperties);
 
         if (baseSchema.bsonType != null) {
             Set<String> bsonTypes = polymorphicBsonTypeToStringSet(baseSchema.bsonType);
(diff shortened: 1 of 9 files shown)


src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -42,7 +50,7 @@ public ScalarProperties(String name, BsonTypeInfo type) {
     public Map<String, MongoJsonSchema> properties;
     public Set<MongoJsonSchema> anyOf;
     public Set<String> required;
-    public MongoJsonSchema items;
+    public Set<MongoJsonSchema> items;
```
rychipman, line 53: I don't think this type should be changing -- conceptually, we only maintain a single schema that describes all elements of an array, instead of allowing separate validators for values at different indexes.

When we see an array-style items schema, I think we want to merge the index schemas into a single `MongoJsonSchema` that is an `anyOf` of each of the item schemas

rychipman (changes requested):
one high-level comment re: the desired behavior here

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -164,6 +176,78 @@ private static MongoJsonSchema toMongoJsonSchema(JsonSchema baseSchema) {
...
+                                                new JsonReader(val.asDocument().toJson()),
+                                                DecoderContext.builder().build())))
+                .collect(Collectors.toSet());
+    }
+
+    /**
+     * Converts a polymorphic additionalProperties which can either be a boolean or a Document to a
+     * boolean.
+     *
+     * @param polymorphocAdditionalProperties The original polymorphic additionalProperties field.
```
bucaojit, line 223: Typo here is copied in a few places
```suggestion
     * @param polymorphicAdditionalProperties The original polymorphic additionalProperties field.
```

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -164,6 +176,78 @@ private static MongoJsonSchema toMongoJsonSchema(JsonSchema baseSchema) {
...
+
+        // The only expected types for additionalProperties are Document or Boolean
+        if (!(polymorphocAdditionalProperties.isBoolean()
+                || polymorphocAdditionalProperties.isDocument())) {
+            throw new BsonInvalidOperationException(
+                    "Value expected to be of type "
+                            + BsonType.BOOLEAN
+                            + " or "
+                            + BsonType.DOCUMENT
+                            + " but  is of unexpected type "
```
bucaojit, line 241:
```suggestion
                            + " but is of unexpected type "
```
extra space

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -164,6 +176,78 @@ private static MongoJsonSchema toMongoJsonSchema(JsonSchema baseSchema) {
...
+        }
+
+        // The only expected types for Items are BsonArray or BsonDocument
+        if (!(polymorphicItems.isArray() || polymorphicItems.isDocument())) {
+            throw new BsonInvalidOperationException(
+                    "Value expected to be of type "
+                            + BsonType.ARRAY
+                            + " or "
+                            + BsonType.DOCUMENT
+                            + " but  is of unexpected type "
```
bucaojit, line 199:
```suggestion
                            + " but is of unexpected type "
```

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
```diff
@@ -164,6 +176,78 @@ private static MongoJsonSchema toMongoJsonSchema(JsonSchema baseSchema) {
...
+     * Converts a polymorphic additionalProperties which can either be a boolean or a Document to a
+     * boolean.
+     *
+     * @param polymorphocAdditionalProperties The original polymorphic additionalProperties field.
+     * @return the corresponding boolean value.
+     */
+    private static boolean toMongoJsonSchemaAdditionalProperties(
+            BsonValue polymorphocAdditionalProperties) {
+        if (polymorphocAdditionalProperties == null) {
+            // By default, additional properties is false
```
bucaojit, line 229: In the [document](https://docs.mongodb.com/manual/reference/operator/query/jsonSchema/) linked in the comments, the behavior description says additionalProperties defaults to `true`.  Is this defaulting to false specific to the JDBC behavior?

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
nbagnard: Fixed

src/main/java/com/mongodb/jdbc/MongoJsonSchema.java
nbagnard: umh, this was the original behavior, no change there. I'm not sure why it was like that.

nbagnard: Taking this back. If no value is provided, it seems to make more sense to default to false, no?
This is also what isAny expects.

bucaojit: Yes, let's keep it default to false as that is the current behavior.

rychipman (approved):
LGTNM

bucaojit (approved):
LGTM

nbagnard:
evergreen merge

PR #96, opened 2022-03-18
SQL-789 : Enable and populate tests

nbagnard:
Enable and add expected result sets which should have been done as part of SQL-668.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -101,9 +101,12 @@ functions:
       working_dir: mongo-jdbc-driver
       script: |
           ${PREPARE_SHELL}
-          ./resources/run_adl.sh start && ./gradlew runDataLoader
+          ./resources/run_adl.sh start &&
+          ./gradlew runDataLoader &&
           ./gradlew clean integrationTest -x test
+          EXITCODE=$?
           ./resources/run_adl.sh stop
+          exit $EXITCODE
 
   "export variables":
     - command: shell.exec
--- a/.evg.yml
+++ b/.evg.yml
@@ -101,8 +101,7 @@ functions:
       working_dir: mongo-jdbc-driver
       script: |
           ${PREPARE_SHELL}
-          ./resources/run_adl.sh start
-          ./gradlew runDataLoader
+          ./resources/run_adl.sh start && ./gradlew runDataLoader
           ./gradlew clean integrationTest -x test
           ./resources/run_adl.sh stop
 
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
@@ -145,8 +145,7 @@ public class DataLoader {
                     MongoDatabase database = mongoClient.getDatabase(entry.db);
                     if (entry.collection != null) {
                         loadCollection(entry, database);
-                    }
-                    else if (entry.view != null) {
+                    } else if (entry.view != null) {
                         generateSchema(entry.db, entry.view);
                     }
                 }
@@ -159,6 +158,7 @@ public class DataLoader {
 
     /**
      * Loads a collection with the information provided in the TestDataEntry.
+     *
      * @param entry the collection entry.
      * @param database The database to add the collection to.
      */
@@ -194,12 +194,7 @@ public class DataLoader {
             for (Map<String, Object> index : entry.nonuniqueIndexes) {
                 String indexName = collection.createIndex(new Document(index));
                 System.out.println(
-                        "Created index "
-                                + indexName
-                                + " on "
-                                + entry.db
-                                + "."
-                                + entry.collection);
+                        "Created index " + indexName + " on " + entry.db + "." + entry.collection);
             }
         }
         if (entry.schema != null) {
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
@@ -14,6 +14,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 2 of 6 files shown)


.evg.yml
```diff
@@ -101,8 +101,7 @@ functions:
       working_dir: mongo-jdbc-driver
       script: |
           ${PREPARE_SHELL}
-          ./resources/run_adl.sh start
-          ./gradlew runDataLoader
+          ./resources/run_adl.sh start && ./gradlew runDataLoader
```
mattChiaravalloti, line 104: clever 🚀

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
```diff
@@ -42,13 +42,15 @@
 
     private List<TestDataEntry> datasets;
     private Set<Pair<String, String>> collections;
+    private Set<Pair<String, String>> views;
```
mattChiaravalloti, line 45: What is the purpose of this variable? It is declared here, initialized on line 53, and then updated on line 73, but is otherwise never used or referenced anywhere else.

If you are trying to add views to the data in ADL, you'll need to define the view definition in the ADL config (`resources/integration_test/testdata/adl_db_config.json`), _not_ in the mongodb data.

We previously had this type of set up in an earlier iteration of the `DataLoader` (defining views in the `resources/integration_test/testdata/*.yml` files) but we realized that ADL doesn't distinguish between views and collections in mongodb. As in, if you have a mongodb datasource in ADL, ADL treats all views and collections from that mongodb as collections. The only views ADL acknowledges are those defined in the config.

mattChiaravalloti (commented):
The test updates look good, but I'm curious about the `view` stuff added to the dataloader code

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
nbagnard: There is already a view in ADL, but when using it in the integration tests it is currently returning the error `MongoServerError: failed to query for tenant schema for "integration_test"."baz": schema not found`. We found out with Oliver that this seems to be an issue with the local ADL.
For integration tests, we are installing a new local ADL which has all the configuration for the views and collections we want to use. Next, the DataLoader will drop all the collections, add them back, populate them and generate/set their schema, before we start, using the Yaml file. Unfortunately, today, when a collection is dropped, retrieving the schema for the view fails, even after the collection schema is set.
So, I had to add a view section in the Yaml file matching the one in our ADL config to know there is a view we need to generate the schema for (but not load data into), before we can run the unit tests.

nbagnard: As for the variable, I'm looking into it, I think you are right, it is not used.

.evg.yml
```diff
@@ -101,8 +101,7 @@ functions:
...
           ./gradlew clean integrationTest -x test
           ./resources/run_adl.sh stop
```
bucaojit, line 106:
```suggestion
          ./resources/run_adl.sh start && 
          ./gradlew runDataLoader &&
          ./gradlew clean integrationTest -x test &&
          ./resources/run_adl.sh stop
```
Let's add the && for all the commands here, that way evergreen will see the final exit 1 if it occurs.

bucaojit (changes requested):

.evg.yml
nbagnard: umh, so my understanding is that it won't execute the next command unless the previous one as an exit status != 1.
I do want to stop ADL even if integration tests fail. Adding one after runDataLoader is Ok. I wasn't sure what if Gradle was playing nice with the exit status, so I tested it and it does. Adding it right now.

bucaojit: There needs to be more logic in the script then to always run `./resources/run_adl.sh stop` but also return the exit code of 1 if one of the earlier steps fails.  
Something like (didn't test this myself):
```
./resources/run_adl.sh start && 
./gradlew runDataLoader &&
./gradlew clean integrationTest -x test
EXITCODE=$?
./resources/run_adl.sh stop
exit $EXITCODE
```

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
mattChiaravalloti: Ok that makes sense. So it looks like we don't even need a set that tracks all of the views, we just need to check if the dataset is a view during schema generation (a change already in this PR). Looks good 👍

mattChiaravalloti (approved):
lgtm pending the outcome of the conversation you and Oliver are having wrt commands in evergreen 🌲

.evg.yml
nbagnard: ok, let me try this with a "patch" using a broken dataloader and see what it does
Thank you for the help here :)

nbagnard: Tested the proposed solution, it works perfectly! Thank you @bucaojit :)

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
```diff
@@ -132,49 +139,10 @@ public void loadTestData() throws IOException {
...
-                        }
-                    }
-                    if (entry.schema != null) {
-                        setSchema(entry.db, entry.collection, entry.schema);
-                    } else {
-                        generateSchema(entry.db, entry.collection);
+                    if (entry.collection != null) {
+                        loadCollection(entry, database);
+                    } else if (entry.view != null) {
+                        generateSchema(entry.db, entry.view);
```
bucaojit, line 145: Does the view schema have to be generated after the source collection is loaded and has a generated schema? In our case `bar`.  If so, then we may need to add logic to ensure the view schema is generated after the collections are loaded/schema generated.  
Or if the items in the yaml file is processed in the order it is in the file, it may be enough to add the `baz` view entry at the end of dbmd_variable_result_set_data.yml since that's where the `bar` collection is in.

nbagnard: ah, good point. I'll add some more logic to handle views in the end (probably bringing back the separate list).

nbagnard: Done!

nbagnard:
evergreen merge

nbagnard:
evergreen merge

bucaojit (approved):
Reviewed the latest changes to IntegrationTestUtils, looks good.

nbagnard:
> Reviewed the latest changes to IntegrationTestUtils, looks good.

Thank you! Always good to have someone take a second look :)

nbagnard:
evergreen merge

PR #97, opened 2022-03-22
SQL-795: Allow deserializing empty json schemas

nbagnard:


rychipman (changes requested):
my only request is that we add a regression test for this fix!

nbagnard:
@rychipman : I've added a unit test only because I think that it should be enough and with the current bug on our local ADL it is a little bit tricky to add an integration test.

rychipman (approved):
Unit test SGTM -- thanks!
LGTM

nbagnard:
evergreen merge

nbagnard:
evergreen merge

bucaojit:
evergreen merge

PR #98, opened 2022-03-22
SQL-800: Implemented getString for Document and Array columns

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-800

getString() for Document and Array columns return extJSON string.

Added array value to the row in `types_other` collection.   Added integration test to query array and document values as string.

### DIFF
--- a/resources/integration_test/testdata/integration.yml
+++ b/resources/integration_test/testdata/integration.yml
@@ -91,13 +91,14 @@ dataset:
     docsExtJson:
       - {
         _id: 0,
-        array: [1,2,3],
+        array: [1,2,3,"$oid": "000000000000000000000003", "$timestamp": {"t":200, "i":0}],
         dbPointer: { "$dbPointer": { "$ref": "namespace", "$id": { "$oid": "000000000000000000000001" } } },
         javascript: {"$code": "function(){ }"},
         javascriptWithScope: {"$code": "function(){ }", "$scope": {"foo":"bar"}},
         maxKey: {"$maxKey": 1},
         minKey: {"$minKey": 1},
-        object: {"foo":"bar"},
+        object: {"foo":"bar", "objId": {"$oid": "000000000000000000000002"}, "value": 3,
+                 "time": {"$timestamp": {"t":200, "i":0}}},
         objectId: { "$oid": "000000000000000000000001" },
         regularExpression: {"$regularExpression": {pattern: "a(bc)*", "options": ""}},
         symbol: { "$symbol": "symbol" },
--- a/resources/integration_test/testdata/integration.yml
+++ b/resources/integration_test/testdata/integration.yml
@@ -91,18 +91,18 @@ dataset:
     docsExtJson:
       - {
         _id: 0,
-        array: [ 1, 2, 3, "$oid": "000000000000000000000003", "$timestamp": {"t":200, "i":0}],
-        dbPointer: { "$dbPointer": { "$ref": "namespace", "$id": {"$oid": "000000000000000000000001" }}},
-        javascript: {"$code": "function(){ }"},
-        javascriptWithScope: {"$code": "function(){ }", "$scope": {"foo":"bar"}},
-        maxKey: {"$maxKey": 1},
-        minKey: {"$minKey": 1},
-        object: {"foo":"bar", "objId": {"$oid": "000000000000000000000002"}, "value": 3,
-                 "time": {"$timestamp": {"t":200, "i":0}}},
+        array: [ 1, 2, 3, "$oid": "000000000000000000000003", "$timestamp": { "t": 200, "i": 0 }],
+        dbPointer: { "$dbPointer": { "$ref": "namespace", "$id": { "$oid": "000000000000000000000001" }}},
+        javascript: { "$code": "function(){ }" },
+        javascriptWithScope: { "$code": "function(){ }", "$scope": { "foo": "bar" }},
+        maxKey: { "$maxKey": 1 },
+        minKey: { "$minKey": 1 },
+        object: { "foo": "bar", "objId": { "$oid": "000000000000000000000002" }, "value": 3,
+                 "time": { "$timestamp": { "t": 200, "i": 0 }}},
         objectId: { "$oid": "000000000000000000000001" },
-        regularExpression: {"$regularExpression": {pattern: "a(bc)*", "options": ""}},
+        regularExpression: { "$regularExpression": { pattern: "a(bc)*", "options": "" }},
         symbol: { "$symbol": "symbol" },
-        timestamp: {"$timestamp": {"t":100, "i":0}}
+        timestamp: { "$timestamp": { "t": 100, "i": 0 }}
         # Skip reason: SQL-395
         # undefined: {"$undefined":true},
       }
--- a/resources/integration_test/testdata/integration.yml
+++ b/resources/integration_test/testdata/integration.yml
@@ -91,10 +91,10 @@ dataset:
     docsExtJson:
       - {
         _id: 0,
-        array: [1,2,3,"$oid": "000000000000000000000003", "$timestamp": {"t":200, "i":0}],
-        dbPointer: { "$dbPointer": { "$ref": "namespace", "$id": { "$oid": "000000000000000000000001" } } },
+        array: [1,2,3,{"$oid": "000000000000000000000003"}, {"$timestamp": {"t":200, "i":0}}],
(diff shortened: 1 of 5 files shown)


src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
```diff
@@ -661,16 +661,31 @@ public static boolean compareRow(List<Object> expectedRow, ResultSet actualRow)
...
-                    if (!expected_obj.equals(actual_obj)) {
-                        System.err.println(
-                                "Expected Bson Other value "
-                                        + expected_obj
-                                        + " but is "
-                                        + actual_obj
-                                        + " for column "
-                                        + (i + 1));
-                        return false;
+                    if (expected_obj instanceof String) {
```
bucaojit, line 664: Added this branch to use getString() instead of getObject() if the expected input is a string.  Which will be from an `expected_result` entry as opposed to `expected_result_extended_json` which wouldn't have a String

rychipman (changes requested):

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -388,7 +388,7 @@ protected String getString(BsonValue o) throws SQLException {
             case DECIMAL128:
                 return o.asDecimal128().getValue().toString();
             case DOCUMENT:
-                return handleStringConversionFailure(DOCUMENT);
+                return o.asDocument().toString();
```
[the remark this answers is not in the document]
bucaojit: Switched to to use `toJson()` and pass in the settings to print extended json.

resources/integration_test/tests/query.yaml
```diff
@@ -188,53 +188,59 @@ tests:
...
+                         false, false, false]
     expected_is_writable: [false, false, false, false, false, false, false, false,
-                           false, false, false]
+                           false, false, false, false]
+
+  - description: query_array_and_document_as_string
+    db: integration_test
+    sql: SELECT `array`, `object` FROM types_other
+    expected_result:
+      - [ "[BsonInt32{value=1}, BsonInt32{value=2}, BsonInt32{value=3}]", "{\"foo\": \"bar\"}" ]
```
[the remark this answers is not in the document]
bucaojit: Made the change to serialize the array to extended JSON.  
Many thanks to @nbagnard for coming up with a solution with the `NoCheckStateJsonWriter` class which allows the BsonArray to be passed in directly to the BsonArrayCodec.

Added an objectid and timestamp to the test for the document and array.

resources/integration_test/tests/query.yaml
```diff
@@ -188,13 +188,15 @@ tests:
     sql: SELECT * FROM types_other
     expected_result_extended_json: [ {
       "0": 0,
-      "1": [1,2,3],
+      "1": [{"$numberInt": "1"}, {"$numberInt": "2"}, {"$numberInt": "3"},
```
nbagnard, line 191: Just a suggestion: Maybe we could use the relaxed extended JSON format. This would help make the data a little less verbose and easier to display in a UI based tool?

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -373,9 +385,15 @@ protected String getString(BsonValue o) throws SQLException {
         if (checkNull(o)) {
             return null;
         }
+        JsonWriterSettings settings = JsonWriterSettings.builder().outputMode(JsonMode.EXTENDED).build();
```
nbagnard: The settings and registry can be final static variables, there should be no need to create new ones each time toString is called.

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -373,9 +385,15 @@ protected String getString(BsonValue o) throws SQLException {
...
+        CodecRegistry registry = fromProviders(new BsonValueCodecProvider());
         switch (o.getBsonType()) {
```
nbagnard, line 389: I am wondering if it would not make sense to convert all values using extended Json in relaxed mode no matter the type. This would make things more coherent and predictable for everyone.

src/main/java/com/mongodb/jdbc/NoCheckStateJsonWriter.java
```diff
@@ -0,0 +1,17 @@
+package com.mongodb.jdbc;
+
+import java.io.Writer;
+import org.bson.json.JsonWriter;
+import org.bson.json.JsonWriterSettings;
+
+public class NoCheckStateJsonWriter extends JsonWriter {
```
nbagnard, line 7: Maybe we could add a comment on this one.
Something like this maybe :
JsonWriter which allow to write any Json value without validating it is constructing a valid document.

nbagnard (changes requested):

resources/integration_test/tests/query.yaml
```diff
@@ -188,53 +188,64 @@ tests:
...
+                         false, false, false]
     expected_is_writable: [false, false, false, false, false, false, false, false,
-                           false, false, false]
+                           false, false, false, false]
+
+  - description: query_array_and_document_as_string
+    db: integration_test
+    sql: SELECT `array`, `object` FROM types_other
+    expected_result:
+      - ["[{\"$numberInt\": \"1\"}, {\"$numberInt\": \"2\"}, {\"$numberInt\": \"3\"}, \
```
rychipman, line 246: small suggestion: let's use single quotes for this string so that we don't need to escape all the double quotes

resources/integration_test/testdata/integration.yml
```diff
@@ -91,18 +91,18 @@ dataset:
     docsExtJson:
       - {
         _id: 0,
+        array: [1,2,3,"$oid": "000000000000000000000003", "$timestamp": {"t":200, "i":0}],
```
rychipman, line 94: I was a little surprised to see that this parsed at all -- does YAML let you specify a document as key-value pairs without curly braces even inside a bracketed array?

Either way, I'd suggest adding the curly braces just to make it a bit more explicit
```suggestion
        array: [1,2,3,{"$oid": "000000000000000000000003"}, {"$timestamp": {"t":200, "i":0}}],
```

rychipman (changes requested):
just a couple tiny things beyond natacha's comments

resources/integration_test/tests/query.yaml
bucaojit: Switched to `JsonMode.RELAXED`

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
bucaojit: Made the change to make them static final variables.

src/main/java/com/mongodb/jdbc/NoCheckStateJsonWriter.java
bucaojit: Added a comment 👍

nbagnard (approved):
LGTM 😄 🎉

resources/integration_test/tests/query.yaml
bucaojit: The string written by the JsonWriter in getString() uses double quotes which wouldn't match if I switched these to single quotes in the expected results.

resources/integration_test/testdata/integration.yml
bucaojit: Added the curly braces. 

I checked to see what the row looks like after SnakeYaml loads it and I see that it adds the curly braces, which is probably why the parsing worked:
in yml file:
array: [1,2,3,"$oid": "000000000000000000000003", "$timestamp": {"t":200, "i":0}],
println in Java of the entry:
array=[1, 2, 3, {$oid=000000000000000000000003}, {$timestamp={t=200, i=0}}]

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
bucaojit: Discussed offline and will decide how to handle this as part of the type mapping.

resources/integration_test/tests/query.yaml
bucaojit: It just occurred to me what you meant.  Made the change 👍

rychipman (approved):
LGTM

bucaojit:
evergreen merge

PR #99, opened 2022-03-25
SQL-785: Database property is required

nbagnard:


src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -299,11 +318,13 @@ private static String removePrefix(String prefix, String s) {
 
     // getConnectionString constructs a valid MongoDB connection string which will be used as an input to the mongoClient.
     // If there are required fields missing, those fields will be returned in DriverPropertyInfo[] with a null connectionString
-    private Pair<ConnectionString, DriverPropertyInfo[]> getConnectionString(
+    private Pair<ConnectionString, DriverPropertyInfo[]> getConnectionSettings(
             String url, Properties info) throws SQLException {
+        List<DriverPropertyInfo> propertyInfoList = new ArrayList<>();
```
bucaojit, line 323: I don't see the `propertyInfoList` variable used?

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -153,39 +151,59 @@ protected MongoConnection getUnvalidatedConnection(String url, Properties info)
...
-                throw new SQLException("password specified without user");
-            }
-            if (driverPropertyInfo[0].name.equals(PASSWORD)) {
-                throw new SQLException("user specified without password");
-            }
-            String[] propertyNames = new String[driverPropertyInfo.length];
-            for (int i = 0; i < propertyNames.length; ++i) {
-                propertyNames[i] = driverPropertyInfo[i].name;
+        if (missingRequiredProperties.length != 0) {
+            for (DriverPropertyInfo info : missingRequiredProperties) {
```
bucaojit, line 173: Not sure I understand the logic here.  There is a for loop through the properties, however it looks like the first one will always throw the  SQLException.  Or was it meant to loop through the properties and append the errors to a string and throw that exception at the end?

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -221,25 +239,26 @@ private MongoConnection createConnection(ConnectionString cs, Properties info)
...
+                                    + " is "
+                                    + MONGOSQL_DIALECT);
                 }
-                return new MongoSQLConnection(cs, info.getProperty(DATABASE), logLevel, logDir);
+                return new MongoSQLConnection(cs, database, logLevel, logDir);
             default:
-                throw new SQLClientInfoException(
-                        String.format("invalid dialect '%s'", dialect),
-                        Collections.singletonMap(DIALECT, ClientInfoStatus.REASON_VALUE_INVALID));
+                throw new SQLException(
```
bucaojit, line 255: Good catch, it looks like `SQLClientInfoException` is mainly thrown in the `setClientInfo()` method.

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -87,113 +90,112 @@ void testDBURL() throws SQLException {
...
+            fail("The connection should fail because a mandatory connection setting is missing.");
+        } catch (SQLException e) {
+            // Expected failure
+            assertEquals(
+                    MongoDriver.CONNECTION_ERROR_SQLSTATE,
+                    e.getSQLState(),
+                    "Expect SQL state "
+                            + MongoDriver.CONNECTION_ERROR_SQLSTATE
+                            + " but got "
+                            + e.getMessage());
```
bucaojit, line 107:
```suggestion
                            + e.getSQLState());
```
Suggestion to match the type, SQL state.

bucaojit (changes requested):

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
nbagnard: You are right!

pmeredit (approved):
LGTM

nbagnard:
evergreen merge

PR #100, opened 2022-03-31
SQL-804: Make MongoSQLResultSet.getObject().toString() consistent with getString()

mattChiaravalloti:
This PR updates `MongoSQLResultSet.getObject()` to return a value such that calling `toString()` on it is consistent with calling `MongoSQLResultSet.getString()`. Both versions now produce what `getString()` was already returning. As discussed in standup, this was accomplished by wrapping the result of `getObject()` for `Types.OTHER` types in a new wrapper class that overrides `toString()` as needed.

~Note: we separately discussed having all `Types.OTHER` types stringify to extended JSON, but that change is outside the scope of this ticket. We technically haven't even decided if that is the direction we want to go as that is still up for debate in the design doc. Therefore, this ticket does not introduce any new behavior to the stringification -- it exactly matches what `getString()` was already doing.~

After the design discussion and seeing this ticket is included in the epic, I will update stringification to use extended json for the OTHER types as part of this ticket.

### DIFF
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
@@ -6,8 +6,8 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
 import com.mongodb.jdbc.MongoColumnInfo;
-import com.mongodb.jdbc.MongoSQLResultSetMetaData;
 import com.mongodb.jdbc.MongoSQLBsonValue;
+import com.mongodb.jdbc.MongoSQLResultSetMetaData;
 import com.mongodb.jdbc.integration.testharness.models.TestEntry;
 import com.mongodb.jdbc.integration.testharness.models.Tests;
 import java.io.File;
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
@@ -684,25 +684,15 @@ public class IntegrationTestUtils {
                         ExtJsonValue expectedAsExtJsonValue =
                                 new ExtJsonValue((BsonValue) expected_obj);
                         if (!expectedAsExtJsonValue.equals(actual_obj)) {
-                            System.err.println(
-                                    "Expected Bson Other BsonValue value "
-                                            + expected_obj
-                                            + " but is "
-                                            + actual_obj
-                                            + " for column "
-                                            + (i + 1));
-                            return false;
-                        }
-                    } else {
-                        Object actual_obj = actualRow.getObject(i + 1);
-                        if (!expected_obj.equals(actual_obj)) {
-                            return "Expected Bson Other value "
+                            return "Expected Bson Other BsonValue value "
                                     + expected_obj
                                     + " but is "
                                     + actual_obj
                                     + " for column "
                                     + (i + 1);
                         }
+                    } else {
+                        throw new IllegalArgumentException("unsupported expected value class: " + expected_obj.getClass());
                     }
                     break;
                 default:
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
@@ -681,7 +681,8 @@ public class IntegrationTestUtils {
                         }
                     } else if (expected_obj instanceof BsonValue) {
                         Object actual_obj = actualRow.getObject(i + 1);
-                        ExtJsonValue expectedAsExtJsonValue = new ExtJsonValue((BsonValue) expected_obj);
+                        ExtJsonValue expectedAsExtJsonValue =
+                                new ExtJsonValue((BsonValue) expected_obj);
                         if (!expectedAsExtJsonValue.equals(actual_obj)) {
                             System.err.println(
                                     "Expected Bson Other BsonValue value "
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
@@ -5,6 +5,7 @@ import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
+import com.mongodb.jdbc.ExtJsonValue;
 import com.mongodb.jdbc.MongoColumnInfo;
 import com.mongodb.jdbc.MongoSQLResultSetMetaData;
 import com.mongodb.jdbc.integration.testharness.models.TestEntry;
@@ -30,6 +31,7 @@ import java.util.Map;
 import java.util.Objects;
 import org.bson.BsonDocument;
 import org.bson.BsonInt32;
+import org.bson.BsonValue;
 import org.bson.Document;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
@@ -676,6 +678,19 @@ public class IntegrationTestUtils {
                                     + actual_obj
                                     + " for column "
                                     + (i + 1);
+                        }
+                    } else if (expected_obj instanceof BsonValue) {
+                        Object actual_obj = actualRow.getObject(i + 1);
+                        ExtJsonValue expectedAsExtJsonValue = new ExtJsonValue((BsonValue) expected_obj);
+                        if (!expectedAsExtJsonValue.equals(actual_obj)) {
+                            System.err.println(
+                                    "Expected Bson Other BsonValue value "
+                                            + expected_obj
+                                            + " but is "
+                                            + actual_obj
+                                            + " for column "
+                                            + (i + 1));
+                            return false;
                         }
                     } else {
                         Object actual_obj = actualRow.getObject(i + 1);
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
@@ -7,7 +7,7 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 import com.mongodb.jdbc.MongoColumnInfo;
 import com.mongodb.jdbc.MongoSQLResultSetMetaData;
-import com.mongodb.jdbc.MongoSQLValue;
+import com.mongodb.jdbc.MongoSQLBsonValue;
 import com.mongodb.jdbc.integration.testharness.models.TestEntry;
 import com.mongodb.jdbc.integration.testharness.models.Tests;
 import java.io.File;
@@ -681,8 +681,8 @@ public class IntegrationTestUtils {
                         }
                     } else if (expected_obj instanceof BsonValue) {
                         Object actual_obj = actualRow.getObject(i + 1);
-                        MongoSQLValue expectedAsExtJsonValue =
-                                new MongoSQLValue((BsonValue) expected_obj);
+                        MongoSQLBsonValue expectedAsExtJsonValue =
+                                new MongoSQLBsonValue((BsonValue) expected_obj);
                         if (!expectedAsExtJsonValue.equals(actual_obj)) {
                             return "Expected Bson Other BsonValue value "
                                     + expected_obj
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
@@ -5,7 +5,7 @@ import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
-import com.mongodb.jdbc.ExtJsonValue;
+import com.mongodb.jdbc.MongoSQLValue;
 import com.mongodb.jdbc.MongoColumnInfo;
 import com.mongodb.jdbc.MongoSQLResultSetMetaData;
 import com.mongodb.jdbc.integration.testharness.models.TestEntry;
@@ -681,8 +681,8 @@ public class IntegrationTestUtils {
                         }
                     } else if (expected_obj instanceof BsonValue) {
                         Object actual_obj = actualRow.getObject(i + 1);
-                        ExtJsonValue expectedAsExtJsonValue =
-                                new ExtJsonValue((BsonValue) expected_obj);
+                        MongoSQLValue expectedAsExtJsonValue =
+                                new MongoSQLValue((BsonValue) expected_obj);
                         if (!expectedAsExtJsonValue.equals(actual_obj)) {
                             return "Expected Bson Other BsonValue value "
                                     + expected_obj
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
@@ -5,9 +5,9 @@ import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.junit.jupiter.api.Assertions.fail;
 
-import com.mongodb.jdbc.MongoSQLValue;
 import com.mongodb.jdbc.MongoColumnInfo;
 import com.mongodb.jdbc.MongoSQLResultSetMetaData;
+import com.mongodb.jdbc.MongoSQLValue;
 import com.mongodb.jdbc.integration.testharness.models.TestEntry;
 import com.mongodb.jdbc.integration.testharness.models.Tests;
 import java.io.File;
@@ -692,7 +692,8 @@ public class IntegrationTestUtils {
                                     + (i + 1);
                         }
                     } else {
-                        throw new IllegalArgumentException("unsupported expected value class: " + expected_obj.getClass());
+                        throw new IllegalArgumentException(
+                                "unsupported expected value class: " + expected_obj.getClass());
                     }
                     break;
                 default:
--- a/src/main/java/com/mongodb/jdbc/ExtJsonValue.java
+++ b/src/main/java/com/mongodb/jdbc/ExtJsonValue.java
@@ -1,0 +1,114 @@
+package com.mongodb.jdbc;
+
+import org.bson.BsonArray;
+import org.bson.BsonValue;
+import org.bson.codecs.BsonValueCodecProvider;
+import org.bson.codecs.Codec;
+import org.bson.codecs.EncoderContext;
+import org.bson.codecs.configuration.CodecRegistry;
+import org.bson.json.JsonMode;
+import org.bson.json.JsonWriterSettings;
+
+import java.io.StringWriter;
+import java.sql.Date;
+import java.text.SimpleDateFormat;
+
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+
+/**
+ *  ExtJsonValue is a wrapper for BsonValue. The purpose of this class is to
+ *  override the toString() method to produce the extended JSON representation
+ *  of a BsonValue, rather than the java driver's default BsonValue.toString()
+ *  output.
+ *
+ *  The driver's BsonValue class is abstract and intentionally cannot be
+ *  extended by third parties. The driver explains this is to keep the BSON
+ *  type system closed.
+ */
+public class ExtJsonValue {
+    static final JsonWriterSettings JSON_WRITER_SETTINGS =
+            JsonWriterSettings.builder().outputMode(JsonMode.RELAXED).build();
+    static final CodecRegistry CODEC_REGISTRY = fromProviders(new BsonValueCodecProvider());
+    static final EncoderContext ENCODER_CONTEXT = EncoderContext.builder().build();
+
+    // dateFormat cannot be static due to a threading bug in the library.
+    protected SimpleDateFormat dateFormat = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss.SSS'Z'");
+
+    private BsonValue v;
+
+    public ExtJsonValue(BsonValue v) {
+        this.v = v;
+    }
+
+    /**
+     * @return The underlying BsonValue
+     */
+    public BsonValue getBsonValue() {
+        return this.v;
+    }
+
+    @Override
+    public String toString() {
+        switch (this.v.getBsonType()) {
+            case ARRAY:
+                Codec<BsonArray> codec = CODEC_REGISTRY.get(BsonArray.class);
+                StringWriter writer = new StringWriter();
+                codec.encode(
+                        new NoCheckStateJsonWriter(writer, JSON_WRITER_SETTINGS),
+                        this.v.asArray(),
+                        ENCODER_CONTEXT);
+                writer.flush();
+                return writer.toString();
+            case BOOLEAN:
+                return this.v.asBoolean().getValue() ? "true" : "false";
+            case DATE_TIME:
+                Date d = new Date(this.v.asDateTime().getValue());
+                return dateFormat.format(d);
+            case DECIMAL128:
+                return this.v.asDecimal128().getValue().toString();
+            case DOCUMENT:
+                return this.v.asDocument().toJson(JSON_WRITER_SETTINGS);
+            case DOUBLE:
+                return Double.toString(this.v.asDouble().getValue());
+            case INT32:
+                return Integer.toString(this.v.asInt32().getValue());
+            case INT64:
+                return Long.toString(this.v.asInt64().getValue());
+            case NULL:
+                return null;
+            case OBJECT_ID:
+                return this.v.asObjectId().getValue().toString();
+            case STRING:
+                return this.v.asString().getValue();
+            case UNDEFINED:
+                // this is consistent with $convert in mongodb.
+                return null;
+            case BINARY:
+            case DB_POINTER:
+            case END_OF_DOCUMENT:
+            case JAVASCRIPT:
+            case JAVASCRIPT_WITH_SCOPE:
+            case MAX_KEY:
+            case MIN_KEY:
+            case REGULAR_EXPRESSION:
+            case SYMBOL:
+            case TIMESTAMP:
+            default:
+                return "<invalid>"; // TODO - what should we do for these types?
+        }
+    }
+
+    @Override
+    public int hashCode() {
+        return this.v.hashCode();
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        // Compare the wrapped BsonValue for equality
+        return this.v.equals(((ExtJsonValue) o).v);
+    }
+}
--- a/src/main/java/com/mongodb/jdbc/ExtJsonValue.java
+++ b/src/main/java/com/mongodb/jdbc/ExtJsonValue.java
@@ -1,5 +1,10 @@
 package com.mongodb.jdbc;
 
+import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
+
+import java.io.StringWriter;
+import java.sql.Date;
+import java.text.SimpleDateFormat;
 import org.bson.BsonArray;
 import org.bson.BsonValue;
 import org.bson.codecs.BsonValueCodecProvider;
@@ -9,21 +14,13 @@ import org.bson.codecs.configuration.CodecRegistry;
 import org.bson.json.JsonMode;
 import org.bson.json.JsonWriterSettings;
 
-import java.io.StringWriter;
-import java.sql.Date;
-import java.text.SimpleDateFormat;
-
-import static org.bson.codecs.configuration.CodecRegistries.fromProviders;
-
 /**
- *  ExtJsonValue is a wrapper for BsonValue. The purpose of this class is to
- *  override the toString() method to produce the extended JSON representation
- *  of a BsonValue, rather than the java driver's default BsonValue.toString()
- *  output.
+ * ExtJsonValue is a wrapper for BsonValue. The purpose of this class is to override the toString()
+ * method to produce the extended JSON representation of a BsonValue, rather than the java driver's
+ * default BsonValue.toString() output.
  *
- *  The driver's BsonValue class is abstract and intentionally cannot be
- *  extended by third parties. The driver explains this is to keep the BSON
- *  type system closed.
+ * <p>The driver's BsonValue class is abstract and intentionally cannot be extended by third
+ * parties. The driver explains this is to keep the BSON type system closed.
  */
 public class ExtJsonValue {
     static final JsonWriterSettings JSON_WRITER_SETTINGS =
@@ -40,9 +37,7 @@ public class ExtJsonValue {
         this.v = v;
     }
 
-    /**
-     * @return The underlying BsonValue
-     */
+    /** @return The underlying BsonValue */
     public BsonValue getBsonValue() {
         return this.v;
     }
--- a/src/main/java/com/mongodb/jdbc/ExtJsonValue.java
+++ b/src/main/java/com/mongodb/jdbc/ExtJsonValue.java
@@ -44,6 +44,10 @@ public class ExtJsonValue {
 
     @Override
     public String toString() {
+        if (this.v == null) {
+            return null;
+        }
+
         switch (this.v.getBsonType()) {
             case ARRAY:
                 Codec<BsonArray> codec = CODEC_REGISTRY.get(BsonArray.class);
--- a/src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
+++ b/src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
@@ -115,22 +115,25 @@ public class MongoSQLResultSet extends MongoResultSet<BsonDocument> implements R
         }
         switch (columnType) {
             case Types.BIGINT:
+                return getLong(o);
             case Types.INTEGER:
             case Types.SMALLINT:
             case Types.TINYINT:
                 return getInt(o);
             case Types.BINARY:
+            case Types.LONGVARBINARY:
+            case Types.VARBINARY:
                 return o.asBinary().getData();
             case Types.BIT:
             case Types.BOOLEAN:
                 return getBoolean(o);
-            case Types.CHAR:
             case Types.DOUBLE:
             case Types.FLOAT:
                 return getDouble(o);
             case Types.DECIMAL:
             case Types.NUMERIC:
                 return o.asDecimal128().decimal128Value().bigDecimalValue();
+            case Types.CHAR:
             case Types.LONGNVARCHAR:
             case Types.LONGVARCHAR:
             case Types.NCHAR:
@@ -160,7 +163,6 @@ public class MongoSQLResultSet extends MongoResultSet<BsonDocument> implements R
             case Types.DATE:
             case Types.DISTINCT:
             case Types.JAVA_OBJECT:
-            case Types.LONGVARBINARY:
             case Types.NCLOB:
             case Types.REF:
             case Types.REF_CURSOR:
@@ -170,7 +172,6 @@ public class MongoSQLResultSet extends MongoResultSet<BsonDocument> implements R
             case Types.TIME:
             case Types.TIME_WITH_TIMEZONE:
             case Types.TIMESTAMP_WITH_TIMEZONE:
-            case Types.VARBINARY:
                 // not supported
                 break;
         }
--- a/src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
+++ b/src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
@@ -144,10 +144,10 @@ public class MongoSQLResultSet extends MongoResultSet<BsonDocument> implements R
                 if (o.getBsonType() == BsonType.NULL) {
                     return null;
                 }
-                // These types are wrapped in MongoSQLValue so that if
-                // they are stringified via toString() they will be
-                // represented by extended JSON.
-                return new MongoSQLValue(o);
+                // These types are wrapped in MongoSQLBsonValue so that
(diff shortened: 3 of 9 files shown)


src/main/java/com/mongodb/jdbc/ExtJsonValue.java
```diff
@@ -0,0 +1,109 @@
...
+            case END_OF_DOCUMENT:
+            case JAVASCRIPT:
+            case JAVASCRIPT_WITH_SCOPE:
+            case MAX_KEY:
+            case MIN_KEY:
+            case REGULAR_EXPRESSION:
+            case SYMBOL:
+            case TIMESTAMP:
+            default:
+                return "<invalid>"; // TODO - what should we do for these types?
```
mattChiaravalloti, line 92: This is the biggest question: what should we do with these other types? In `getString()` these cases throw exceptions. We don't have that option here, so I'm not sure what we want to do. This string is just a placeholder until we decide. We could default to `this.v.toString()` for these for now, pending what we decide as part of the design process.

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -35,10 +35,12 @@
     protected static int INT_COL = 6;
     // foo.d
     protected static int ANY_COL = 7;
+    // foo.doc
+    protected static int DOC_COL = 8;
```
mattChiaravalloti, line 39: the labels have to be alphabetically ordered, which is why this is the new 8 as opposed to being introduced as 10

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -262,17 +277,40 @@ void testGetStringValues() throws Exception {
...
         assertEquals("b", mongoSQLResultSet.getString(STRING_COL_LABEL));
+        assertNull(mongoSQLResultSet.getString(INT_NULLABLE_COL_LABEL));
         assertNull(mongoSQLResultSet.getString(NULL_COL));
         assertEquals("4", mongoSQLResultSet.getString(INT_COL_LABEL));
-        assertEquals(null, mongoSQLResultSet.getString(ANY_COL_LABEL));
+        assertNull(mongoSQLResultSet.getString(ANY_COL_LABEL));
+        assertEquals("[5, 6, 7]", mongoSQLResultSet.getString(ARRAY_COL_LABEL));
+        assertEquals("{\"c\": 5}", mongoSQLResultSet.getString(DOC_COL_LABEL));
+
+        // Check that getObject().toString() matches getString()
```
mattChiaravalloti, line 296: This is the primary place where I added new testing. By adding the document value `doc`, I am covering the two cases this currently primarily impacts: arrays and documents.

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
mattChiaravalloti: resolved offline: going to change these to use their extended json representation.

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
```diff
@@ -0,0 +1,109 @@
...
+            case DOUBLE:
+                return Double.toString(this.v.asDouble().getValue());
+            case INT32:
+                return Integer.toString(this.v.asInt32().getValue());
+            case INT64:
+                return Long.toString(this.v.asInt64().getValue());
+            case NULL:
+                return null;
+            case OBJECT_ID:
+                return this.v.asObjectId().getValue().toString();
```
nbagnard, line 75: I did a run a quick test and BsonObjectId.getValue().toString() does not return the relaxed extended Json representation.
BsonObjectId.getValue().toString() => 62478aa9cc79dc251e626f7d
Relaxed extended JSON representation (Same a canonical for oid) => {"$oid": "62478aa9cc79dc251e626f7d"}

Do we want a special treatment for OIDs? IMO the extended JSON representation would be good there because it makes it clear its not a simple string. I thought this was something we wanted to aim for, but I remember OIDs are a bit of a muddy ground so maybe I missed something.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
nbagnard: We should aim for covering all BSON types and on top of checking that getString and getObjectToString match, we should also make sure that the values are valid relaxed extended JSON. This can be done with something like `Document.parse("{\"val\":" + extendedJsonVal + "}")`. If the parsing does not fail, then it means extendedJsonVal is a valid extended json value.

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -266,6 +288,10 @@ BsonDocument generateRow() {
         array.add(new BsonInt32(3));
         foo.put(ARRAY_COL_LABEL, array);
 
+        BsonDocument fooSubDoc = new BsonDocument();
+        fooSubDoc.put(INT_COL_LABEL, new BsonInt32(5));
```
bucaojit, line 292: Would it be possible to add a BsonValue that has type information such as objectId, that would help confirm that the toString() is writing the extended json value.

bucaojit (changes requested):
Great work adding the ExtJsonValue class and making the toString() outputs consistent.

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
mattChiaravalloti: This was initially implemented before we decided this ticket would cover all types (not just object and array) so it does not have the work we discussed in standup on friday! will work on that today. @bucaojit I like that idea, but I will file a follow-up ticket to introduce that.

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * ExtJsonValue is a wrapper for BsonValue. The purpose of this class is to override the toString()
+ * method to produce the extended JSON representation of a BsonValue, rather than the java driver's
+ * default BsonValue.toString() output.
+ *
+ * <p>The driver's BsonValue class is abstract and intentionally cannot be extended by third
+ * parties. The driver explains this is to keep the BSON type system closed.
+ */
+public class ExtJsonValue {
```
rychipman, line 25: One drive-by request from me: could we make this something more generically named than `ExtJsonValue` (perhaps something like `MongoSQLValue`)? If we ever want to do something in the future like serialize these strings differently or make string serialization configurable, this class name might become outdated (but we won't be able to change it without it _technically_ being backwards breaking)

rychipman (changes requested):

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
mattChiaravalloti: now that we decided this ticket will cover stringifying all types as extended json, going to update the unit tests to cover all types

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
mattChiaravalloti: sure thing. I wasn't really sure what name would be good here. I'll make this change today.

nbagnard: The proposed name on the design doc was "MongoSQLBsonValue" but "MongSQLValue" is just as good.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/IntegrationTestUtils.java
```diff
@@ -673,18 +675,23 @@ public static boolean compareRow(List<Object> expectedRow, ResultSet actualRow)
...
                                             + expected_obj
                                             + " but is "
                                             + actual_obj
                                             + " for column "
                                             + (i + 1));
                             return false;
                         }
+                    } else {
+                        throw new IllegalArgumentException(
+                                "unsupported expected value class: " + expected_obj.getClass());
```
mattChiaravalloti, line 694: @bucaojit I left the previous case as `else if (expected_obj instanceof BsonValue)` so we can throw this more meaningful exception in case someone someday adds a test that has some other class as an expected result.

src/main/java/com/mongodb/jdbc/MongoSQLValue.java
```diff
@@ -0,0 +1,108 @@
...
+                StringWriter w = new StringWriter();
+                c.encode(
+                        new NoCheckStateJsonWriter(w, JSON_WRITER_SETTINGS),
+                        this.v,
+                        ENCODER_CONTEXT);
+                w.flush();
+                return w.toString();
+            case END_OF_DOCUMENT:
+            default:
+                return "<invalid>";
```
mattChiaravalloti, line 91: Still not sure what to do for `END_OF_DOCUMENT` (the `default` is technically unreachable since the switch statement exhaustively matches all variants of the `BsonType` enum. Should I return a static string like I have here, or perhaps `this.v.toString()`, or should we throw an `IllegalStateException` or other runtime exception?

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -284,4 +361,239 @@ MongoJsonSchemaResult generateSchema() {
         schemaResult.schema.mongoJsonSchema = generateMongoJsonSchema();
         return schemaResult;
     }
+
+    protected static MongoJsonSchema generateMongoJsonSchemaAllTypes() {
```
mattChiaravalloti, line 365: I added a "collection" to the mock that contains all bson types. I added this `generateMongoJsonSchemaAllTypes` method and the `generateRowAllTypes` method so we could create a mock result set that has all types. I only use this "collection" for testing the `getString()` and `getObject().toString()` behavior since that is all that is really in scope for this ticket.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
```diff
@@ -37,17 +37,17 @@ void initMocks() {
 
     @Test
     void testGetColumnCount() throws SQLException {
-        assertEquals(9, MongoSQLResultSetMetaDataTest.resultSetMetaData.getColumnCount());
+        assertEquals(10, MongoSQLResultSetMetaDataTest.resultSetMetaData.getColumnCount());
```
mattChiaravalloti, line 40: Now that I added the "all types collection", should I rollback these initial test updates that simply added a document to the previous mock collection? Technically this change is now outside the scope of this ticket and doesn't add any new coverage for work done _in this ticket_, but does add some additional coverage we didn't previously have in the unit tests. Since it is already written, I'd gladly leave it. But I can also take it away very easily.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -285,6 +320,186 @@ void testGetStringValues() throws Exception {
...
+        // Test getObject().toString() result for each BSON type
+        assertEquals("1.0", mongoSQLResultSetAllTypes.getObject(ALL_DOUBLE_COL_LABEL).toString());
+        assertEquals("str", mongoSQLResultSetAllTypes.getObject(ALL_STRING_COL_LABEL).toString());
+        assertEquals(
+                "{\"x\": 10, \"y\": {\"$oid\": \"" + ALL_OBJECT_ID_VAL.toString() + "\"}}",
+                mongoSQLResultSetAllTypes.getObject(ALL_DOC_COL_LABEL).toString());
+        assertEquals(
+                "[7, 8, 9]", mongoSQLResultSetAllTypes.getObject(ALL_ARRAY_COL_LABEL).toString());
+        // TODO: Java driver doesn't seem to serialize binary using extended json
+        //        assertEquals("{\"$binary\": {\"base64\": \"\", \"subType\": \"0\"}}", mongoSQLResultSetAllTypes.getObject(ALL_BINARY_COL_LABEL).toString());
```
mattChiaravalloti, line 406: Still need to look into why this is happening. Even using the java driver's extended json serialization methods results in `BsonBinary` being stringified in the wrong format:
```
Expected : {"$binary": {"base64": "", "subType": "0"}}
Actual   : BsonBinary{type=0, data=[]}
```
Could be a misunderstanding of the java driver api, or could be a bug in the java driver. My current guess is it is the latter case, since the binary codepath is the same as all the types.OTHER bson types' codepath and they all get stringified as ext json.

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
mattChiaravalloti: done:
1. PR updated to have extended json stringification for all types.OTHER types, and
2. filed [SQL-813](https://jira.mongodb.org/browse/SQL-813) to follow-up on Oliver's suggestion

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
mattChiaravalloti: done

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
mattChiaravalloti: done, but could still change this to `MongoSQLBsonValue`. Sorry I missed that in the design 😓

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
mattChiaravalloti: Oh I see. Obvious mistake. The `getObject()` method returns a `BsonBinary` for binary types because bson binary maps to `Types.BINARY`, not `Types.OTHER`. Easy fix.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -285,6 +320,186 @@ void testGetStringValues() throws Exception {
...
+
+        // Note that getObject() returns null for NULL and UNDEFINED BSON values, so we check
+        // manually that their stringification returns what is expected.
+        assertNull(mongoSQLResultSetAllTypes.getObject(ALL_UNDEFINED_COL_LABEL));
+        assertNull(new MongoSQLValue(new BsonUndefined()).toString());
+        assertNull(mongoSQLResultSetAllTypes.getObject(ALL_NULL_COL_LABEL));
+        assertNull(new MongoSQLValue(new BsonNull()).toString());
+
+        // Test getObject().toString() is valid relaxed extended JSON
+        // TODO
```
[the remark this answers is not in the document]
nbagnard: I don't think that we are testing that the value are valid extended JSON value which are strictly matching the extended JSON spec. We are only testing that both getString and getObject.toString() are the same. If we were testing that the values are valid extended JSON, then you would see a failure for the Timestamp.
Try this out :
long unixTime = System.currentTimeMillis();
BsonDateTime timestamp = new BsonDateTime(unixTime);
Date date = new Date(timestamp.getValue());
extendedJsonVal = printExtendedJsonValue(timestamp, JsonMode.RELAXED);
System.out.println(dateFormat.format(date));
System.out.println(extendedJsonVal);

You'll notice that the 2 string representations are different.

Also, the date from timestamp.getValue() is also not in UTC even though the timezone implies it is, while the extended Json value is in UTC. I'll file a separate ticket for this.

src/main/java/com/mongodb/jdbc/MongoSQLValue.java
```diff
@@ -0,0 +1,108 @@
...
+            return null;
+        }
+
+        switch (this.v.getBsonType()) {
+            case BOOLEAN:
+                return this.v.asBoolean().getValue() ? "true" : "false";
+            case DATE_TIME:
+                Date d = new Date(this.v.asDateTime().getValue());
+                return dateFormat.format(d);
+            case DECIMAL128:
```
nbagnard, line 52: Same as for Date. This will not return the relaxed extended JSON representation.
The value we provide is : _121684.45864468_
The relaxed extended JSON representation is : _{"$numberDecimal": "121684.45864468"}_

nbagnard (changes requested):

src/main/java/com/mongodb/jdbc/MongoSQLValue.java
```diff
@@ -0,0 +1,108 @@
...
+    @Override
+    public String toString() {
+        if (this.v == null) {
+            return null;
+        }
+
+        switch (this.v.getBsonType()) {
+            case BOOLEAN:
+                return this.v.asBoolean().getValue() ? "true" : "false";
+            case DATE_TIME:
```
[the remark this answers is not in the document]
mattChiaravalloti: Oh, I just misunderstood. I was under the impression we just wanted the extended json string representation for types that map to Types.OTHER, not for all types. I can make the update to do extended json for all types.

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
mattChiaravalloti: @nbagnard I'm not sure I understand what your suggestion is here. What specific test methods are you suggesting I update/remove/addd? Can you provide an example of what you're envisioning?

src/test/java/com/mongodb/jdbc/MongoSQLResultSetMetaDataTest.java
mattChiaravalloti: Yea it looks like the initial tests for result set included only types that didn't map to types.OTHER, but that's just speculation.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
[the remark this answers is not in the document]
mattChiaravalloti: Yea sorry, noted on another thread: I didn't know we wanted to change getString() for BSON types that correspond to non-Types.OTHER JDBC types. However, even after we update to make sure getString()/getObject().toString() return extended json representations for all types, I'm not sure what additional coverage this additional test will provide.

We'll have the following assertions for every bson type:
```
assertEquals("{$oid: ...}", rs.getString(OID_COL));
assertEquals("1.2", rs.getString(DOUBLE_COL));
assertEquals("{$date: ...}, rs.getString(DATE_COL));
... // etc

assertEquals("$oid: ...}", rs.getObject(OID_COL).toString());
assertEquals("1.2", rs.getObject(DOUBLE_COL).toString());
assertEquals("{$date: ...}, rs.getObject(DATE_COL).toString());
// etc
```
Two assertions for each bson type where the expected string is the extended json representation. Since the expected string is already the extended json string representation, isn't that sufficient to show that the output of getString()/getObject().toString() is _valid_ extended json? It seems the additional test to ensure the output is valid would really just be testing that our expected strings in the other tests are valid.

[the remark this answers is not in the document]
nbagnard: It would if we could be 100% sure that the string we put there is the correct extended json representation and using an extended Json compatible encoder works to reconstruct the BsonValue. As you can see, it is not the case for date_time and Decimal128 right now. Let's see after we discuss if we are indeed using the extended representation for every data type or if the others in the team also thought this was just for data types mapping to Other. After that, we can see how to make maybe change this test slightly to cover both it if needed.

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
mattChiaravalloti: done. the code turned out slightly longer in each case: 96 lines (new) vs. 95 (old) for schema, and 83 lines (new) vs 36 (old) for data, but we got to cut the comment out from the schema method and I think it should be more readable now. lmk what you think!

src/main/java/com/mongodb/jdbc/MongoSQLValue.java
mattChiaravalloti: Updated to return `v.toString()`

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -388,36 +339,12 @@ protected String getString(BsonValue o) throws SQLException {
...
-            case ARRAY:
-                Codec codec = CODEC_REGISTRY.get(BsonArray.class);
-                StringWriter writer = new StringWriter();
-                codec.encode(
-                        new NoCheckStateJsonWriter(writer, JSON_WRITER_SETTINGS),
-                        o.asArray(),
-                        ENCODER_CONTEXT);
-                writer.flush();
-                return writer.toString();
             case BINARY:
```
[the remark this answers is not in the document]
mattChiaravalloti: done (updated `getString` to support all types)

src/main/java/com/mongodb/jdbc/MongoSQLValue.java
mattChiaravalloti: done

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -138,7 +119,7 @@ protected Object getObject(BsonValue o, int columnType) throws SQLException {
             case Types.BIGINT:
                 return getInt(o);
             case Types.BINARY:
-                return o.asBinary();
+                return new MongoSQLValue(o.asBinary());
```
mattChiaravalloti, line 122: It does feel a bit strange to only wrap _some_ types in `getObject` with `MongoSQLValue`. It may not be totally clear to future maintainers why some are wrapped (Binary, Decimal, Other, Timestamp) and some are not (all others that do not throw an exception). The distinction right now is that anything that needs special extended json stringification is wrapped. Not sure if that's the best criteria. Maybe we should wrap everything?

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
mattChiaravalloti: Yea, maybe part of the JDBC Test Infrastructure epic could be cleaning up some of the clutter we introduced to the unit tests over the last few months 👍

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -285,6 +320,192 @@ void testGetStringValues() throws Exception {
                 mongoSQLResultSet.getUnicodeStream(STRING_COL_LABEL).read(new byte[100], 0, 100));
     }
 
+    @Test
+    public void testGetStringAllTypes() throws Exception {
```
nbagnard, line 324: I looked into this a bit more and I have a suggestion which I hope can make both of us happy.
I agree that adding an extra test to verify that the String is a valid extended JSON value seems a bit too much.
What we could do is : 
- Extract the values out in separate variables
- Use these snippets in generateRowAllTypes
- Use the them snippet as the expected value

This way, we know that the values are valid extended JSON because they were parsed without any issue by generateRowAllTypes , we verify it's the correct expected value for getString and because testGetObjectToStringMatchesGetString checks that getString == getObject.toString we are good all the way.

Below is a partial example :
`    
public abstract class MongoSQLMock {
...
protected static final String ALL_DOUBLE_COL_VAL = "5";
    protected static final String ALL_TIMESTAMP_COL_VAL = "{\"$timestamp\": {\"t\": 1412180887, \"i\": 1}}";
 ....
    static BsonDocument generateRowAllTypes() {
        String doc =
                    "{"
                        + "    \"all\": {"
                         ....
                         + "        \"timestamp\": "
                        +  ALL_TIMESTAMP_COL_VAL
                        + ","
                        + "        \"long\": "
                        + ALL_DOUBLE_COL_VAL
                        + ","
                       ....
    }
...
}

--------------------------------------------------------------------

class MongoSQLResultSetTest extends MongoSQLMock {
....
    @Test
    public void testGetStringAllTypes() throws Exception {
    ....
            assertEquals( 
                ALL_TIMESTAMP_COL_VAL,
                mongoSQLResultSetAllTypes.getString(ALL_TIMESTAMP_COL_LABEL));
            assertEquals(
               ALL_DOUBLE_COL_VAL, 
               mongoSQLResultSetAllTypes.getString(ALL_LONG_COL_LABEL));
    ....
    }
...
}
`

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -284,4 +358,191 @@ MongoJsonSchemaResult generateSchema() {
...
+        String schema =
+                ""
```
nbagnard, line 364: Nit: I think this empty string can be safely removed. Or does it come from Spotless?

src/test/java/com/mongodb/jdbc/MongoSQLMock.java
```diff
@@ -284,4 +358,191 @@ MongoJsonSchemaResult generateSchema() {
...
+                        + "}";
+
+        return MongoDriver.registry
+                .get(MongoJsonSchema.class)
+                .decode(new JsonReader(schema), DecoderContext.builder().build());
+    }
+
+    static BsonDocument generateRowAllTypes() {
+        String doc =
+                ""
```
nbagnard, line 464: Nit: I think this empty string can be safely removed. Or does it come from Spotless?

src/main/java/com/mongodb/jdbc/MongoSQLValue.java
```diff
@@ -0,0 +1,104 @@
...
+ * <p>The driver's BsonValue class is abstract and intentionally cannot be extended by third
+ * parties. The driver explains this is to keep the BSON type system closed. Therefore, this class
+ * does not extend BsonValue, instead it contains a BsonValue member.
+ */
+public class MongoSQLValue {
+    static final JsonWriterSettings JSON_WRITER_SETTINGS =
+            JsonWriterSettings.builder().outputMode(JsonMode.RELAXED).build();
+    static final EncoderContext ENCODER_CONTEXT = EncoderContext.builder().build();
+
+    // dateFormat cannot be static due to a threading bug in the library.
```
nbagnard, line 25: I think this is probably unused now, isn't it?

nbagnard (changes requested):
Looking good!
I only have a few comments (clean-up stuff) and one suggestion for making sure the produced values are extended JSON :)

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
mattChiaravalloti: done.

src/main/java/com/mongodb/jdbc/ExtJsonValue.java
mattChiaravalloti: Updated to be consistent with design doc.

nbagnard:
(Putting in the right PR)
ah, bummer, I looked at the spec again because I was surprised that all integer related types are mapped to getInt and I think that there are a couple of types for which getObject().toString will not match getString if we want to match the JDBC spec.
See B.3 JDBC Types Mapped to Java Object Types from the spec:

    DECIMAL/NUMERIC => java.math.BigDecimal
    BINARY/VARBINARY/LONGVARBINARY => byte[]
    TIMESTAMP => java.sql.Timestamp

I would still keep the getString() value to be the extended JSON string though because I did not find anything regarding what getString is supposed to return. The spec only says that it should return a value for all mainstream SQL/JDBC types (we were not obligated to return a String for OTHER for example according to the spec, but it looks like they just missed it, same for Null).

It seems safer to follow the spec for getObject, just in case someone's application uses getObject unconditionally and then work directly on the objects with the assumption it match the spec and that getObject for a TIMESTAMP return a java.sql.timestamp for example.
WDYT?

mattChiaravalloti:
> (Putting in the right PR) ah, bummer, I looked at the spec again because I was surprised that all integer related types are mapped to getInt and I think that there are a couple of types for which getObject().toString will not match getString if we want to match the JDBC spec. See B.3 JDBC Types Mapped to Java Object Types from the spec:
> 
> ```
> DECIMAL/NUMERIC => java.math.BigDecimal
> BINARY/VARBINARY/LONGVARBINARY => byte[]
> TIMESTAMP => java.sql.Timestamp
> ```
> 
> I would still keep the getString() value to be the extended JSON string though because I did not find anything regarding what getString is supposed to return. The spec only says that it should return a value for all mainstream SQL/JDBC types (we were not obligated to return a String for OTHER for example according to the spec, but it looks like they just missed it, same for Null).
> 
> It seems safer to follow the spec for getObject, just in case someone's application uses getObject unconditionally and then work directly on the objects with the assumption it match the spec and that getObject for a TIMESTAMP return a java.sql.timestamp for example. WDYT?

discussed offline -- I updated the `getObject` method to return the instances of the appropriate classes for the ones you listed above (although `VARBINARY` AND `LONGVARBINARY` are still left as invalid for now. I believe there is another ticket to update the JDBC type for BSON binary).

This is now implemented and up to date!

bucaojit (approved):
LGTM!  
Incredible work, and great to see all the test improvements. 

There is one unit test failing for the date column: 
```
AssertionFailedError: expected: <2020-12-25 12:13:14.0> but was: <2020-12-25 17:13:14.0>
```

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -285,6 +320,175 @@ void testGetStringValues() throws Exception {
...
+                ALL_TIMESTAMP_COL_VAL,
+                mongoSQLResultSetAllTypes.getString(ALL_TIMESTAMP_COL_LABEL));
+        assertEquals(
+                ALL_DECIMAL_COL_VAL, mongoSQLResultSetAllTypes.getString(ALL_DECIMAL_COL_LABEL));
+        assertEquals(
+                ALL_MIN_KEY_COL_VAL, mongoSQLResultSetAllTypes.getString(ALL_MIN_KEY_COL_LABEL));
+        assertEquals(
+                ALL_MAX_KEY_COL_VAL, mongoSQLResultSetAllTypes.getString(ALL_MAX_KEY_COL_LABEL));
+
+        // Note that for "string" and "long", the ALL_X_COL_VAL must use extended json style-syntax in the
```
nbagnard, line 354: Using this code in the Mock will let use both canonical and relaxed extended Json represention :
Document d = Document.parse(_extendedJson_);
BsonDocument bsonDoc =  d.toBsonDocument(Document.class, MongoClient.getDefaultCodecRegistry());

For example, the snippet above accepts {"val": 1} and {"val": {"$numberInt": "1"}} as the _extendedJson_ value.

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -132,136 +114,65 @@ protected Object getObject(BsonValue o, int columnType) throws SQLException {
...
-                return o.asBinary();
+                return o.asBinary().getData();
             case Types.BIT:
-                return getBoolean(o);
-            case Types.BLOB:
-                // not supported
-                break;
             case Types.BOOLEAN:
                 return getBoolean(o);
             case Types.CHAR:
```
nbagnard, line 127: Looks you missed moving CHAR with LONGVARCHAR, VARCHAR, etc....

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -132,136 +114,65 @@ protected Object getObject(BsonValue o, int columnType) throws SQLException {
...
-                // not supported
-                break;
-            case Types.TIMESTAMP:
-                return getTimestamp(o);
             case Types.TIMESTAMP_WITH_TIMEZONE:
-                // not supported
-                break;
-            case Types.TINYINT:
-                return getInt(o);
             case Types.VARBINARY:
```
nbagnard, line 173: Could we move VARBINARY and LONGVARBINARY with BINARY, the same way we group BIGINT, INT, SMALLINT, TINYINT and CHAR, VARCHAR, LONGVARCHAR?

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
```diff
@@ -132,136 +114,65 @@ protected Object getObject(BsonValue o, int columnType) throws SQLException {
             return null;
         }
         switch (columnType) {
-            case Types.ARRAY:
-                // not supported
-                break;
             case Types.BIGINT:
```
nbagnard, line 117: BIGINT should return a Long. Otherwise we might encounter overflow issues. Especially since we correctly map BSON_LONG to BIGINT.

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -285,6 +320,175 @@ void testGetStringValues() throws Exception {
...
+                ALL_SYMBOL_COL_VAL,
+                mongoSQLResultSetAllTypes.getObject(ALL_SYMBOL_COL_LABEL).toString());
+        assertEquals(
+                ALL_JAVASCRIPT_WITH_SCOPE_COL_VAL,
+                mongoSQLResultSetAllTypes
+                        .getObject(ALL_JAVASCRIPT_WITH_SCOPE_COL_LABEL)
+                        .toString());
+        assertEquals(
+                ALL_INT_COL_VAL, mongoSQLResultSetAllTypes.getObject(ALL_INT_COL_LABEL).toString());
+        assertEquals(
```
nbagnard, line 405: This should not pass because mongoSQLResultSetAllTypes.getObject(ALL_TIMESTAMP_COL_LABEL) will return a java.sql.Timestamp and toString on it should return yyyy-mm-dd hh:mm:ss.fffffffff, not an extended JSON value like {"$timestamp": {"t": 1412180887, "i": 1}}.
I am a little surprised Evergreen is happy. Could you double-check? I am not sure what I missed here ?🧐

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
```diff
@@ -285,6 +320,175 @@ void testGetStringValues() throws Exception {
...
+                mongoSQLResultSetAllTypes.getObject(ALL_JAVASCRIPT_COL_LABEL).toString());
+        assertEquals(
+                mongoSQLResultSetAllTypes.getString(ALL_SYMBOL_COL_LABEL),
+                mongoSQLResultSetAllTypes.getObject(ALL_SYMBOL_COL_LABEL).toString());
+        assertEquals(
+                mongoSQLResultSetAllTypes.getString(ALL_JAVASCRIPT_WITH_SCOPE_COL_LABEL),
+                mongoSQLResultSetAllTypes
+                        .getObject(ALL_JAVASCRIPT_WITH_SCOPE_COL_LABEL)
+                        .toString());
+        assertEquals(
```
nbagnard, line 481: Same here, I am not sure how this assert passes since getString should return {"$timestamp": {"t": 1412180887, "i": 1}} and getObject.toString should return something like "2020-12-25 12:13:14.0" (value formatted with yyyy-mm-dd hh:mm:ss.fffffffff)

nbagnard (changes requested):

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
nbagnard: Got it, thanks Matt for the extra details on the fun of Bson.Timestamp :)

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
mattChiaravalloti: I can make this change, but does this really matter? The only JDBC types that this method will ever encounter are the ones associated with BSON types in `BsonTypeInfo`. `BsonTypeInfo` does not have any reference to `Types.VARBINARY` or `Types.LONGVARBINARY`.

(The same thing is true for `Types.SMALLINT`, `Types.TINYINT`, `Types.BOOLEAN`, `Types.FLOAT`, `Types.Numeric`, `Types.CHAR`, `Types.VARCHAR`, `Types.LONGNVARCHAR`, `Types.NCHAR`, `Types.NVARCHAR`, and `Types.Real`, actually).

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
mattChiaravalloti: Discussed offline: bson timestamp != bson date

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
mattChiaravalloti: discussed offline: bson timestamp != bson date

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
nbagnard: No, it does not matter technically until we change any mapping. But it makes sense to at least have the same behaviour for all types with either only providing a value for the SQL type we support or always try to get the value with the getter for the "group". Otherwise, I feel like it opens the door for questions about why treating String related types differently than binary related type? Even if the answer is "Why not" :P

src/test/java/com/mongodb/jdbc/MongoSQLResultSetTest.java
mattChiaravalloti: discussed offline: I updated `MongoSQLBsonValue` to stringify all BsonValues using relaxed extended json. We will discuss in standup whether this is the behavior we want for bson strings, which will now be delimited with `"` characters.

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
mattChiaravalloti: done

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
mattChiaravalloti: good enough haha, updated!

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
mattChiaravalloti: done

src/main/java/com/mongodb/jdbc/MongoSQLResultSet.java
nbagnard: thanks :)

nbagnard (approved):
LGTM once we decide what we want to do for String values.
I see you already made the change to have them be extended JSON, so you'll either merge as is or simply revert back, in both case, no need to re-request review :)
Thank you for going through all of this. So many small details to take into account.

mattChiaravalloti:
evergreen merge

mattChiaravalloti:
evergreen merge

mattChiaravalloti:
just realized the commit queue task went purple the other day (system failure). resubmitted.

mattChiaravalloti:
@rychipman It looks like this cannot be merged until you approve since your last comment moved you to a "Requested Changes" state. (I attempted to remove you as a reviewer, but that moved you from the yellow circle to the red box). From evergreen:
> 405 1 review requesting changes and 2 approving reviews by reviewers with write access.

rychipman (approved):
LGTM!

mattChiaravalloti:
evergreen merge

PR #101, opened 2022-04-06
SQL-791: Adding smoke test for built artifacts

bucaojit:
JIRA: https://jira.mongodb.org/browse/SQL-791

Added a smoke test that will block the publish-maven task if it fails.
Starts a local ADL to connect to, retrieve metadata, and run a query.
Smoke test uses the built shadowJar  `mongodb-jdbc-'+getReleaseVersion()+'-all.jar` as its dependency.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -127,7 +127,7 @@ functions:
         ./resources/run_adl.sh start &&
         ./gradlew runDataLoader &&
         ./gradlew clean shadowJar &&
-        ./gradlew :smoketest:test
+        ./gradlew :smoketest:test -Psmoketest
         EXITCODE=$?
         ./resources/run_adl.sh stop
         exit $EXITCODE
--- a/.evg.yml
+++ b/.evg.yml
@@ -36,12 +36,14 @@ buildvariants:
       - name: "build"
       - name: "test-unit"
       - name: "test-integration"
+
   - name: release
     display_name: "Release"
     expansions:
       _platform: ubuntu1804-64-jdk-8
     run_on: ubuntu1804-build
     tasks:
+      - name: "test-smoke"
       - name: "publish-maven"
 
 tasks:
@@ -57,7 +59,14 @@ tasks:
     commands:
       - func: "run integration test"
 
+  - name: "test-smoke"
+    commands:
+      - func: "run smoke test"
+
   - name: "publish-maven"
+    depends_on:
+      - name: "test-smoke"
+        variant: "release"
     commands:
       - func: "publish maven"
 
@@ -107,6 +116,21 @@ functions:
           EXITCODE=$?
           ./resources/run_adl.sh stop
           exit $EXITCODE
+
+  "run smoke test":
+    command: shell.exec
+    type: test
+    params:
+      working_dir: mongo-jdbc-driver
+      script: |
+        ${PREPARE_SHELL}
+        ./resources/run_adl.sh start &&
+        ./gradlew runDataLoader &&
+        ./gradlew clean shadowJar &&
+        ./gradlew :smoketest:run
+        EXITCODE=$?
+        ./resources/run_adl.sh stop
+        exit $EXITCODE
 
   "export variables":
     - command: shell.exec
--- a/.evg.yml
+++ b/.evg.yml
@@ -127,7 +127,7 @@ functions:
         ./resources/run_adl.sh start &&
         ./gradlew runDataLoader &&
         ./gradlew clean shadowJar &&
-        ./gradlew :smoketest:run
+        ./gradlew :smoketest:test
         EXITCODE=$?
         ./resources/run_adl.sh stop
         exit $EXITCODE
--- a/smoketest/src/main/java/com/mongodb/jdbc/smoketest/SmokeTest.java
+++ b/smoketest/src/main/java/com/mongodb/jdbc/smoketest/SmokeTest.java
@@ -1,0 +1,70 @@
+package com.mongodb.jdbc.smoketest;
+
+import java.sql.Connection;
+import java.sql.DatabaseMetaData;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+
+/**
+ * SmokeTest runs a test on built artifacts to verify that connection,
+ * metadata retrieval, and querying is successful
+ */
+public class SmokeTest {
+    static final String URL = "jdbc:mongodb://localhost";
+    static final String DB = "integration_test";
+    public static final String MONGOSQL = "mongosql";
+
+    public static Connection getBasicConnection(String url, String db)
+            throws SQLException {
+        Properties p = new java.util.Properties();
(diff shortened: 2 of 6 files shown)


.evg.yml
```diff
@@ -108,6 +117,21 @@ functions:
...
           exit $EXITCODE
 
+  "run smoke test":
+    command: shell.exec
+    type: test
+    params:
+      working_dir: mongo-jdbc-driver
+      script: |
+        ${PREPARE_SHELL}
+        ./resources/run_adl.sh start &&
```
bucaojit, line 127: @nbagnard We discussed about adding the start/stop adl as a pre/post functions, I found that the pre/post steps run for all tasks.  Also that tasks are independent of each other, such as being started in different directories.  Starting it in the task is the way to run the local adl.

nbagnard: I see. Yes, it makes sense.

.evg.yml
```diff
@@ -36,12 +36,14 @@ buildvariants:
...
       - name: "test-integration"
+
   - name: release
     display_name: "Release"
     expansions:
       _platform: ubuntu1804-64-jdk-8
     run_on: ubuntu1804-build
     tasks:
+      - name: "test-smoke"
       - name: "publish-maven"
```
nbagnard, line 47: I looked at the logs from the Evergreen run and publish-maven is rebuilding the jars.
 [2022/04/06 16:21:58.327] Tasks to be executed: [task ':compileJava', task ':processResources', task ':classes', task ':jar', task ':generateMetadataFileForMavenJavaPublication', task ':generatePomFileForMavenJavaPublication', task ':initializeSonatypeStagingRepository', task ':javadoc', task ':javadocJar', task ':shadowJar', task ':sourceJar', task ':signMavenJavaPublication', task ':publishMavenJavaPublicationToSonatypeRepository', task ':publishToSonatype', task ':publishMaven', task ':demo:publishMaven', task ':smoketest:publishMaven']
 
 I also see that it is trying to run "smoketest:publishMaven".
 I am wondering if we could have the Gradle handle this all the way through and not have to run a special task from Evergreen?
 Because right now, the jar we smoke test and the one we publish are not the same ones. They are built the same way and I see little possibility of mishap, but still, I would prefer to smoke test the "final" artifacts.
 
 Maybe in deploy.gradle, we could have for both snapshot and releases a dependency on smoketest?
 Or I am thinking if the jars files are "ready" when the publish task runs that it will not rebuild them?
 Right now, the logs show something hinting that it's not the case :
 ` [2022/04/06 16:22:34.528] Build cache key for task ':shadowJar' is b641979972b8ba0cf1a9212a3614588c

 [2022/04/06 16:22:34.528] Task ':shadowJar' is not up-to-date because:

 [2022/04/06 16:22:34.528]   No history is available.

 [2022/04/06 16:22:34.528] *******************`

smoketest/src/main/java/com/mongodb/jdbc/smoketest/SmokeTest.java
```diff
@@ -0,0 +1,70 @@
...
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Properties;
+
+/**
+ * SmokeTest runs a test on built artifacts to verify that connection,
+ * metadata retrieval, and querying is successful
+ */
+public class SmokeTest {
```
mattChiaravalloti, line 15: Why make this an executable with a `main` instead of a junit test with two test methods?

smoketest/src/main/java/com/mongodb/jdbc/smoketest/SmokeTest.java
```diff
@@ -0,0 +1,70 @@
...
+        p.setProperty("user", System.getenv("ADL_TEST_LOCAL_USER"));
+        p.setProperty("password", System.getenv("ADL_TEST_LOCAL_PWD"));
+        p.setProperty("authSource", System.getenv("ADL_TEST_LOCAL_AUTH_DB"));
+        p.setProperty("ssl", "false");
+        p.setProperty("database", db);
+        return DriverManager.getConnection(URL, p);
+    }
+
+    public static void databaseMetadataTest(DatabaseMetaData dbMetadata) throws SQLException {
+        ResultSet rs = dbMetadata.getColumns(null, "%", "%", "%");
```
mattChiaravalloti, line 33: should we add smoke-test-specific data to the `resources/integration-test/testdata` directory so we can safely edit the existing data as needed for integration tests without breaking these tests? This also applies to the `queryTest` method below which queries `class` and expects 5 rows.

mattChiaravalloti (changes requested):

smoketest/src/main/java/com/mongodb/jdbc/smoketest/SmokeTest.java
nbagnard: That's seems a little bit overkill, no?
To be fair, we probably don't need to check the actual row count, making sure that it's over 1 let us check that we can retrieve data and in this case even exhaust a resultset without problem. Smoke test should be light, just making sure the produced lib has no issues and can load/be used with no errors.

mattChiaravalloti: Checking that it is just greater than 1 is a good idea. I just don't want these tests to be tied so specifically to data that isn't strictly for them, but just ensuring the result set is non-empty is a great way of doing that without introducing a new dataset.

bucaojit: I was debating on whether to put a greater than equals when I was initially writing this.   A greater than 1 sounds good.  I'll make that change.

.evg.yml
bucaojit: Not quite sure why it runs the smoketest:publishMaven task.  But yes, it's a different jar file since they are in different tasks.  I was thinking as they are built in the same manner with the shadowJar that it would still be valid.  I will see what can be done in gradle to test the same jar file.

nbagnard: Not sure why either, it seems that it runs all the project tasks.
Yes, the jars are built the same way, I think it just modifies or add the pom file, and sign the jar.
So, the chances of a problem are very, very low. If you don't find a way to run smoke test from maven publish or cache the jars which have been smoke tested easily, I'd say, don't worry and we can keep it the way it is done right now.

smoketest/src/main/java/com/mongodb/jdbc/smoketest/SmokeTest.java
bucaojit: That's a good idea, I switched it to junit tests

mattChiaravalloti (approved):

nbagnard (approved):
LGTM! One less thing to worry about when releasing the driver 🎉

.evg.yml
bucaojit: Created investigation ticket to look into whether the smoke test can be run on the same jars that are published: [SQL-852](https://jira.mongodb.org/browse/SQL-852)

bucaojit:
evergreen merge

PR #102, opened 2022-05-03
SQL-876: Propagate timeout from connection validation

nbagnard:
Timeout errors were not propagated properly and the connection validation was incorrectly successful leading to unclear errors when executing queries.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -436,14 +436,21 @@ public abstract class MongoConnection implements Connection {
         if (isClosed) {
             throw new SQLException("Connection is closed.");
         }
+
         // We use createStatement to test the connection. Since we are not allowed
         // to set the timeout adhoc on the calls, we use Executor to run a blocked call with timeout.
         ExecutorService executor = Executors.newCachedThreadPool();
         Future<Void> future = executor.submit(new ConnValidation());
-        if (timeout > 0) {
-            future.get(timeout, TimeUnit.SECONDS);
-        } else {
-            future.get();
+
+        try {
+            if (timeout > 0) {
+                future.get(timeout, TimeUnit.SECONDS);
+            } else {
+                future.get();
+            }
+        } finally {
+            future.cancel(true);
+            executor.shutdown();
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -421,38 +421,39 @@ public Void call() throws SQLException {
...
-            // handle the timeout
-            return false;
-        } catch (InterruptedException e) {
-            // handle the interrupt
-            return false;
-        } catch (ExecutionException e) {
-            // handle connection error
+            testConnection(timeout);
+        } catch (Exception ex) {
+            // Invalid connection, the dummy query failed
```
bucaojit, line 455: This would be a good spot to use the logger to log the exception string.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -421,38 +421,39 @@ public Void call() throws SQLException {
...
             throw new SQLException("Input is invalid.");
         }
 
         if (isClosed) {
-            return false;
+            throw new SQLException("Connection is closed.");
         }
         // We use createStatement to test the connection. Since we are not allowed
         // to set the timeout adhoc on the calls, we use Executor to run a blocked call with timeout.
         ExecutorService executor = Executors.newCachedThreadPool();
```
bucaojit, line 441: This executor is not being shutdown, we can add a `try { } finally { executor.shutdown(); }`

bucaojit (changes requested):
Nice work propagating the connection errors correctly. 

A few comments.

src/main/java/com/mongodb/jdbc/MongoConnection.java
nbagnard: oh, yes, added, thank you!

src/main/java/com/mongodb/jdbc/MongoConnection.java
nbagnard: Exceptions are logged "automatically" by the logging aspect when thrown in testConnection.
See LoggingAspect : after () throwing (Exception e) : execution(* *.*(..)) && !within(LoggingAspect)

pmeredit (approved):
LGTM!

bucaojit:
One of the unit tests, testBasicURL, is failing with `java.sql.SQLTimeoutException: Timeout. Can't connect.` 
I took a quick look and prior to the change the `isValid()` would also hit the exception and return false but the return was not being checked.  We may need to change the unit test to account for the exception.

nbagnard:
> One of the unit tests, testBasicURL, is failing with `java.sql.SQLTimeoutException: Timeout. Can't connect.`
> I took a quick look and prior to the change the `isValid()` would also hit the exception and return false but the return was not being checked. We may need to change the unit test to account for the exception.

Oh, this is an issue with the test. Unit tests are not supposed to try to connect "for real". I updated the test to call getUnvalidatedConnection as it should have been doing.

bucaojit (approved):
LGTM!

PR #103, opened 2022-05-13
SQL-883: Adding ClientInfo property to include client name/version to…

bucaojit:
… driver info

`ClientInfo` is an optional property that users can add a client name and version to the mongoDriverInformation, which is used to provide additional client metadata for connections.
Format is:
property["ClientInfo"]="name|version";

The information logged for "tableau-connector|f8807d3" will look like:
```
"clientMetadata": {"driver":{"name":"mongo-java-driver|sync|mongodb-jdbc|tableau-connector","version":"3.12.11|2.0.0-beta-2-6-gc915429|f8807d3"}
```

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -66,7 +66,12 @@ public abstract class MongoConnection implements Connection {
     private static Map<String, FileHandler> fileHandlers = new HashMap<String, FileHandler>();
     private String logDirPath;
 
-    public MongoConnection(ConnectionString cs, String database, Level logLevel, File logDir) {
+    public MongoConnection(
+            ConnectionString cs,
+            String database,
+            Level logLevel,
+            File logDir,
+            String[] clientInfo) {
         Preconditions.checkNotNull(cs);
         this.connectionId = connectionCounter.incrementAndGet();
         // Initializes a parent logger for the connection
@@ -87,14 +92,29 @@ public abstract class MongoConnection implements Connection {
         logger.log(
                 Level.INFO, "Connecting using " + MongoDriver.MONGOSQL_DRIVER_NAME + " " + version);
 
+        MongoDriverInformation mongoDriverInformation;
+        if (clientInfo != null && clientInfo.length == 2) {
+            MongoDriverInformation driverInfoWithClientInfo =
+                    MongoDriverInformation.builder()
+                            .driverName(clientInfo[0])
+                            .driverVersion(clientInfo[1])
+                            .build();
+            mongoDriverInformation =
+                    MongoDriverInformation.builder(driverInfoWithClientInfo)
+                            .driverName(MongoDriver.NAME)
+                            .driverVersion(version)
+                            .build();
+        } else {
+            mongoDriverInformation =
+                    MongoDriverInformation.builder()
+                            .driverName(MongoDriver.NAME)
+                            .driverVersion(version)
+                            .build();
+        }
+
         this.mongoClientSettings = MongoClientSettings.builder().applyConnectionString(cs).build();
-        mongoClient =
-                MongoClients.create(
-                        mongoClientSettings,
-                        MongoDriverInformation.builder()
-                                .driverName(MongoDriver.NAME)
-                                .driverVersion(version)
-                                .build());
+        mongoClient = MongoClients.create(mongoClientSettings, mongoDriverInformation);
+
         isClosed = false;
     }
 
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -1,7 +1,6 @@
 package com.mongodb.jdbc;
 
 import com.google.common.base.Preconditions;
-import com.mongodb.ConnectionString;
 import com.mongodb.MongoClientSettings;
 import com.mongodb.MongoDriverInformation;
 import com.mongodb.client.MongoClient;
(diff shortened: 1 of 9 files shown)


src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -66,7 +66,12 @@ public abstract class MongoConnection implements Connection {
     private static Map<String, FileHandler> fileHandlers = new HashMap<String, FileHandler>();
     private String logDirPath;
 
-    public MongoConnection(ConnectionString cs, String database, Level logLevel, File logDir) {
+    public MongoConnection(
+            ConnectionString cs,
+            String database,
+            Level logLevel,
+            File logDir,
+            String[] clientInfo) {
```
nbagnard, line 74: If we keep adding more and more connection properties to MongoConnection maybe we'll want to have a dedicated data structure for it. It seems that this is still Ok right now, but if you think it makes sense to switch to a data structure (a simple POJO will do), feel free to do it.

nbagnard (changes requested):

mattChiaravalloti (changes requested):
agreed with Natacha's suggestion regarding refactoring the builder code. I'll check back when that's addressed!

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -66,7 +66,12 @@ public abstract class MongoConnection implements Connection {
     private static Map<String, FileHandler> fileHandlers = new HashMap<String, FileHandler>();
     private String logDirPath;
 
-    public MongoConnection(ConnectionString cs, String database, Level logLevel, File logDir) {
+    public MongoConnection(
```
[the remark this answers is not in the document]
bucaojit: I made a change to add the driver name to appName and append `|<clientInfo>` if it exists.  
Example log line looks like:
```
2022-05-16T19:01:05.310-0700    ^[[34m[INFO]^[[0m       frontend.command        command execution complete      {"version": "built-without-version-string", "gitVersion": "built-without-git-spec", "pid": 10152, "connectionID": 209902546, "remoteIP": "127.0.0.1", "hostname": "localhost", "tenantID": "5ecd3e8aa1b9b07b3ddd1b4b", "dataLakeType": "UNKNOWN", "correlationID": "16efc1f96e543ff09579239d", "x-request-id": "16efc1f96e543ff09579239d", "commandName": "isMaster", "appName": "mongodb-jdbc|tableau-connector", "elapsed": "7µs", "elapsedNanos": 7000}
```
The way the code change is, if the user adds `appName` to their connection string then that will be used instead.

It's now in both appName and MongoDriverInformation.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -87,14 +92,29 @@ public MongoConnection(ConnectionString cs, String database, Level logLevel, Fil
         logger.log(
                 Level.INFO, "Connecting using " + MongoDriver.MONGOSQL_DRIVER_NAME + " " + version);
 
+        MongoDriverInformation mongoDriverInformation;
```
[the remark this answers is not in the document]
bucaojit: Made the first suggestion to keep the ordering of driver followed by client info.  Let me know if it is as expected.

src/main/java/com/mongodb/jdbc/MongoConnection.java
bucaojit: I added a POJO MongoConnectionProperties to address this.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -82,19 +82,35 @@ public MongoConnection(ConnectionString cs, String database, Level logLevel, Fil
...
+            mdiBuilder = MongoDriverInformation.builder(driverInfoWithClientInfo);
+        } else {
+            mdiBuilder = MongoDriverInformation.builder();
+        }
+        MongoDriverInformation mongoDriverInformation =
+                mdiBuilder.driverName(MongoDriver.NAME).driverVersion(version).build();
+
+        this.mongoClientSettings =
+                MongoClientSettings.builder()
+                        .applicationName(appName.toString())
```
nbagnard, line 109: The driver version and connector version will not be in `appName`. Is this intentional?

nbagnard: Also, just saw this while connecting with mongosh : appName=mongosh+1.2.2
What about we follow the convention and use something similar like `driverName+driverVersion|tableau+connectorVersion`?

bucaojit: Sounds good, I'll add the version to the appName.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -82,19 +82,35 @@ public MongoConnection(ConnectionString cs, String database, Level logLevel, Fil
...
-        this.mongoClientSettings = MongoClientSettings.builder().applyConnectionString(cs).build();
-        mongoClient =
-                MongoClients.create(
-                        mongoClientSettings,
-                        MongoDriverInformation.builder()
-                                .driverName(MongoDriver.NAME)
-                                .driverVersion(version)
-                                .build());
+        MongoDriverInformation.Builder mdiBuilder;
+        if (connectionProperties.getClientInfo() != null
```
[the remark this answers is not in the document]
bucaojit: Made the change to store `connectionProperties.getClientInfo()` in a variable here.

src/main/java/com/mongodb/jdbc/MongoConnectionProperties.java
```diff
@@ -0,0 +1,46 @@
...
+import com.mongodb.ConnectionString;
+import java.io.File;
+import java.util.logging.Level;
+
+public class MongoConnectionProperties {
+    public ConnectionString connectionString;
+    public String database;
+    public Level logLevel;
+    public File logDir;
+    public String[] clientInfo;
```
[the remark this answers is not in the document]
bucaojit: It was to match the other POJOs such as `JsonSchema` which has public variables.  I have changed them to `protected` and added setters.

src/main/java/com/mongodb/jdbc/MongoConnection.java
bucaojit: Here's an example output: `"appName": "mongodb-jdbc+2.0.0-beta-2-7-g1c59d6b-dirty|tableau-connector+f8807d3"`

src/main/java/com/mongodb/jdbc/MongoConnectionProperties.java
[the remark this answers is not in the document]
nbagnard: You don't need getters and setters if the fields are protected because you used them in the same package.
However, I think making them private and only allow setting the value via the constructor makes sense because all the info should be known at the time of creation and not modified.

nbagnard (approved):
Only a few things that you could polish out, but I leave it to you and Matt to decide if these are blockers.
The logic itself looks good to me 👍

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -59,6 +59,7 @@ public class MongoDriver implements Driver {
     static final String MONGODB_SRV_URL_PREFIX = JDBC + "mongodb+srv:";
     public static final String USER = "user";
     public static final String PASSWORD = "password";
+    public static final String CLIENT_INFO = "ClientInfo";
```
mattChiaravalloti, line 62:
```suggestion
    public static final String CLIENT_INFO = "clientInfo";
```
[nit] the other options start with lower case letters -- should we follow that pattern here as well?

mattChiaravalloti (approved):
Looks good! I added one last small request that you might want to consider before merging 👍

src/main/java/com/mongodb/jdbc/MongoDriver.java
bucaojit: Sounds good, I'll make the change.

bucaojit:
evergreen merge

PR #105, opened 2022-06-02
SQL-900: Reenable skipped JDBC integration tests

bucaojit:
Reenabled the skipped tests for tickets that have been fixed: SQL-659, SQL-668.

One question, MHOUSE-1459 is closed as won't fix.  The returned result set is empty.  I was debating whether to remove the skip reason and return the empty result set as that is the expected behavior now or keep skipped?

EDIT: Checked with Natacha about MHOUSE-1459 and will keep the tests skipped.

### DIFF
--- a/resources/integration_test/tests/dbmd_variable_result_sets_index_methods.yml
+++ b/resources/integration_test/tests/dbmd_variable_result_sets_index_methods.yml
@@ -4,12 +4,12 @@ tests:
   - description: getBestRowIdentifier no catalog filter returns results for all matching tables
     db: integration_test
     meta_function: ["getBestRowIdentifier", null, null, "foo", 0, false]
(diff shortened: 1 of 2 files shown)


nbagnard (approved):
LGTM!

mattChiaravalloti (changes requested):
For the ones that are marked as blocked on _both_ SQL-668 and MHOUSE-1459, can you remove the SQL-668 portion since they are now only blocked on the MHOUSE ticket?

bucaojit:
I removed the SQL-668 from the skip reasons.

mattChiaravalloti (approved):
ty! looks good 🎉

bucaojit:
evergreen merge

PR #106, opened 2022-06-02
SQL-573: Add Apache 2.0 license

mattChiaravalloti:
This PR adds the apache 2.0 license to the repo, along with copyright headers to all source code files. It looks like the release already includes the license. I updated `spotless` to handle this for us. Now, `spotlessCheck` will fail if a source file does not start with a license header, and `spotlessApply` will write the header for us whenever it is missing.

**Open Question**
Should the header be added to the java files in `demo` and `smoketest`? If so, should that also be done through `spotless`? Right now, those directories are _not_ inspected by `spotless`, so I think I'd have to update their `gradle.build` files and probably update the evergreen task to check those as well.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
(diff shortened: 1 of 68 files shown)


rychipman (approved):
LGTM! I'm ambivalent about the license headers in `demo` and `smoketest`, and I'm also not totally sure what the java ecosystem convention would be here. I'll defer to @nbagnard on that one!

nbagnard:
> I'm ambivalent about the license headers in `demo` and `smoketest`, and I'm also not totally sure what the java ecosystem convention would be here.

I think we can leave it. I don't know of a specific convention regarding test apps. In my experience, most of the time the header is just added de-facto to any new file in the project.

mattChiaravalloti:
evergreen merge

PR #108, opened 2022-06-07
SQL-916: ResultSet.close fails when calling Statement.close if Statement.closeOnClompetion is true

nbagnard:
A simple fix I had shelved for a long time now.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoResultSet.java
+++ b/src/main/java/com/mongodb/jdbc/MongoResultSet.java
@@ -159,7 +159,7 @@ public abstract class MongoResultSet<T> implements ResultSet {
         }
         cursor.close();
         closed = true;
-        if (statement != null && statement.isCloseOnCompletion()) {
+        if (statement != null && !statement.isClosed && statement.isCloseOnCompletion()) {
             statement.close();
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -159,7 +159,7 @@ public void close() throws SQLException {
         }
         cursor.close();
         closed = true;
-        if (statement != null && statement.isCloseOnCompletion()) {
+        if (statement != null && !statement.isClosed && statement.isCloseOnCompletion()) {
```
mattChiaravalloti, line 162: Question about `(MongoStatement).isCloseOnCompletion()`: do we want to keep the `checkClosed()` call at the start of that method? I'm not sure there is a reason to check if the connection is closed or not to return the `closeOnCompletion` flag for a statement. Maybe instead of adding this check here, we should instead update the `isCloseOnCompletion` method?

mattChiaravalloti (changes requested):
Agreed with Varsha re: can this easily be tested? If so, let's add or update a test!

nbagnard:
> Agreed with Varsha re: can this easily be tested? If so, let's add or update a test!

Very good point, I'll add a test 👍

src/main/java/com/mongodb/jdbc/MongoResultSet.java
mattChiaravalloti: Oh, thanks for pointing that out! That's too bad 🤷‍♀️

mattChiaravalloti (approved):

nbagnard:
evergreen merge

PR #110, opened 2022-06-15
SQL-170: Map BSON boolean to JDBC boolean instead of bit

EthanHardyMongo:
I modified BsonTypeInfo.java so that BSON_BOOL maps BSON boolean to JDBC Boolean. Additionally, I changed the integration tests that were expecting sql type BIT to instead expect sql type BOOLEAN.

### DIFF
--- a/resources/integration_test/tests/dbmd_constant_result_sets.yml
+++ b/resources/integration_test/tests/dbmd_constant_result_sets.yml
@@ -80,7 +80,7 @@ tests:
          array, false]
       - [false, false, null, -2, false, null, null, null, 0, 0, 1, 0, 0, 0, 0, 0,
          binData, false]
-      - [false, false, null, -7, false, null, null, null, 0, 0, 1, 0, 1, 3, 0, 0,
+      - [false, false, null, 16, false, null, null, null, 0, 0, 1, 0, 1, 3, 0, 0,
          bool, false]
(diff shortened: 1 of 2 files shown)


mattChiaravalloti:
Some nits about the PR title (which ultimately becomes the commit message):
- Let's try to start with present tense active verbs (changed `Mapped` => `Map`)
- No need to comment on test updates for something like this, that's always implied!

mattChiaravalloti (changes requested):
This looks good but it looks like there are failures on evergreen! We can look into what else needs to be changed together tomorrow 👍

bucaojit (changes requested):
Nice work getting your first PR up!  🎉 

There are a few integration tests still failing because the type for `BIT` which is -7 is still in a few of the tests that return the DATA_TYPE.  
If you find those and replace them with 16, which is the value for [BOOLEAN](https://github.com/JetBrains/jdk8u_jdk/blob/94318f9185757cc33d2b8d527d36be26ac6b7582/src/share/classes/java/sql/Types.java#L259) then the integration tests will pass.

mattChiaravalloti (approved):
Great work! When you get a second approval you can comment "evergreen merge" on this PR and that'll result in your first ticket being merged 🎊

bucaojit (approved):
LGTM!

Good job getting the integration tests working! 🔢

PR #115, opened 2022-07-14
SQL-813: Make extended json stringification strictness configurable in JDBC driver

kli19:


bucaojit (changes requested):
Good job on making the property change for the BsonValue!

There are additional parts needed to obtain the property from the properties file and it would start in MongoDriver.java.  Tracing through the changes needed to add the `clientinfo` [property](https://github.com/mongodb/mongo-jdbc-driver/blob/ac3c7edd2133fd64fe3da2ffa1672ec0b80b1d59/src/main/java/com/mongodb/jdbc/MongoDriver.java#L65) may help, [here](https://github.com/mongodb/mongo-jdbc-driver/blob/ac3c7edd2133fd64fe3da2ffa1672ec0b80b1d59/src/main/java/com/mongodb/jdbc/MongoDriver.java#L272) is where it gets the value.  We can add something like an `extjsonmode` property, and if property is not set default to the current behavior of `Relaxed`.

Once obtained, the value will need the passed down to where `MongoBsonValue` is used.  I see it is mainly in `MongoResultSet.java` where we can call the constructor you modified with the value from the `extjsonmode` property.

mattChiaravalloti:
Replacing myself with Varsha since I may not get to quickly enough!

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
```diff
@@ -1162,7 +1163,7 @@ void sameMetaDataOnDifferentRowsEvenWithDifferentBsonTypes() throws SQLException
                             return doc;
                         });
 
-        mockResultSet = new MongoResultSet(mongoStatement, cursor, sameMetadataSchema);
+        mockResultSet = new MongoResultSet(mongoStatement, cursor, sameMetadataSchema, false);
```
varsha804, line 1166: Should we add some tests for when `isExt` is `true`?

src/main/java/com/mongodb/jdbc/MongoBsonValue.java
```diff
@@ -34,14 +34,21 @@
...
 public class MongoBsonValue {
-    static final JsonWriterSettings JSON_WRITER_SETTINGS =
-            JsonWriterSettings.builder().outputMode(JsonMode.RELAXED).build();
+    private JsonWriterSettings JSON_WRITER_SETTINGS;
     static final EncoderContext ENCODER_CONTEXT = EncoderContext.builder().build();
 
     private BsonValue v;
 
-    public MongoBsonValue(BsonValue v) {
+    public MongoBsonValue(BsonValue v, boolean isExt) {
```
varsha804, line 42: [nit] Just for better readability
```suggestion
    public MongoBsonValue(BsonValue v, boolean isExtended) {
```

varsha804 (changes requested):
Looking good! 💯 Just one question about tests for now

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
kli19: I was thinking I should add a test case too but I was having a hard time figuring out how to go about that :(

Is there a particular test that I can reference as a template?

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -279,8 +280,12 @@ private MongoConnection createConnection(ConnectionString cs, Properties info)
                             + ". Expected format <name>+<version>.");
         }
 
+        boolean extJsonMode =
+                Boolean.parseBoolean(info.getProperty(EXT_JSON_MODE.getPropertyName()));
```
bucaojit, line 284: Nice, you were able to get the property set for `extjsonmode`.  

A boolean for the `value` may be a big confusing as both RELAXED and EXTENDED are technically still resulting in extended json.  Let's have the property take the values `RELAXED` or `EXTENDED` and obtain it as a string.  A similar example is with the `loglevel` or `logdir` values.  
ie
```
// Earlier in the file declare the RELAXED string, or an enum for RELAXED and EXTENDED
static final String RELAXED = "RELAXED";

// Get the value here, RELAXED as default
String extJsonModeVal = info.getProperty(EXT_JSON_MODE.getPropertyName(), RELAXED);

// set `extJsonMode` to true/false based on the string value otherwise throw exception if invalid value entered
```

bucaojit (changes requested):
Changes look good with the property changes.  Great work finding all the areas that need the extra parameter.

One comment regarding the property type.

Also, please add a test in MongoDriverTest.java similar to `testClientInfoProperty()` that checks that the valid input values work, and that an invalid value returns an exception.

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
kli19: I added a test in MongoDriverTest.java based on Oliver's comment :)

src/main/java/com/mongodb/jdbc/MongoDriver.java
kli19: The `testClientInfoProperty()` example was helpful in writing a test :)

bucaojit (changes requested):
Adding a `Request change` to add the MongoResultSet test change.

varsha804 (approved):
lgtm!

bucaojit (approved):
Thanks for adding the unit tests.

LGTM! 💯

kli19:
evergreen merge

PR #121, opened 2022-11-02
SQL-645: refactor MongoResultSet to use BsonTypeInfo

terakilobyte:
This refactors `MongoResultSet` to use `BsonTypeInfo` and significantly cleans up the switch statements.

I've also removed the "end_of_document" check from `MongoResultSet` as it wasn't used in any meaningful way other than it couldn't be converted in any of the switches. Since it was a BsonValue that is being passed into the various conversion methods, there's actually no way for it to be an "end_of_document" since it can't be created as a concrete type to derive a `BsonType` from (let alone a `BsonTypeInfo`).

This also updates the go version used in the build task, addressing SQL-1083.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
+++ b/src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
@@ -22,6 +22,7 @@ import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Set;
 import org.bson.BsonType;
+import org.bson.BsonValue;
 
 public enum BsonTypeInfo {
     BSON_DOUBLE("double", BsonType.DOUBLE, Types.DOUBLE, false, 15, 15, 2, 15, 15, 8),
@@ -56,16 +57,18 @@ public enum BsonTypeInfo {
     BSON_DECIMAL("decimal", BsonType.DECIMAL128, Types.DECIMAL, false, 34, 34, 10, 34, 34, 16),
     BSON_MINKEY("minKey", BsonType.MIN_KEY, Types.OTHER, false, 0, 0, 0, null, null, null),
     BSON_MAXKEY("maxKey", BsonType.MAX_KEY, Types.OTHER, false, 0, 0, 0, null, null, null),
-    BSON_BSON("bson", BsonType.UNDEFINED, Types.OTHER, false, 0, 0, 0, null, null, null);
+    BSON_BSON("bson", BsonType.UNDEFINED, Types.OTHER, false, 0, 0, 0, null, null, null),
+    BSON_DOCUMENT("document", BsonType.DOCUMENT, Types.OTHER, false, 0, 0, 0, null, null, null);
 
     // BSON_TYPE_NAMES is the set of all valid BSON type names as listed
-    // here: https://docs.mongodb.com/manual/reference/bson-types/#bson-types
+    // here: https://mongodb.github.io/mongo-java-driver/3.12/javadoc/org/bson/BsonType.html
     // Note that BsonTypeInfo contains a BSON_BSON variant which is intentionally
     // omitted from this set. That is because "bson" is our catch-all for a
     // value that has "any" type, but is not an actual specific BSON type.
     private static final Set<String> BSON_TYPE_NAMES =
             new HashSet<>(
                     Arrays.asList(
+                            BSON_DOCUMENT.bsonName,
                             BSON_DOUBLE.bsonName,
                             BSON_STRING.bsonName,
                             BSON_OBJECT.bsonName,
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -56,16 +57,18 @@ public enum BsonTypeInfo {
     BSON_DECIMAL("decimal", BsonType.DECIMAL128, Types.DECIMAL, false, 34, 34, 10, 34, 34, 16),
     BSON_MINKEY("minKey", BsonType.MIN_KEY, Types.OTHER, false, 0, 0, 0, null, null, null),
     BSON_MAXKEY("maxKey", BsonType.MAX_KEY, Types.OTHER, false, 0, 0, 0, null, null, null),
-    BSON_BSON("bson", BsonType.UNDEFINED, Types.OTHER, false, 0, 0, 0, null, null, null);
+    BSON_BSON("bson", BsonType.UNDEFINED, Types.OTHER, false, 0, 0, 0, null, null, null),
+    BSON_DOCUMENT("document", BsonType.DOCUMENT, Types.OTHER, false, 0, 0, 0, null, null, null);
```
mattChiaravalloti, line 61: We already have the `BSON_OBJECT` variant which represents the same thing.

The problem is we have two words for the same concept 😞 The [docs](https://www.mongodb.com/docs/manual/reference/bson-types/) list "object" as the BSON type name, whereas the Java (and, I think, all other) driver(s) use the term "document". Given the extensive use of `BSON_OBJECT` throughout other files in the JDBC driver, I'd say we should defer to using that one instead of this new one. However, it is also an option to update the name of `BSON_OBJECT` to `BSON_DOCUMENT` if we prefer the term "document" to "object".

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
```diff
@@ -228,6 +291,8 @@ public static BsonTypeInfo getBsonTypeInfoByName(String typeName) throws SQLExce
                         return BSON_DOUBLE;
                     case 7:
                         return BSON_DECIMAL;
+                    case 8:
+                        return BSON_DOCUMENT;
```
mattChiaravalloti, line 295:
```suggestion
```
Since the mongodb [docs](https://www.mongodb.com/docs/manual/reference/bson-types/) and ADL use the name "object" instead of "document", I'm not sure this change is correct. Or at least, it may be dead code since we won't ever get a schema from ADL that uses the word "document" to describe this type.

src/test/java/com/mongodb/jdbc/BsonTypeInfoTest.java
```diff
@@ -20,7 +20,11 @@
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+import java.nio.charset.StandardCharsets;
 import java.sql.SQLException;
+import org.bson.*;
```
mattChiaravalloti, line 25: In the past, I know we've tried to avoid wildcard imports. I'm not sure if this is something asserted by spotless, or if it something we want to continue to do. I believe no other files have wildcard imports but we can check that. If we do have them in other places, then there is nothing to do here. If we do not use them or decide we do not want them, can you replace this with the actual imports needed?

You can also configure your IDE/editor to not use wildcard imports when there are multiple classes imported from the same package!

mattChiaravalloti (changes requested):
Good start, and I'm glad to see we're getting the drive-by go 1.18 fix in 🚗 

Just a few comments on this. Looks like you picked up java again just fine!

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
terakilobyte: I had questions about this when I was making the addition. I think for our uses, having a concrete `BsonDocument` converted to a `BsonTypeInfo.BSON_OBJECT` would be okay. Thoughts?

src/main/java/com/mongodb/jdbc/MongoResultSet.java
```diff
@@ -1422,7 +1261,7 @@ public Blob getBlob(int columnIndex) throws SQLException {
     }
 
     protected Clob getClob(BsonValue o) throws SQLException {
-        return new SerialClob(getString(o).toCharArray());
+        return new SerialClob(Objects.requireNonNull(getString(o)).toCharArray());
```
nbagnard, line 1264: Objects.requireNonNull will throw if the o is null. This means that when the data is NULL then the driver will throw an exception when getClob is called.
I think that it should return Null instead.
Add the null check before converting the data and return Null if necessary
```suggestion
        if (checkNull(o)) {
            return null;
        }
        return new SerialClob(getString(o).toCharArray());
```

nbagnard (changes requested):
Pretty good! I makes things a little more compact and easier to read :)
I only have comment regarding the behaviour of getClob. It looks like we had a bug there even before and we should return Null.

src/main/java/com/mongodb/jdbc/MongoResultSet.java
terakilobyte: Great catch!

nbagnard (approved):
LGTM 👍 🎉

src/main/java/com/mongodb/jdbc/BsonTypeInfo.java
mattChiaravalloti: Agreed, it will be okay! It's the same type and the `BsonTypeInfo` we've already used in all other classes 👍

mattChiaravalloti (approved):
lgtm 👍

PR #122, opened 2022-12-14
SQL-1132: adds support for default db in url

terakilobyte:
This commit adds support for specifying the database in the url. When specifying a default database, the `authSource` option *must* be specified unless the default database is also the authentication database (not the case for ADF). Because of the changes I removed a test and changed the URL in another test since they became valid.

Documentation changes will be required, and I think it might make sense to ask Atlas to update the connection modal for jdbc to show the default database in the URL, i.e. `jdbc:mongodb://your-server/default?ssl=true&authSource=admin` and perhaps remove the database field? cc @alexiantonino on this idea.

I did also check with @alexiantonino and we won't be making any changes to the tableau connection interface. In that instance, the database specified will always override the URI database if someone were to include it there, but it requires fewer changes outside of us.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -370,7 +370,6 @@ public class MongoDriver implements Driver {
             throw new SQLException(e);
         }
 
-
         ParseResult result = normalizeConnectionOptions(originalConnectionString, info);
         String user = result.user;
         char[] password = result.password;
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -380,7 +380,9 @@ public class MongoDriver implements Driver {
 
         // A database to connect to is required. If they have not specified one, look in the connection string for a
         // database. The specified database in the connect window will always override the uri database.
-        if ((!info.containsKey(DATABASE.getPropertyName()) || info.getProperty(DATABASE.getPropertyName()).isEmpty()) && originalConnectionString.getDatabase() != null) {
+        if ((!info.containsKey(DATABASE.getPropertyName())
+                        || info.getProperty(DATABASE.getPropertyName()).isEmpty())
+                && originalConnectionString.getDatabase() != null) {
             info.setProperty(DATABASE.getPropertyName(), originalConnectionString.getDatabase());
         }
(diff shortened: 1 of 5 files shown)


src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -93,18 +93,12 @@ void testBasicURL() throws SQLException {
     }
 
     @Test
-    void testDBURL() throws SQLException {
+    void testURIDBURL() throws SQLException {
```
nbagnard, line 96: Can you add a test which shows what happens when their is a default DB supplied in both the URI and the properties.
- For `jdbc:mongodb://localhost/foo?authSource=admin` + `database=myPropsDB`, the result should be default DB = myPropsDB, authSource = admin.
- For `jdbc:mongodb://localhost/admin` + `database=foo`, the result should be default DB = foo, authSource = admin.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -377,10 +377,16 @@ private Pair<ConnectionString, DriverPropertyInfo[]> getConnectionSettings(
         char[] password = result.password;
 
         List<DriverPropertyInfo> mandatoryConnectionProperties = new ArrayList<>();
+
+        // A database to connect to is required. If they have not specified one, look in the connection string for a
```
nbagnard, line 381: The only part which worries me a bit is this: `String authDatabase = originalConnectionString.getDatabase();`
I believe this means that we never use the actual authSource setting a user provides.
This will depend on what the code at the bottom of this function does
```
                new ConnectionString(
                        buildNewURI(
                                originalConnectionString.getHosts(),
                                user,
                                password,
                                authDatabase,
                                result.normalizedOptions));
```

nbagnard (changes requested):
There are a couple of things that I am not 100% sure we are handling correctly.
Let's revisit this when you are back.
Until then happy Holidays and safe travel :)

terakilobyte:
@nbagnard I pushed some updates and a new connection string test file for us to discuss that I believe captures the exact behavior we want.

pmeredit (approved):
LGTM! The results agree with what I did in the ODBC driver.

nbagnard (approved):
LGTM!

terakilobyte:
evergreen merge

PR #126, opened 2023-02-03
SQL-1055: Update the list of supported functions

mattChiaravalloti:
This PR updates the list of supported functions in the JDBC driver. The function list now includes functions that were added to MongoSQL for Tableau support. Only functions that are directly defined in the dialect -- not those that are defined via Tableau formulae -- are included, as per the ticket comment.

This PR also updates references to "ADL" to "ADF", particularly the environment variable names. This makes this repo consistent with the odbc repo. Evergreen variables have been modified accordingly.


.evg.yml
```diff
@@ -175,15 +175,14 @@ functions:
           MDBJDBC_VER: "$MDBJDBC_VER"
           JAVA_HOME: "$JAVA_HOME"
           PREPARE_SHELL: |
-            export ADL_TEST_LOCAL_USER=${adl_test_local_user}
-            export ADL_TEST_LOCAL_PWD=${adl_test_local_pwd}
-            export ADL_TEST_LOCAL_HOST=${adl_test_local_host}
```
mattChiaravalloti, line 180: This variable is not used anywhere in the repo and the evergreen variable was never defined, so I removed it here.

mattChiaravalloti:
Looks like there may be some issues with the new variable names. Will investigate.

Also, there are unit tests I need to update!

mattChiaravalloti:
Ah, I know the issue -- I forgot you need to press "save" all the way at the bottom of the evergreen page! will update and re-run a patch to verify the changes 👍

nbagnard (commented):
I am not sure that we want to list all the different permutations with the different data types for the parameters. I couldn't remember of a driver which does that. I think they usually return one row with the type having the most coverage.
I know we are more precise than most DBs/dialect and don't support auto casting, so this list makes sense in this regard.
I am just not sure it adds that much value for JDBC users or if it does not bloat results too much.
WDYT?

README.md
```diff
@@ -120,17 +120,17 @@ Integration testing requires a local MongoDB and Atlas Data Lake instance to be
 ```
 
 ### Running Local mongod and Atlas Data Lake
```
bucaojit, line 122:
```suggestion
### Running Local mongod and Atlas Data Federation
```

README.md
```diff
@@ -120,17 +120,17 @@ Integration testing requires a local MongoDB and Atlas Data Lake instance to be
...
-`run_adl.sh` is a helper script that will start a local mongod and Atlas Data Lake instance, used for integration testing.
+`run_adf.sh` is a helper script that will start a local mongod and Atlas Data Lake instance, used for integration testing.
```
bucaojit, line 123:
```suggestion
`run_adf.sh` is a helper script that will start a local mongod and Atlas Data Federation instance, used for integration testing.
```

resources/run_adf.sh
```diff
@@ -1,18 +1,18 @@
 #!/bin/bash
 # 
-# Usage: run_adl.sh <operation>
+# Usage: run_adf.sh <operation>
 # operation: 'start' or 'stop'
 #
 # This script will start a local mongod and Atlas Data Lake instance, used for integration testing.
```
bucaojit, line 6:
```suggestion
# This script will start a local mongod and Atlas Data Federation instance, used for integration testing.
```

resources/integration_test/tests/dbmd_variable_result_sets_function_methods.yml
```diff
@@ -30,42 +30,258 @@ tests:
...
+        removed., 1, TRIM]
+      - [def, null, TRIM, returns the string str with all remstr prefixes and/or suffixes
+        removed., 1, TRIM]
+      - [def, null, UPPER, returns the provided string with all characters changed
+        to uppercase., 1, UPPER]
+    row_count: 114
 
   - description: getFunctions_partial_filter_with_wildcard_substrings_returns_matching_functions
     db: integration_test
     meta_function: [ getFunctions, integration_test, null, '%S%' ]
```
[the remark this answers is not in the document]
bucaojit: It seems fine to me to have the full test set to be thorough, as long as it doesn't cause the running of the tests to take too long.

bucaojit (commented):
I took a brief look and saw a few places where we can update `Atlas Data Lake` to `Atlas Data Federation`.  Changes look great and I will continue with reviewing on Monday.

bucaojit:
There is another adl reference in [gradle.properties](https://github.com/mongodb/mongo-jdbc-driver/blob/655fdf34c83ad00348bfdd7866599c9cdb0cf35e/gradle.properties#L2) for the artifactId.

src/main/java/com/mongodb/jdbc/MongoFunctions.java
```diff
@@ -79,17 +93,85 @@ public static MongoFunctions getInstance() {
...
+                                        BsonTypeInfo.BSON_DECIMAL.getBsonName(),
+                                        "returns the absolute value of the given number.",
+                                        new String[] {BsonTypeInfo.BSON_DECIMAL.getBsonName()},
+                                        FunctionCategory.NUM_FUNC),
                                 new MongoFunction(
                                         "BIT_LENGTH",
                                         BsonTypeInfo.BSON_LONG.getBsonName(),
                                         "returns length of string in bits",
                                         new String[] {BsonTypeInfo.BSON_STRING.getBsonName()}),
+                                new MongoFunction(
```
bucaojit, line 125: This looks to be missing a `FunctionCategory.NUM_FUNC` as a last parameter.  Same with the `LOG` and `POW` categories.

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -383,15 +383,25 @@ void testGetSQLKeywords() throws SQLException {
 
     @Test
     void testGetNumericFunctions() throws SQLException {
-        final String expectedFunctions = "";
+        final String expectedFunctions =
+                "ABS,ABS,ABS,ABS,"
```
bucaojit, line 387: A few of the newly added functions will be added to this list after updating the function category in `MongoFunctions.java`

bucaojit (changes requested):

bucaojit:
@nbagnard I find it helpful to have all the function permutations, unless it is documented elsewhere.  Although if other driver's don't normally handle things that way it does make sense to reduce it to one entry or fewer for example, MOD and LOG instead of 16 entries could show how they can be used with INTEGER and DOUBLE to reduce it to 4 entries?

resources/integration_test/tests/dbmd_variable_result_sets_function_methods.yml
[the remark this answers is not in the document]
mattChiaravalloti: Cool, yea these tests ran locally in essentially no time at all so I'll stick with it!

README.md
mattChiaravalloti: Good catch -- I was so focused on the acronym, I forgot to look for the actual words 😅 I'll make this change and do an audit myself to make sure I didn't miss any others

mattChiaravalloti:
> There is another adl reference in [gradle.properties](https://github.com/mongodb/mongo-jdbc-driver/blob/655fdf34c83ad00348bfdd7866599c9cdb0cf35e/gradle.properties#L2) for the artifactId.

That one I wasn't sure if I could/should change since that will start to change the name of the released artifacts, right? Is it safe/okay to make that change?

src/main/java/com/mongodb/jdbc/MongoFunctions.java
mattChiaravalloti: The JDBC spec has a "CEILING" function but not a "CEIL" function. From previous PRs and conversations, I remember us deciding not to mark anything with JDBC labels (such as "numeric", "system", etc) unless they were exact matches with the spec. Even though our "CEIL" behaves the same as their "CEILING", the name is different so I omitted the label intentionally.

mattChiaravalloti:
> @nbagnard I find it helpful to have all the function permutations, unless it is documented elsewhere. Although if other driver's don't normally handle things that way it does make sense to reduce it to one entry or fewer for example, MOD and LOG instead of 16 entries could show how they can be used with INTEGER and DOUBLE to reduce it to 4 entries?

For both of you, which types would we choose for the unary functions? Should we do DECIMAL since it is the "highest" numeric type in our hierarchy? Or should we do DOUBLE since that is a more common type?

src/main/java/com/mongodb/jdbc/MongoFunctions.java
bucaojit: Ah that makes sense!  Thanks for clarifying.

bucaojit:
> There is another adl reference in [gradle.properties](https://github.com/mongodb/mongo-jdbc-driver/blob/655fdf34c83ad00348bfdd7866599c9cdb0cf35e/gradle.properties#L2) for the artifactId.

>>That one I wasn't sure if I could/should change since that will start to change the name of the released artifacts, right? Is it safe/okay to make that change?

So far I have only seen `mongodb-jdbc` for the artifact ID such as [here](https://search.maven.org/artifact/org.mongodb/mongodb-jdbc).  Let's keep it as is for now until we have a better idea of where it is used.

nbagnard:
Thank you both for your comments.
It looks like even though the list is pretty long for MOD for example, it is representing the functions we support realistically and trying to find out which type we should use for 'number' gets us into the tricky discussion of DECIMAL vs DOUBLE.
Let's keep it this way then :)

nbagnard:
> So far I have only seen `mongodb-jdbc` for the artifact ID such as [here](https://search.maven.org/artifact/org.mongodb/mongodb-jdbc). Let's keep it as is for now until we have a better idea of where it is used.

It's probably not used.
Gradle uses the project name as the artifact id. The default project name is the project’s directory name, or it can be defined in settings.gradle (this is where you'll see `rootProject.name = 'mongodb-jdbc'`).
And we don't override the artifactid in publish.gradle.

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
```diff
@@ -383,15 +383,25 @@ void testGetSQLKeywords() throws SQLException {
 
     @Test
     void testGetNumericFunctions() throws SQLException {
-        final String expectedFunctions = "";
+        final String expectedFunctions =
```
nbagnard, line 386: You don't need to have the function names repeated here. It is a little different than the ones from getFunctions which are giving details about the arguments, etc... Here, if a function name is in the list, it means that it support the Open Group CLI specification for this function.

nbagnard (changes requested):
This is looking good! That was a pile of functions and permutation to go through!
I only have 2 very small comments :)

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
mattChiaravalloti: When I updated this to just have the names once, the unit test fails with
```
expected: <ABS,COS,DEGREES,FLOOR,MOD,RADIANS,ROUND,SIN,SQRT,TAN> but was: <ABS,ABS,ABS,ABS,COS,COS,COS,COS,DEGREES,DEGREES,DEGREES,DEGREES,FLOOR,FLOOR,FLOOR,FLOOR,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,RADIANS,RADIANS,RADIANS,RADIANS,ROUND,ROUND,ROUND,ROUND,ROUND,ROUND,ROUND,ROUND,SIN,SIN,SIN,SIN,SQRT,SQRT,SQRT,SQRT,TAN,TAN,TAN,TAN>
Expected :ABS,COS,DEGREES,FLOOR,MOD,RADIANS,ROUND,SIN,SQRT,TAN
Actual   :ABS,ABS,ABS,ABS,COS,COS,COS,COS,DEGREES,DEGREES,DEGREES,DEGREES,FLOOR,FLOOR,FLOOR,FLOOR,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,MOD,RADIANS,RADIANS,RADIANS,RADIANS,ROUND,ROUND,ROUND,ROUND,ROUND,ROUND,ROUND,ROUND,SIN,SIN,SIN,SIN,SQ ...
```

I suspect this is because we collect the function names [here](https://github.com/mongodb/mongo-jdbc-driver/blob/655fdf34c83ad00348bfdd7866599c9cdb0cf35e/src/main/java/com/mongodb/jdbc/MongoFunctions.java#L196) by grabbing each function in a given category. We just return that list (with duplicates) directly [here](https://github.com/mongodb/mongo-jdbc-driver/blob/655fdf34c83ad00348bfdd7866599c9cdb0cf35e/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java#L324), then in the test we use a simple `assertEquals` to compare.

We could avoid the duplicates by not marking each variant in `MongoFunctions.java` with the category -- maybe we can just mark the first? Or we can remove duplicates before stringifying the lists? The former idea could be a good option, but I suppose the latter is the safer choice since that means we won't lose information if we edit/delete some/all of the duplicates. What do you think?

README.md
mattChiaravalloti: updated

README.md
mattChiaravalloti: updated

resources/run_adf.sh
mattChiaravalloti: updated

src/test/java/com/mongodb/jdbc/MongoDatabaseMetaDataTest.java
nbagnard: As discussed offline, removing duplicates before turning the list into a string is better/safer.

mattChiaravalloti: resolved offline: went with option 2. We no longer add duplicates to the lists, so now the test has just one instance of each 👍

nbagnard (approved):
LGTM! 🎉

bucaojit (approved):
LGTM! 💯

mattChiaravalloti:
evergreen merge

PR #129, opened 2023-04-20
SQL-1328: Report authorized Collections for getTables and getColumns

bucaojit:
Tested on remote ADF by adding collection actions on specific collections and running `tryGetTables()` in the TestUtils class. 
Code without the change gets `Command failed with error 13 (Unauthorized): 'not authorized,`.  With this change the tables are listed.  I considered adding an integration test to verify this, I didn't find a simple way to set the configuration for ADF but it seems possible.

In testing also found that databases that user is not authorized to use returns as an empty string "", added some code to filter that out.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
@@ -406,19 +406,38 @@ public class MongoDatabaseMetaData implements DatabaseMetaData {
 
     // Helper for getting a stream of all database names.
     private Stream<String> getDatabaseNames() {
-        return this.conn.mongoClient.listDatabaseNames().into(new ArrayList<>()).stream();
+        return this.conn
+                .mongoClient
+                .listDatabaseNames()
+                .into(new ArrayList<>())
+                .stream()
+                .filter(dbName -> !dbName.isEmpty());
+    }
+
+    // Helper for getting a list of collection names from the db
+    // Using runCommand instead of listCollections as listCollections does not support authorizedCollections option
+    private ArrayList<Document> getCollectionsFromRunCommand(MongoDatabase db) {
+        Document listCollectionsResult =
+                db.runCommand(
+                        new Document("listCollections", 1)
+                                .append("authorizedCollections", true)
+                                .append("nameOnly", true));
+        ArrayList<Document> firstBatch =
+                (ArrayList<Document>)
+                        listCollectionsResult
+                                .get("cursor", Document.class)
+                                .getList("firstBatch", Document.class);
+
+        return firstBatch;
     }
 
     // Helper for getting a stream of MongoListCollectionsResults from the argued db that match
     // the argued filter.
     private Stream<MongoListTablesResult> getTableDataFromDB(
             String dbName, Function<MongoListTablesResult, Boolean> filter) {
-        return this.conn
-                .getDatabase(dbName)
-                .withCodecRegistry(MongoDriver.registry)
-                .listCollections(MongoListTablesResult.class)
-                .into(new ArrayList<>())
+        return getCollectionsFromRunCommand(this.conn.getDatabase(dbName))
                 .stream()
+                .map(MongoListTablesResult::new)
                 .filter(filter::apply);
     }
 
@@ -1522,10 +1541,11 @@ public class MongoDatabaseMetaData implements DatabaseMetaData {
             Pattern tableNamePatternRE,
             Pattern columnNamePatternRE,
             Function<GetColumnsDocInfo, BsonDocument> bsonSerializer) {
-        MongoDatabase db = this.conn.getDatabase(dbName).withCodecRegistry(MongoDriver.registry);
-
-        return db.listCollectionNames()
-                .into(new ArrayList<>())
+        MongoDatabase db = this.conn.getDatabase(dbName);
+        return getCollectionsFromRunCommand(db)
+                .stream()
+                .map(collection -> collection.get("name", String.class))
+                .collect(Collectors.toList())
                 .stream()
 
                 // filter only for collections matching the pattern
@@ -1539,10 +1559,12 @@ public class MongoDatabaseMetaData implements DatabaseMetaData {
                         tableName ->
                                 new Pair<>(
                                         new Pair<>(dbName, tableName),
-                                        db.runCommand(
-                                                new BsonDocument(
-                                                        "sqlGetSchema", new BsonString(tableName)),
-                                                MongoJsonSchemaResult.class)))
+                                        db.withCodecRegistry(MongoDriver.registry)
+                                                .runCommand(
+                                                        new BsonDocument(
+                                                                "sqlGetSchema",
+                                                                new BsonString(tableName)),
+                                                        MongoJsonSchemaResult.class)))
 
                 // filter only for collections that have schemas
                 .filter(p -> isValidSchema(p.right()))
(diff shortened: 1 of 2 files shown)


terakilobyte (approved):
LGTM. It appears evergreen is having some issues with Gradle?

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -406,19 +406,38 @@ public ResultSet getTableTypes() throws SQLException {
...
+                .stream()
+                .filter(dbName -> !dbName.isEmpty());
+    }
+
+    // Helper for getting a list of collection names from the db
+    // Using runCommand instead of listCollections as listCollections does not support authorizedCollections option
+    private ArrayList<Document> getCollectionsFromRunCommand(MongoDatabase db) {
+        Document listCollectionsResult =
+                db.runCommand(
+                        new Document("listCollections", 1)
```
nbagnard, line 422: Since we are now using a listCollections command, can we create a JIRA to add server side filtering using the `filter` field?
For both the JDBC and ODBC driver, please.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -1539,10 +1559,12 @@ private Stream<BsonDocument> getColumnsFromDB(
                         tableName ->
                                 new Pair<>(
                                         new Pair<>(dbName, tableName),
-                                        db.runCommand(
-                                                new BsonDocument(
-                                                        "sqlGetSchema", new BsonString(tableName)),
-                                                MongoJsonSchemaResult.class)))
+                                        db.withCodecRegistry(MongoDriver.registry)
```
nbagnard, line 1562: Was this not working when applied at the beginning of the function?
I don't know the impact of setting the codec registry, but it seems that if we can avoid doing it for each collection, that'd be good. Worst case, it won't make a difference, best case it does not slow the stream processing.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -406,19 +406,38 @@ public ResultSet getTableTypes() throws SQLException {
...
+                .into(new ArrayList<>())
+                .stream()
+                .filter(dbName -> !dbName.isEmpty());
+    }
+
+    // Helper for getting a list of collection names from the db
+    // Using runCommand instead of listCollections as listCollections does not support authorizedCollections option
+    private ArrayList<Document> getCollectionsFromRunCommand(MongoDatabase db) {
+        Document listCollectionsResult =
+                db.runCommand(
```
nbagnard, line 421: I am also wondering if we could use this runCommand : <TResult> TResult runCommand(Bson command, Class<TResult> resultClass) with TResult being MongoIterable<MongoListTablesResult> or ListCollectionsIterable<MongoListTablesResult>.
This way you don't need the new constructor for MongoListTablesResult and we don't have a special deserialization process for this and it looks like we could have a proper batch cursor and even use iterators out of the box.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -406,19 +406,38 @@ public ResultSet getTableTypes() throws SQLException {
...
+                        new Document("listCollections", 1)
+                                .append("authorizedCollections", true)
+                                .append("nameOnly", true));
+        ArrayList<Document> firstBatch =
+                (ArrayList<Document>)
+                        listCollectionsResult
+                                .get("cursor", Document.class)
+                                .getList("firstBatch", Document.class);
```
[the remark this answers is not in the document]
nbagnard: It looks like the Java driver has a mechanism for getting a cursor on runCommand. While I think Nathan you asked and it seems the results should always be in 1 doc, I feel that if we can have cursor support, I'd feel much better. Especially when I read the doc for the listCollections command : https://www.mongodb.com/docs/manual/reference/command/listCollections/#mongodb-data-listCollections.cursor

nbagnard (changes requested):
The changes are looking OK, but I'd love it if we could have cursors support to make sure that we'll never "loose" collections.
If you can give it a look and see if you can get it to work, inspiring yourself from what the Java driver does for ListCollections, that would be awesome.
If it actually ends up looking super tricky, let's create a follow up JIRA.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: Sure, created JIRAs for JDBC and ODBC drivers.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: The runCommand() was sending an empty document with the codec registry set, I was able to resolve the issue by adding `MongoClientSettings.getDefaultCodecRegistry()` to our codec registry.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: I added the `MongoRunCmdListTablesResult ` class for this.  I looked into `ListCollectionsIterable` and `MongoIterable` but could not find way to make them work.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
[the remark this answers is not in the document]
bucaojit: I tried getting the cursor mechanism to work, I don't think it will for listCollections since the cursor needs the caller to specify a collection name for the `getMore` command and listCollections isn't specific to a single collection.  I have an attempt at [this branch](https://github.com/bucaojit/mongo-jdbc-driver/tree/getMoreCursorAttempt), then I realized I was not setting the right collection name.

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
nbagnard: Yep, I looked into it a bit too.
I'm pretty sure there is a way, but it does not seem trivial because ListCollectionsOperation does not have the ability to let us set authorizedCollections. We would have to extends a few classes to get this to work.
I am wondering why this is not a problem for Java driver users.
However, I am still concern that we will discard Collections if there is more than 1 batch of data.
Can you ask the Java driver team or check other tools (mongosh, other drivers) to see how they deal with this?

terakilobyte: We don't have to worry about batches. The command response is a bit quirky, but unless Server changes something, all collections will be returned under `firstBatch`. [Here is the relevant code](https://github.com/mongodb/mongo/blob/master/src/mongo/db/commands/list_collections.cpp#L255-L261).

nbagnard (approved):
LGTM!
Nathan gave me more context on the listCollection command and the "useless" cursor. All good to go :)

src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
bucaojit: Thanks Nathan for helping clarify that and providing the relevant code. 👍

bucaojit:
evergreen merge

PR #133, opened 2023-06-02
SQL-1348: bump java driver version to 4.9.1

jchemburkar:


nbagnard (approved):
LGTM!
You are right, I don't think we were looking at any specific bug fix from the Java driver.

jchemburkar:
evergreen merge

PR #136, opened 2023-07-28
fix formatting

terakilobyte:
There was an errant space in our evergreen file that prevented the download-center-update task from being seen and running


nbagnard (approved):
LGTM!
Thanks for figuring this one out

terakilobyte:
evergreen merge

PR #241, opened 2023-11-09
SQL-1707: put result set in select list order

jchemburkar:
To test, ran the SQL-1773 version of mongosql locally in ADF. tried each of unit tests and integration tests. Failing tests are the two result set tests expecting alphabetical order but getting select order


bucaojit (approved):
LGTM

pmeredit (approved):
LGTM!

PR #244, opened 2023-12-23
SQL-1835:  The driver disables all logger when connecting

nbagnard:
MongoDriver used to reset the log manager configuration with the configuration of the logging.properties from inside our jar. This is problematic for other applications using JUL as their configuration is deleted.
Instead, we now do the configuration for our Filehandler programmatically.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -25,6 +25,7 @@ import com.mongodb.client.MongoDatabase;
 import com.mongodb.jdbc.logging.AutoLoggable;
 import com.mongodb.jdbc.logging.DisableAutoLogging;
 import com.mongodb.jdbc.logging.MongoLogger;
+import com.mongodb.jdbc.logging.MongoSimpleFormatter;
 import java.io.File;
 import java.io.IOException;
 import java.sql.Array;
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -633,9 +647,13 @@ private void initConnectionLogger(Integer connection_id, Level logLevel, File lo
                         synchronized (this) {
                             if (!fileHandlers.containsKey(logDirPath)) {
                                 String logPath = logDirPath + File.separator + "connection.log";
-                                FileHandler fileHandler = new FileHandler(logPath);
+                                // Create a new file handler with the configuration provided instead of relying on
+                                // properties. This way, our handler configuration is not affected by other application
+                                // using JUL
+                                FileHandler fileHandler =
+                                        new FileHandler(logPath, 10000000, 1, true);
```
bucaojit, line 654: In the `logging.properties` that was are moving away from the Filehandler.count is `5`, is the change to `1` here intentional?

Another question here is, do we want to eventually make some of these values configurable such as the `limit` and `count` values?  If so, we can have a JIRA to address it if we later decide to implement it.

bucaojit (commented):
Code looks good, great idea with using the connection hashcode. 

I had one comment.

jchemburkar (approved):
changes look good to me!

src/main/java/com/mongodb/jdbc/MongoConnection.java
nbagnard: Yes, this is to keep the same behavior as before.
If you look at the previous code, you'll see that we were using the FileHandler(String pattern) constructor and it ignores the properties and sets limit to 0 and count to 1.
```
public FileHandler(String pattern) throws IOException, SecurityException {
        if (pattern.length() < 1 ) {
            throw new IllegalArgumentException();
        }
        checkPermission();
        configure();
        this.pattern = pattern;
        this.limit = 0;
        this.count = 1;
        openFiles();
    }
```
Using 5 for count actually creates a new files for each "batch" of 5 connections.

bucaojit (approved):
LGTM, thanks for the explanation!

PR #246, opened 2024-02-08
SQL-1881: make functions silent that use sensitive vars

terakilobyte:
I missed variables due to checking the wrong log. Additionally, even though the local testing variables aren't really sensitive I've made that function silent as well for good measure.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -111,7 +111,7 @@ functions:
       working_dir: mongo-jdbc-driver
       script: |
           ${PREPARE_SHELL}
-          ./gradlew clean test 
+          ./gradlew clean test
 
   "run integration test":
     command: shell.exec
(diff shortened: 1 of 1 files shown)


.evg.yml
```diff
@@ -145,6 +145,7 @@ functions:
   "export variables":
     - command: shell.exec
       params:
+        silent: true
```
nbagnard, line 148: This is great, but the lines below where we try to print the env are useless now. Maybe we can remove them or comment them saying that since the task is silent it won't be visible?
I am a little unsure why we had to dump it, but it must have been for a good reason.

nbagnard (approved):
LGTM!
I just have one very small comment, but it is not blocking.

.evg.yml
terakilobyte: @nbagnard I've removed the env dump lines.
@pmeredit It looks like you added the env dump lines. Were they needed? If so, I will break this up into two functions.

pmeredit: They were used in debugging. It's fine to not dump them now.

pmeredit (approved):
These are the only functions flagged in the spreadsheet, LGTM!

PR #247, opened 2024-02-21
SQL-1916: Fail retrieving info by column name if the column name is duplicated

nbagnard:
In order to not break the existing behavior (and to match MySQL and Postgres), I did not rename the columns. 
Instead, the driver will fail retrieving information by column name if there is more than one column with the requested name.

### DIFF
--- a/src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
@@ -44,8 +44,11 @@ import org.bson.BsonDocument;
 import org.bson.BsonInt32;
 import org.bson.BsonString;
 import org.bson.Document;
+import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
 import org.yaml.snakeyaml.constructor.Constructor;
+import org.yaml.snakeyaml.inspector.TagInspector;
+import org.yaml.snakeyaml.nodes.Tag;
 
 public class DataLoader {
     public static final String TEST_DATA_DIRECTORY = "resources/integration_test/testdata";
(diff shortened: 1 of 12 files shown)


terakilobyte (approved):
LGTM, nice work! I can't find anything that strikes me as needing a comment.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
```diff
@@ -57,7 +60,20 @@ public class DataLoader {
                     + ":"
                     + System.getenv("ADF_TEST_LOCAL_PWD")
                     + "@localhost";
-    private static Yaml yaml = new Yaml(new Constructor(TestData.class));
+    private static Yaml yaml;
+
+    static {
+        TagInspector allowGlobalTags =
```
bucaojit, line 66: Nice thanks for upgrading snakeyaml.  If I understand correctly, is this change to support the use of tags in the yaml such as `!!org.bson.BsonTimestamp`?

src/main/java/com/mongodb/jdbc/MongoResultSetMetaData.java
```diff
@@ -118,51 +119,68 @@ private void processDataSource(
...
-    // in MongoResultSet to retrieve data by label.
-    protected String getDatasource(String columnLabel) {
-        return columnLabels.get(columnLabel).datasource;
+    // This gets the datasource for a given columnLabel.
+    // It is only used in our unit tests.
+    protected String getDatasource(String columnLabel) throws Exception {
+        List<DatasourceAndIndex> columnsForLabel = columnLabels.get(columnLabel);
+        if (columnsForLabel.size() > 1) {
+            throw new Exception(
+                    "Multiple columns with the same label exists. Please use indexes instead.");
```
bucaojit, line 180:
```suggestion
throw new Exception(String.format(
            "Multiple columns with the label '%s' exist. Use indexes to avoid ambiguity.", columnLabel));
```
How about adding the label in the exception?  Also suggesting a change to the rest of the message which you can accept if you prefer it.

bucaojit (approved):
LGTM.  
Nice work on the snakeyaml and test changes.

src/integration-test/java/com/mongodb/jdbc/integration/testharness/DataLoader.java
nbagnard: ah, yes, you are correct that's what it's for.
I should have mentioned that. I also marked Snakeyaml as test dependency only so that it is not bundled and also not an issue to allow global tags.

I upgraded SnakeYaml and guava because they both have vulnerabilities reported.
I did that when listing dependencies and forgot to split the PR.

PR #248, opened 2024-04-08
SQL-1939: Create spawnable RFC8252 server

bucaojit:
Test it by running the server:
```
./gradlew task runServer
```
And testing the endpoints
Accepted result and return the parameters:
http://localhost:27097/redirect?code=wabberjocky&state=foobar

Error page:
http://localhost:27097/redirect?error=someerror&error_description=couldnotauth


src/main/java/com/mongodb/jdbc/oidc/OidcResponse.java
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.jdbc.oidc;
+
```
pmeredit, line 18: We should probably put a TODO to remove these classes once the Java driver changes land. I don't think we'll forget, but it's worth having here.

src/main/java/com/mongodb/jdbc/oidc/RFC252ServerMain.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+package com.mongodb.jdbc.oidc;
+
+import java.io.IOException;
+/**
+ * Main class to start the RFC8252HttpServer and wait for the OIDC response Used for testing the
+ * serving of the HTML pages and the OIDC response
+ */
+public class RFC252ServerMain {
```
pmeredit, line 24:
```suggestion
public class TestRFC252ServerMain {
```

We can go ahead and name this `Test` to call out that we are only using this for testing (like we did on the ODBC side, also the file name will have to change, of course)

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
```diff
@@ -0,0 +1,256 @@
...
+ * handles the OIDC authorization code flow by providing endpoints for the callback and redirection.
+ * The server listens on a specified port (default is 27017) and processes incoming HTTP requests.
+ */
+public class RFC8252HttpServer {
+    public static final int DEFAULT_REDIRECT_PORT = 27097;
+
+    // SQL-2008: make sure this page exists and possibly update the link if the
+    // docs team has a preference
+    private static final String LOGIN_ERROR_URI =
+            "https://www.mongodb.com/docs/atlas/data-federation/query/sql/drivers/odbc/connect/oidc_login_error";
```
pmeredit, line 48:
```suggestion
            "https://www.mongodb.com/docs/atlas/security-oidc";
```

pmeredit (approved):
Looks great! Just a few changes to make, but no reason to retag me since they are all minor

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
```diff
@@ -0,0 +1,256 @@
...
+     * @param statusCode the HTTP status code
+     * @throws IOException if an I/O error occurs while sending the response
+     */
+    private void sendResponse(HttpExchange exchange, String response, int statusCode)
+            throws IOException {
+        exchange.getResponseHeaders().set("Content-Type", "text/html; charset=utf-8");
+        exchange.sendResponseHeaders(statusCode, response.getBytes(StandardCharsets.UTF_8).length);
+        OutputStream os = exchange.getResponseBody();
+        os.write(response.getBytes(StandardCharsets.UTF_8));
+        os.close();
```
nbagnard, line 254: Do you need to close the outputstream or should you only close the channel?
In any case, I think you need to have try/catch or finally blocks here or in callers to make sure that you always close the exchange even when a exception is raised.
Otherwise, we will be leaking connections I think because the exchange will never be closed and so the outputstream and connection will stay open.

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
```diff
@@ -0,0 +1,256 @@
...
+        @Override
+        public void handle(HttpExchange exchange) throws IOException {
+            Map<String, String> queryParams = parseQueryParams(exchange);
+            OidcResponse oidcResponse = new OidcResponse();
+
+            if (queryParams.containsKey(CODE)) {
+                oidcResponse.setCode(queryParams.get(CODE));
+                oidcResponse.setState(queryParams.getOrDefault(STATE, ""));
+                if (!putOidcResponse(exchange, oidcResponse)) {
+                    return;
```
nbagnard, line 136: We should close the exchange or the http connection will stay open, isn't it?

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
```diff
@@ -0,0 +1,256 @@
...
+     * Parses the query parameters from the HTTP exchange.
+     *
+     * @param exchange the HTTP exchange
+     * @return a map containing the parsed query parameters
+     * @throws UnsupportedEncodingException if the encoding is not supported
+     */
+    private Map<String, String> parseQueryParams(HttpExchange exchange)
+            throws UnsupportedEncodingException {
+        Map<String, String> queryParams = new HashMap<>();
+        String query = exchange.getRequestURI().getQuery();
```
nbagnard, line 206: If there are any `&` or `=` as part of the values passed in (we know it shouldn't be part of the keys at least) then the logic you have using split below won't work.
You need to use getRawQuery, especially since you are decoding the values already when processing the params, you are already all set.

nbagnard (changes requested):

build.gradle
```diff
@@ -162,6 +167,7 @@ dependencies {
     api "org.mongodb:mongodb-driver-sync:$mongodbDriverVersion"
     integrationTestImplementation "org.junit.jupiter:junit-jupiter:$junitJupiterVersion"
     integrationTestImplementation group: 'org.junit.jupiter', name: 'junit-jupiter-api', version: junitJupiterVersion
+    implementation 'org.thymeleaf:thymeleaf:3.1.2.RELEASE'
```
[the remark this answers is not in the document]
bucaojit: Sure, will do.  I had it in mind to do when I added the dependency and forgot about doing it.  Thanks for catching that 👍

src/main/java/com/mongodb/jdbc/oidc/OidcResponse.java
bucaojit: Added  TODO to remove the class

src/main/java/com/mongodb/jdbc/oidc/RFC252ServerMain.java
bucaojit: Sounds good, renamed it.

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
bucaojit: I see.  I added a try with resource for the output stream and added a `finally` for closing the exchange.

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
bucaojit: The `sendResponse()` will close it in the false case for `putOidcResponse()`.

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
bucaojit: I made this change and added some test cases for having a `&` or `=` in the value.  Good catch.

src/main/java/com/mongodb/jdbc/oidc/RFC8252HttpServer.java
```diff
@@ -0,0 +1,256 @@
...
+
+            if (queryParams.containsKey(CODE)) {
+                oidcResponse.setCode(queryParams.get(CODE));
+                oidcResponse.setState(queryParams.getOrDefault(STATE, ""));
+                if (!putOidcResponse(exchange, oidcResponse)) {
+                    return;
+                }
+                exchange.getResponseHeaders().set(LOCATION, ACCEPTED_ENDPOINT);
+                exchange.sendResponseHeaders(HttpURLConnection.HTTP_MOVED_TEMP, -1);
+                exchange.close();
```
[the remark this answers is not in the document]
nbagnard: yes, and after doing a bit more research I believe that if we make sure to always close the http connection after each flow we should be ok.
My worry was that we would end up in a situation where after some kind of issue, users would get "socket already in use" every time. Looks like it shouldn't be a problem.

nbagnard (approved):
LGTM!

PR #254, opened 2024-04-11
SQL-1953: Fix the build script to update the json feed correctly

nbagnard:
The issue was that the shell type was not specified on the Evergreen task and we used a specific Bash syntax.

Here is a task which I ran in a debug branch with the tag debug0.0.0 which work:
https://spruce.mongodb.com/task/mongo_jdbc_driver_test_version_from_tag_release_release_smoke__debug0.0.0_24_04_11_23_36_00/logs?execution=0


terakilobyte (approved):
wow, nice sleuthing!

PR #255, opened 2024-04-12
SQL-1940: Add doAuthCodeFlow for OIDC authentication

bucaojit:
Authentication flow can be tested by running:
```
./gradlew runAuthFlow
```

The server will open a `Login Succeeded` and the terminal will show the access and refresh token.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/oidc/IdpServerInfo.java
+++ b/src/main/java/com/mongodb/jdbc/oidc/IdpServerInfo.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2024-present MongoDB, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.mongodb.jdbc.oidc;
+
+import java.util.List;
+
+// TODO: This class is a placeholder for the IdpServerInfo,
+//       it will be removed when Java Driver OIDC support is added.
+public class IdpServerInfo {
+    private String issuer;
+    private String clientId;
+    private List<String> requestScopes;
+
+    public IdpServerInfo(String issuer, String clientId, List<String> requestScopes) {
+        this.issuer = issuer;
+        this.clientId = clientId;
+        this.requestScopes = requestScopes;
+    }
+
+    public String getIssuer() {
+        return issuer;
+    }
+
+    public String getClientId() {
+        return clientId;
+    }
+
+    public List<String> getRequestScopes() {
+        return requestScopes;
+    }
+}
--- a/src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
+++ b/src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
@@ -35,16 +35,17 @@ import com.nimbusds.openid.connect.sdk.OIDCTokenResponseParser;
 import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
 import java.awt.*;
 import java.net.URI;
+import java.time.Duration;
 import java.util.List;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
-public class OIDCAuthFlow {
+public class OidcAuthFlow {
 
-    private static final Logger logger = Logger.getLogger(OIDCAuthFlow.class.getName());
+    private static final Logger logger = Logger.getLogger(OidcAuthFlow.class.getName());
 
-    public OIDCCredential doAuthCodeFlow(CallbackContext callbackContext) {
-        IdpServerInfo idpServerInfo = callbackContext.getIdpInfo();
+    public OidcCallbackResult doAuthCodeFlow(OidcCallbackContext callbackContext) {
+        IdpInfo idpServerInfo = callbackContext.getIdpInfo();
         if (idpServerInfo == null) {
             logger.severe("IdpServerInfo is null");
             return null;
@@ -109,7 +110,7 @@ public class OIDCAuthFlow {
             }
 
             // Wait for the authorization response from the local HTTP server.
-            OIDCResponse response = server.getOidcResponse();
+            OidcResponse response = server.getOidcResponse();
             if (response == null || !state.getValue().equals(response.getState())) {
                 logger.severe("OIDC response is null or returned an invalid state");
                 return null;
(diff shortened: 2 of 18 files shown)


src/main/java/com/mongodb/jdbc/oidc/IdpServerInfo.java
```diff
@@ -0,0 +1,45 @@
...
+
+package com.mongodb.jdbc.oidc;
+
+import java.util.List;
+
+// TODO: This class is a placeholder for the IdpServerInfo,
+//       it will be removed when Java Driver OIDC support is added.
+public class IdpServerInfo {
+    private String issuer;
+    private String clientId;
```
pmeredit, line 25: The spec changed to where clientId is now optional, and we need to return some sort of error when this happens because it means human flow is no enabled

pmeredit (changes requested):

src/main/java/com/mongodb/jdbc/oidc/IdpServerInfo.java
bucaojit: Okay thanks for letting me know I'll make the change.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+        // mongodb is not configured to ask for offline_access by default. We prefer always getting a
+        // refresh token when the server allows it.
+        scopes.add(new Scope.Value("offline_access"));
+
+        RFC8252HttpServer server = new RFC8252HttpServer();
+        try {
+            // Resolve OIDC provider metadata using the issuer URI and
+            // extract authorization and token endpoint URIs.
+            OIDCProviderMetadata providerMetadata =
```
pmeredit, line 81: We should make sure that scopes we are requesting are in the supported scopes: https://www.javadoc.io/doc/com.nimbusds/oauth2-oidc-sdk/5.0/com/nimbusds/openid/connect/sdk/op/OIDCProviderMetadata.html#getScopes()

For instance, if the IdP does not support offline_access (refresh) we don't want to just fail, we would prefer just not being able to refresh.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -0,0 +1,159 @@
...
+                                    new ClientID(clientID))
+                            .scope(scopes)
+                            .redirectionURI(redirectURI)
+                            .state(state)
+                            .codeChallenge(codeVerifier, CodeChallengeMethod.S256)
+                            .endpointURI(authorizationEndpoint)
+                            .build();
+
+            // Open the browser to the authorization request URI.
+            if (Desktop.isDesktopSupported()) {
```
pmeredit, line 110: I should figure out how to do this in ODBC. Right now ODBC just hangs if it can't launch a browser (you'd think `open` would return an error)

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -0,0 +1,159 @@
...
+                Desktop.getDesktop().browse(request.toURI());
+            } else {
+                logger.severe("Desktop operations not supported");
+                return null;
+            }
+
+            // Wait for the authorization response from the local HTTP server.
+            OidcResponse response = server.getOidcResponse();
+            if (response == null || !state.getValue().equals(response.getState())) {
```
pmeredit, line 119: To be clear, this is the client side (us) confirming that no man-in-the-middle attack occurred. This is a very important step. The code verifier is how the IdP ensures no man-in-the-middle attack occurred, since it could happen in either direction

pmeredit (changes requested):
Looking good, just a few more changes.

nbagnard (changes requested):
It's looking pretty good already :+1:
I have a few questions to make sure I understand things correctly and a couple of small requests. It should be good to go in no time :)

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -0,0 +1,159 @@
...
+import com.nimbusds.oauth2.sdk.id.ClientID;
+import com.nimbusds.oauth2.sdk.id.Issuer;
+import com.nimbusds.oauth2.sdk.id.State;
+import com.nimbusds.oauth2.sdk.pkce.CodeChallengeMethod;
+import com.nimbusds.oauth2.sdk.pkce.CodeVerifier;
+import com.nimbusds.oauth2.sdk.token.Tokens;
+import com.nimbusds.openid.connect.sdk.OIDCTokenResponse;
+import com.nimbusds.openid.connect.sdk.OIDCTokenResponseParser;
+import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
+import java.awt.*;
```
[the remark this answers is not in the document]
bucaojit: I missed this one, I was too focused on removing the wildcards from the `com.nimbusds.oauth2` imports, I'll update this.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
bucaojit: I added a change to check to use getScopes() on the metadata and add the requested scope if it's supported.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -0,0 +1,159 @@
...
+import java.net.URI;
+import java.time.Duration;
+import java.util.List;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class OidcAuthFlow {
+
+    private static final Logger logger = Logger.getLogger(OidcAuthFlow.class.getName());
```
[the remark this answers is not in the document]
bucaojit: Hmm.  How about I add a OidcAuthFlow constructor  with an optional MongoLogger parameter and set that as the parentLogger so when this gets integrated we can pass in the internal handler?  For the manual test case I can have it use the default handler, which logs to console.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -0,0 +1,175 @@
...
+            // Open the browser to the authorization request URI.
+            if (Desktop.isDesktopSupported()) {
+                Desktop.getDesktop().browse(request.toURI());
+            } else {
+                logger.severe("Desktop operations not supported");
+                return null;
+            }
+
+            // Wait for the authorization response from the local HTTP server.
+            OidcResponse response = server.getOidcResponse(Duration.ofMinutes(5));
```
pmeredit, line 132: Let's only use 5 minutes when the `OidcCallbackContext` does not contain a timeout.

pmeredit (approved):
Looks good to me once we use the Duration from the CallbackContext

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
bucaojit: Thanks, I meant to do that after Natacha's comment.  It's set now.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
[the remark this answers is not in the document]
bucaojit: Okay, I implemented this.  Let me know if that works for you.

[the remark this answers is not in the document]
nbagnard: I see you can do both, that's great! :)

nbagnard (approved):
LGTM!

PR #258, opened 2024-04-22
SQL-2049: Make all commit related logic no-op and report auto-commit = true instead of false.

nbagnard:
I tested with an older version of DBeaver and a newer version.
The driver works with both versions with this change. It fails as described when using an older driver with the newer DBeaver.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -227,7 +227,7 @@ public class MongoConnection implements Connection {
         checkConnection();
         logger.log(
                 Level.WARNING,
-                "Changing the auto-commit mode has no effect. The driver doesn't support transactions and is read-only."
+                "Changing the auto-commit mode has no effect. The driver doesn't support transactions and is read-only. "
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -223,26 +223,31 @@ public PreparedStatement prepareStatement(String sql) throws SQLException {
...
     public void setAutoCommit(boolean autoCommit) throws SQLException {
+        // no-op, we only check that the connection is open
         checkConnection();
-        if (autoCommit) {
-            throw new SQLFeatureNotSupportedException(
-                    Thread.currentThread().getStackTrace()[1].toString());
-        }
+        logger.log(
+                Level.WARNING,
+                "Changing the auto-commit mode has no effect. The driver doesn't support transactions and is read-only."
```
bucaojit, line 230:
```suggestion
                "Changing the auto-commit mode has no effect. The driver doesn't support transactions and is read-only. "
```
nit: adding a space after period.

bucaojit (approved):
LGTM

pmeredit (approved):
LGTM!

PR #260, opened 2024-04-26
SQL-1942: Conditionally add OIDC callback

bucaojit:
Tested OIDC through Tableau and with the :demo test, connecting to `sqlenginesoidctest-lxonw.a.query.mongodb-dev.net`

Upgraded version to java driver `5.1.0` which has the OIDC changes.


src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -209,6 +216,33 @@ void testReplURL() throws SQLException {
...
+    void testMongoDBOIDCAuthMechanismWithPassword() throws SQLException {
+        MongoDriver d = new MongoDriver();
+        Properties p = new Properties();
+
+        assertThrows(
+                SQLException.class,
+                () -> d.getUnvalidatedConnection(userPwdOIDCURL, p),
+                "The connection should fail because password is not allowed with MONGODB-OIDC authentication.");
+    }
+
```
bucaojit, line 245: Working on adding some more tests.

bucaojit:
One of the unit tests are failing expecting a different log line.  Changes here didn't update the Logger, I'll see what caused the difference and update the test if needed.

src/main/java/com/mongodb/jdbc/oidc/JdbcOidcCallbackContext.java
```diff
@@ -16,24 +16,28 @@
...
-    public OidcCallbackContext(
+    public JdbcOidcCallbackContext(
             Duration timeout, int version, String refreshToken, IdpInfo idpInfo) {
         this.timeout = timeout;
         this.version = version;
         this.refreshToken = refreshToken;
         this.idpInfo = idpInfo;
     }
 
+    public String getUserName() {
```
pmeredit, line 37: Should this be empty string or null? I'm not sure, but let's be certain. Oh, also, OIDC _can_ have user names even though our test server doesn't require them (because it gets them from the IdP).

pmeredit (approved):
This looks great. Just a few questions, but feel free to merge once you address them one way or another.

src/main/java/com/mongodb/jdbc/oidc/JdbcOidcCallbackContext.java
nbagnard: The interface annotation has @nullable, so Null seems better.
And I agree with Patrick, if providing a username is possible, we should should support that too. Especially since it looks like you allow for a username to be provided via the connection settings for OIDC and only reject password.

src/main/java/com/mongodb/jdbc/oidc/manualtests/TestOidcCallbackWithBadRefreshToken.java
```diff
@@ -16,26 +16,25 @@
...
         String badRefreshToken = "bad-refresh-token";
         OidcCallbackContext context =
-                new OidcCallbackContext(null, 1, badRefreshToken, TestOidcUtils.IDP_INFO);
+                new JdbcOidcCallbackContext(null, 1, badRefreshToken, TestOidcUtils.IDP_INFO);
 
         try {
             OidcCallbackResult result = oidcCallback.onRequest(context);
             System.out.println("This should not print, bad refresh token expected to fail.");
             System.out.println(result);
-        } catch (RefreshFailedException e) {
```
nbagnard, line 37: I think that you should keep this check. You'll need to catch Runtime exceptions and check the cause. It should be an instance of RefreshFailedException or it means that the exception thrown is not what we expect, and therefore the test should "fail".

src/main/java/com/mongodb/jdbc/oidc/manualtests/TestOidcCallbackWithShortTimeout.java
```diff
@@ -38,8 +39,6 @@ public static void main(String[] args) {
             System.out.println(
                     "This should not print, timeout expected. Sign out of the IdP to trigger a timeout.");
             System.out.println(result);
-        } catch (OidcTimeoutException e) {
```
nbagnard, line 41: Here too, you want to catch runtime exception and check the cause to verify that it is the excepted reason for the failure.

nbagnard (changes requested):

src/integration-test/java/com/mongodb/jdbc/integration/MongoIntegrationTest.java
```diff
@@ -52,7 +52,11 @@ public class MongoIntegrationTest {
     private static final String CURRENT_DIR =
             Paths.get(".").toAbsolutePath().normalize().toString();
 
-    static final String URL = "jdbc:mongodb://localhost";
+    private static final String LOCAL_HOST = "localhost";
```
bucaojit, line 55: Added this to be able to configure the hostname.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -102,9 +106,18 @@ public String getPropertyName() {
                     });
     static final String RELAXED = "RELAXED";
     static final String EXTENDED = "EXTENDED";
+    static final String MONGODB_OIDC = AuthenticationMechanism.MONGODB_OIDC.toString();
     public static final String LOG_TO_CONSOLE = "console";
     protected static final String CONNECTION_ERROR_SQLSTATE = "08000";
 
+    // The cache of MongoClient instances, keyed on the connection properties.
+    private static ConcurrentHashMap<Integer, MongoClient> mongoClientCache =
```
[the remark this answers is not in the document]
bucaojit: Sounds good, I changed it to use `WeakReference<MongoClient>` and added some synchronization code.

src/main/java/com/mongodb/jdbc/oidc/JdbcOidcCallbackContext.java
bucaojit: Okay, I added the username in the constructor similar to the [java driver](https://github.com/mongodb/mongo-java-driver/blob/790185f9ed64fed4e5720e2bbbff170d247a7374/driver-core/src/main/com/mongodb/internal/connection/OidcAuthenticator.java#L660), too bad their class isn't public.  For us, this `JdbcOidcCallbackContext` class is only used in the manual OIDC tests.

src/main/java/com/mongodb/jdbc/MongoConnectionProperties.java
```diff
@@ -66,4 +66,18 @@ public String getClientInfo() {
...
     }
+
+    /*
+     * Generate a unique key for the connection properties. This key is used to identify the connection properties in the
+     * connection cache. Properties that do not differentiate a specific client such as the log level are not included in the key.
+     */
+    public Integer generateKey() {
+        StringBuilder keyBuilder = new StringBuilder();
+        keyBuilder.append(connectionString.toString());
+        keyBuilder.append(":db=").append(database);
```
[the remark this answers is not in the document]
bucaojit: Okay, I'll check with Alexi next week.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -63,7 +66,8 @@ public enum MongoJDBCProperty {
         CLIENT_INFO("clientinfo"),
         LOG_LEVEL("loglevel"),
         LOG_DIR("logdir"),
-        EXT_JSON_MODE("extjsonmode");
+        EXT_JSON_MODE("extjsonmode"),
+        AUTH_MECHANISM("authMechanism");
```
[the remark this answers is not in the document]
bucaojit: That works for me.  It's not needed as a property, I've removed it.

src/main/java/com/mongodb/jdbc/oidc/manualtests/TestOidcCallbackWithBadRefreshToken.java
bucaojit: Sure thing, I was being lazy when I updated the `JdbcOidcCallback` class to remove the `throws` and then removed the corresponding catch here.  I added the check for `instance of RefreshFailedException`

src/main/java/com/mongodb/jdbc/oidc/manualtests/TestOidcCallbackWithShortTimeout.java
bucaojit: Same here, added the check for `instanceof OidcTimeoutException`

nbagnard (approved):
LGTM!
Nice use of read-write locks 👍

PR #262, opened 2024-06-03
MHOUSE-11099: Update `run-adf.sh` inline local adf config name

awang2330:
We're changing references to  the resource manager in ADF from acronyms `CRM`/`RM` to the full name, see [MHOUSE-11099](https://github.com/10gen/mongohouse/pull/9795). This includes config names so the referenced inline local file used by the SQL team also needs to be updated. This is necessary until the work in [MHOUSE-11063](https://jira.mongodb.org/browse/MHOUSE-11063) is done.


nbagnard (approved):
LGTM!

PR #263, opened 2024-06-07
SQL-2091: Add SBOM Lite for JDBC

EthanHardyMongo:
Successful run: https://spruce.mongodb.com/task/mongo_jdbc_driver_code_quality_and_correctness_sbom_patch_65d75db0c5e89aab2d87ef90e67a3e0d12571c74_666375a3d3b2b1000758d666_24_06_07_21_03_39/logs?execution=0

SBOM created: https://evg-bucket-mongo-jdbc-driver.s3.amazonaws.com/artifacts/666375a3d3b2b1000758d666/code-quality-and-correctness/ssdlc/mongo-jdbc-driver.cdx.json

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -101,7 +101,6 @@ tasks:
   - name: sbom
     commands:
       - func: "generate sbom"
-      - func: "add team name to sbom"
       - func: "upload sbom"
       - func: "scan sbom"
 
@@ -123,8 +122,6 @@ functions:
           echo "Generating SBOM"
           echo "------------------------------------"
           ./gradlew cyclonedxBom
-
-  "add team name to sbom":
     - command: shell.exec
       type: test
       params:
@@ -148,7 +145,7 @@ functions:
           echo "------------------------------------"
           echo "Adding team name to SBOM"
           echo "------------------------------------"
-          jq '.components[].properties += [{"name": "internal:team_responsible", "value": "Atlas SQL"}]' $SBOM_WITHOUT_TEAM_NAME > $SBOM_FINAL
+          jq '.components[].properties += [{"name": "internal:team_responsible", "value": "Atlas SQL"}]' $SBOM_WITHOUT_TEAM_NAME > $SBOM_LITE
           
           echo "------------------------------------"
           echo "Done generating SBOM"
@@ -187,7 +184,7 @@ functions:
         echo "------------------------------------"
         
         echo "-- Scanning dependency for vulnerabilities --"
-        ./$SBOM_TOOL_DIR/grype sbom:$SBOM_FINAL --fail-on low
+        ./$SBOM_TOOL_DIR/grype sbom:$SBOM_LITE --fail-on low
         echo "---------------------------------------------"
         echo "<<<< Done scanning SBOM"
 
@@ -310,7 +307,7 @@ functions:
             # sbom relevant variables
             export SBOM_TOOL_DIR=sbom_generations
             export SBOM_DIR=$ARTIFACTS_DIR/sboms
-            export SBOM_FINAL=$ARTIFACTS_DIR/sboms/mongo-jdbc-driver.cdx.json
+            export SBOM_LITE=$ARTIFACTS_DIR/sboms/mongo-jdbc-driver.cdx.json
             export SBOM_WITHOUT_TEAM_NAME=$ARTIFACTS_DIR/sboms/sbom_without_team_name.json
           EOT
 
--- a/.evg.yml
+++ b/.evg.yml
@@ -121,7 +121,7 @@ functions:
           echo "------------------------------------"
           echo "Generating SBOM"
           echo "------------------------------------"
-          ./gradlew cyclonedxBom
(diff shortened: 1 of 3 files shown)


.evg.yml
```diff
@@ -212,6 +306,12 @@ functions:
             export JAVA_HOME=${JAVA_HOME}
             export PROJECT_DIRECTORY=${PROJECT_DIRECTORY}
             export MDBJDBC_VER=${MDBJDBC_VER}
+          
+            # sbom relevant variables
+            export SBOM_TOOL_DIR=sbom_generations
+            export SBOM_DIR=$ARTIFACTS_DIR/sboms
+            export SBOM_FINAL=$ARTIFACTS_DIR/sboms/mongo-jdbc-driver.cdx.json
```
nbagnard, line 313: [nit] Let's call this SBOM_LITE from now on to match DEVPROD naming convention.

.evg.yml
```diff
@@ -95,7 +98,98 @@ tasks:
     commands:
       - func: "static code analysis"
 
+  - name: sbom
+    commands:
+      - func: "generate sbom"
+      - func: "add team name to sbom"
```
nbagnard, line 104: I think you can merge this func with the previous one. It doesn't have to be so granular.
I also don't know how I feel with using func versus having a list of commands being part of a function.
It makes the func longer, but at the same time, having all piece be under the same function means that there it is clear that they have to happen in that order and that what the previous function produces is expected to exist for the next function.

.evg.yml
```diff
@@ -212,6 +306,12 @@ functions:
             export JAVA_HOME=${JAVA_HOME}
             export PROJECT_DIRECTORY=${PROJECT_DIRECTORY}
             export MDBJDBC_VER=${MDBJDBC_VER}
+          
+            # sbom relevant variables
+            export SBOM_TOOL_DIR=sbom_generations
+            export SBOM_DIR=$ARTIFACTS_DIR/sboms
+            export SBOM_FINAL=$ARTIFACTS_DIR/sboms/mongo-jdbc-driver.cdx.json
+            export SBOM_WITHOUT_TEAM_NAME=$ARTIFACTS_DIR/sboms/sbom_without_team_name.json
```
nbagnard, line 314: Defining this variable isn't very useful since the Gradle script doesn't use this value.
Variables are great if you want to guarantee that names are the same everywhere and to make maintenance easier.

You can however using it with Gradle and that would be the perfect use of a variable. Although I think the name of the temp file doesn't have to be shared with all task and can simply be defined as part of the func.

I pushed an example on my local branch. I had it ready but never pushed it on my remote branch. Look at both the Evergreen script (line 466) and the gradle.properties file
https://github.com/mongodb/mongo-jdbc-driver/compare/master...nbagnard:mongo-jdbc-driver:ssdlc-poc#diff-73e7b844e5fa69d8104ed91b7c281ea7262bc60bd7be03d515048aff04478fe6R435

build.gradle
```diff
@@ -22,6 +23,31 @@ spotless {
...
+    // Boms destination directory. Defaults to 'build/reports'
+    destination = file("artifacts/sboms")
+    // The file name for the generated BOMs (before the file format suffix). Defaults to 'bom'
+    outputName = "sbom_without_team_name"
+    // The file format generated, can be xml, json or all for generating both. Defaults to 'all'
+    outputFormat = "json"
+    // Exclude BOM Serial Number. Defaults to 'true'
+    includeBomSerialNumber = false
+    // Exclude License Text. Defaults to 'true'
+    includeLicenseText = true
```
nbagnard, line 46: Change this to false. I don't think that we need it.

build.gradle
```diff
@@ -22,6 +23,31 @@ spotless {
...
+    // Override component version. Defaults to the project version
+    componentVersion = "2.0.0"
```
nbagnard, line 48: This variable is not necessary. It will pick up the project version and that's totally fine.

nbagnard (changes requested):
It's looking good.
I have only a few comments to help make our lives easier in the long run

.evg.yml
EthanHardyMongo: Originally, it was one function; however, it would stop executing after `./gradlew cyclonedxBom` for some reason, so I made two separate functions. However, I can combine the two and just have two `- command: shell.exec` in one function.

.evg.yml
EthanHardyMongo: I defined the output name and destination in the `build.gradle` file:
```
cyclonedxBom {
    // Boms destination directory. Defaults to 'build/reports'
    destination = file("artifacts/sboms")
    // The file name for the generated BOMs (before the file format suffix). Defaults to 'bom'
    outputName = "sbom_without_team_name"
}
```
These environmental variables were made to make the evergreen functions easier to read.

nbagnard: ah, I didn't realize that the variables were case insensitive 👍 
I think `destination` is not using the `SBOM_DIR` variable, could you add it and then think would be perfect!

nbagnard (approved):
LGTM!
Pending the last little `destination` variable to sync up between the evergreen script and gradle.properties

PR #264, opened 2024-06-10
SQL-2105: Trace JDBC releases using the Papertrail service

EthanHardyMongo:
Since the releases are created by gradle plugins, testing the tracing seems to be pretty complicated, so we've decided to skip the testing for now. At release time, we will adjust/fix if necessary.


nbagnard (approved):
LGTM!

PR #267, opened 2024-06-19
SQL-2143: Update JDBC Sonatype publishing to use token

bucaojit:
Changes are mainly updating the variable name.  I added the NEXUS_TOKEN_NAME and  NEXUS_TOKEN variables in evergreen. 

A few things, the `nexus-publish` plugin we're using is quite old and doesn't have documentation on whether we can put token name/token password in place of username/password, the successor of the project [gradle-nexus](https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-ossrh) seems to suggest it can be.  So I think we can try it at release time and if it fails look into changing the plugin to `gradle-nexus` or `maven-publish` which is what the java driver uses.


nbagnard (approved):
LGTM!

nbagnard:
> the `nexus-publish` plugin we're using is quite old and doesn't have documentation on whether we can put token name/token password in place of username/password, the successor of the project [gradle-nexus](https://github.com/gradle-nexus/publish-plugin?tab=readme-ov-file#publishing-to-maven-central-via-sonatype-ossrh) seems to suggest it can be. So I think we can try it at release time and if it fails look into changing the plugin to `gradle-nexus`

Let's create a JIRA ticket to swap the plugin for the newer one. If the release succeeds then it can be done later, if it fails, then we will have to address the issue earlier.

Looking at the release notes, we might be able to use [1.3.0](https://github.com/gradle-nexus/publish-plugin/releases/tag/v1.3.0) without problem. If we want to use 2.0 or later, we might have to use to `ubuntu2204-64-jdk-11` instead of `ubuntu2204-64-jdk-8` (which I think is time to do too)

bucaojit:
Sounds good, created https://jira.mongodb.org/browse/SQL-2145 to cover updating the plugin.

PR #271, opened 2024-06-24
SQL-2108: Generate and publish a public third-party dependencies report for JDBC

EthanHardyMongo:


nbagnard (approved):
LGTM!
I can not tell with certainty if this will work, we will have to try a release.

PR #273, opened 2024-06-27
SQL-2185: Generate and publish a public third-party dependencies report for JDBC

EthanHardyMongo:


.evg.yml
```diff
@@ -225,7 +223,7 @@ functions:
         aws_key: ${release_aws_key}
         aws_secret: ${release_aws_secret}
         local_file: artifacts/ssdlc/mongodb-jdbc-${MDBJDBC_VER}.sbom.json
-        remote_file: mongodb-jdbc/mongodb-jdbc-${MDBJDBC_VER}.sbom.json
+        remote_file: mongo-jdbc-driver/mongodb-jdbc-${MDBJDBC_VER}.sbom.json
         content_type: application/json
```
EthanHardyMongo, line 227: Drive by fix from [this ticket](https://jira.mongodb.org/browse/SQL-2182).

gradle/publish.gradle
```diff
@@ -1,6 +1,8 @@
 apply plugin: 'signing'
 apply plugin: 'io.codearte.nexus-staging'
 
+def augmentedSbom = artifacts.add('sbom', file('artifacts/ssdlc/mongo-jdbc-driver.augmented.sbom.json'))
```
nbagnard, line 4: Take a look at this page : https://docs.gradle.org/current/userguide/working_with_files.html
I think you will need to use [Project.getRootDir()](https://docs.gradle.org/current/dsl/org.gradle.api.Project.html#org.gradle.api.Project:rootDir).
My understanding is that otherwise it is the path relative to where the gradle script is

nbagnard (changes requested):

gradle/publish.gradle
EthanHardyMongo: For reference, `$rootDir`, is the same as `Project.getRootDir()`.

gradle/publish.gradle
```diff
@@ -1,6 +1,8 @@
 apply plugin: 'signing'
 apply plugin: 'io.codearte.nexus-staging'
 
+def augmentedSbom = artifacts.add('sbom', file("$rootDir/artifacts/ssdlc/mongo-jdbc-driver.augmented.sbom.json"))
```
nbagnard, line 4: I hope the script doesn't clean up this directory before running.

nbagnard (approved):
I think that should work if the artifact dir is actually there 👍

PR #278, opened 2024-08-06
SQL-2259: Output UUID String representation to match ODBC driver

bucaojit:
Changed the $uuid to be represented in a way that matches what Compass and the ODBC driver displays rather than the base64 binary format. 

Added integration tests to confirm it works with the different uuidRepresentation options.


.evg.yml
```diff
@@ -355,6 +355,7 @@ functions:
     command: shell.exec
     type: test
     params:
+      shell: bash
```
bucaojit, line 358: Drive by fix, saw in evergreen it was complaining about the SECONDS variable which is supported in `bash` and not `sh`.

resources/integration_test/tests/dbmd_variable_result_sets_data_methods.yml
```diff
@@ -66,7 +66,7 @@ tests:
 
   - description: getTables_null_parameters_returns_all_tables
     db: integration_test
-    meta_function: [ getTables, null, null, null, null ]
+    meta_function: [getTables, null, null, null, null]
```
bucaojit, line 69: Changes in these integration test files are due to adding a new `uuid` collection.

src/main/java/com/mongodb/jdbc/MongoBsonValue.java
```diff
@@ -119,6 +134,32 @@ public String toString() {
...
+            UuidRepresentation representationToUse =
+                    this.uuidRepresentation != UuidRepresentation.UNSPECIFIED
+                            ? this.uuidRepresentation
+                            : UuidRepresentation.STANDARD;
+            uuid =
+                    UuidHelper.decodeBinaryToUuid(
+                            binary.getData(), binary.getType(), representationToUse);
+        }
+
+        String uuidStr = uuid.toString();
```
bucaojit, line 156: This is something Alexi and I have been trying to decide on.  It is a better experience to have only the `uuid` in the field for use in the query instead of having to add SUBSTRING() commands to remove the `{$uuid: }` parts.  Tableau does not have the easy transform commands as PowerBI has, at least from what we tried.

terakilobyte (approved):
LGTM. One question, what's the default for the old subtype? I didn't see it here, but it should probably be PYTHON_LEGACY to match ODBC.

src/integration-test/java/com/mongodb/jdbc/integration/MongoIntegrationTest.java
```diff
@@ -231,4 +243,76 @@ private void cleanUp(MongoConnection conn) {
...
     }
+
+    /**
+     * Tests the handling of different UUID representations specified in the URI. The uuid fields
+     * have been pre-loaded into the database, stored in their respective uuid representations
+     * according to their type. This test verifies that each representation is correctly retrieved
+     * and converted to the expected string format.
+     */
+    @Test
+    public void testUUIDRepresentation() {
```
nbagnard, line 254: You'll want to add a test for retrieving subtype 3 with UUID representation "Standard". It should fail.

Another good test would be to retrieve the data with 'type=javalegacy` using "javalegacy", "csharplegacy" and "pythonlegacy" and verify that the uuid value is different.

nbagnard (changes requested):

src/test/java/com/mongodb/jdbc/MongoMock.java
```diff
@@ -154,6 +161,27 @@ public abstract class MongoMock {
     protected static String ALL_MIN_KEY_COL_VAL = "{\"$minKey\": 1}";
     protected static String ALL_MAX_KEY_COL_VAL = "{\"$maxKey\": 1}";
 
+    protected static UUID UUID_VAL = UUID.fromString("00112233-4455-6677-8899-aabbccddeeff");
+    protected static String ALL_UUID_STRING_EXT_VAL =
+            "{\"$uuid\":\"00112233-4455-6677-8899-aabbccddeeff\"}";
+    protected static String ALL_UUID_STRING_VAL = "00112233-4455-6677-8899-aabbccddeeff";
```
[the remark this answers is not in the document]
bucaojit: Sure, made the change.

src/main/java/com/mongodb/jdbc/MongoBsonValue.java
```diff
@@ -119,6 +134,32 @@ public String toString() {
...
+        if (binary.getType() == BsonBinarySubType.UUID_STANDARD.getValue()) {
+            uuid = binary.asUuid();
+        } else {
+            // When UNSPECIFIED, set UuidRepresentation to STANDARD
+            UuidRepresentation representationToUse =
+                    this.uuidRepresentation != UuidRepresentation.UNSPECIFIED
+                            ? this.uuidRepresentation
+                            : UuidRepresentation.STANDARD;
+            uuid =
+                    UuidHelper.decodeBinaryToUuid(
```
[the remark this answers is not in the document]
bucaojit: Thanks for catching that.  I made the change to have it return the binary data with the format you suggested.

src/main/java/com/mongodb/jdbc/MongoBsonValue.java
```diff
@@ -119,6 +134,32 @@ public String toString() {
...
+    // If the BSON binary type is UUID_STANDARD, it directly converts it to a UUID.
+    // Otherwise, it uses the specified or default UUID representation to decode the binary data.
+    private String formatUuid(BsonBinary binary) {
+        UUID uuid;
+
+        if (binary.getType() == BsonBinarySubType.UUID_STANDARD.getValue()) {
+            uuid = binary.asUuid();
+        } else {
+            // When UNSPECIFIED, set UuidRepresentation to STANDARD
+            UuidRepresentation representationToUse =
```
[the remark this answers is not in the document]
bucaojit: Ah okay.  I did think it would be UNSPECIFIED.  I added a check for `null` and set the representation to PythonLegacy if none specified.

src/integration-test/java/com/mongodb/jdbc/integration/MongoIntegrationTest.java
bucaojit: Okay, I added the tests.  When retrieving subtype 3 with STANDARD representation, it now returns the format `{"$binary": {` that was suggested earlier. 
Added the test to retrieve the `javalegacy` data  with the different representation types.

nbagnard (approved):
LGTM!

PR #280, opened 2024-08-23
SQL-2286: Test mongodb+srv style connection string

mattChiaravalloti:
This PR adds an integration test for a `mongodb+srv` style uri. For now, I'm using the `nathan` user but will talk to Alexi today about creating a new dedicated test user on her cluster.

Also, for now I have it expecting the exception that is thrown since Atlas does not natively support the SQL commands. I put a TODO to update it when we eventually implement support. The test will also start failing when we implement support, so the TODO is almost redundant.

This took a while since I initially investigated using the AWS Secret Manager, but it was not as easy as the wiki page implied (the wiki makes reference to scripts without links so it's unclear which scripts they're talking about or what they're supposed to do). Instead, I opted to use evergreen as our secret manager like we do for all other values in our tests -- it's easier and more consistent for now. If we ever have to fully switch from evergreen secret managing to AWS, we should migrate everything at once. I don't think now is the time to make the change!

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -466,6 +466,7 @@ public class MongoDriver implements Driver {
         ConnectionString c =
                 new ConnectionString(
                         buildNewURI(
+                                originalConnectionString.isSrvProtocol(),
                                 originalConnectionString.getHosts(),
                                 user,
                                 password,
@@ -627,6 +628,7 @@ public class MongoDriver implements Driver {
      *     arguments user provided,
      */
     private static String buildNewURI(
+            boolean isSrvProtocol,
             List<String> hosts,
             String user,
             char[] password,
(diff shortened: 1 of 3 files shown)


src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -635,8 +637,8 @@ private static String buildNewURI(
             Properties options)
             throws SQLException {
         // The returned URI should be of the following format:
-        //"mongodb://[user:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[authDatabase][?options]]")
-        String ret = "mongodb://";
+        //"mongodb(+srv)?://[user:password@]host1[:port1][,host2[:port2],...[,hostN[:portN]]][/[authDatabase][?options]]")
+        String ret = isSrvProtocol ? "mongodb+srv://" : "mongodb://";
```
mattChiaravalloti, line 641: This was also something that caused a delay! Without this, we were hitting an unrelated timeout error. This doesn't actually enable Atlas support, of course, since we still attempt to use sql commands. This just ensures the error we get back now is "command not found: sqlGetResultSchema" as opposed to a timeout.

mattChiaravalloti:
The new integration test failed because it couldn't find the environment new variables 😵  But the new ones are set the same way as the old `ADF_*` env variables (which clearly worked again this time), so I'm not sure why they are not being found 🤔 Need to look into it further

mattChiaravalloti:
> The new integration test failed because it couldn't find the environment new variables 😵 But the new ones are set the same way as the old `ADF_*` env variables (which clearly worked again this time), so I'm not sure why they are not being found 🤔 Need to look into it further

I still haven't figured out the reason _why_ the previous variable value led to `SRV_TEST_HOST` to be empty, but I updated it in evergreen and updated the code to handle the new value and it _should_ be working now. I had a patch run essentially successfully, so let's see if that happens on the latest PR patch as well. Really odd behavior..

mattChiaravalloti:
@EthanHardyMongo @terakilobyte can confirm this is ready for review for real 👍 Everything's passing now with the latest commit.

EthanHardyMongo (approved):
LGTM!

terakilobyte (approved):

PR #281, opened 2024-08-28
SQL-2292: Load mongosqltranslate library if found

bucaojit:
Loads library with name `libmongosqltranslate` from the same directory as where the Driver is located.  If that is not found it then loads from the path specified by the `MONGOSQL_TRANSLATE_PATH` env variable.  

I added a JNI test that I was using to confirm it was working, and added the library itself built on Ubuntu x86.  

If the library cannot be loaded, the `mongoSqlTranslateLibraryLoaded` variable will be false.  This can be checked in future implementations that intend to use the library and can at that point error out and report that the library is not loaded. 

```
use jni::JNIEnv;
use jni::objects::{JClass, JString};
use jni::sys::jstring;

#[no_mangle]
pub extern "system" fn Java_com_mongodb_jdbc_MongosqlLibTest_runCommand(
    mut env: JNIEnv,
    _class: JClass,
    input: JString,
) -> jstring {
    let input: String = env.get_string(&input).expect("Couldn't get java string!").into();
    let output = format!("{{\"result\": \"mongosql translation test result success for input: {}\"}}", input);
    env.new_string(output).expect("Couldn't create java string!").into_raw()
}
```

### DIFF
--- a/src/test/java/com/mongodb/jdbc/MongoSQLTranslateLibTest.java
+++ b/src/test/java/com/mongodb/jdbc/MongoSQLTranslateLibTest.java
@@ -49,23 +49,4 @@ public class MongoSQLTranslateLibTest {
         MongosqlLibTest test = new MongosqlLibTest();
         test.testRunCommand();
     }
-
-    @Test
-    void testLibraryLoadingWithEnvironmentVariable() throws Exception {
-        String envPath = System.getenv(MongoDriver.MONGOSQL_TRANSLATE_PATH);
-        assertNotNull(envPath, "MONGOSQL_TRANSLATE_PATH should be set");
-
-        // Test initializeMongoSqlTranslateLibrary, with Environment variable set it should find the library
-        Method initMethod =
-                MongoDriver.class.getDeclaredMethod("initializeMongoSqlTranslateLibrary");
-        initMethod.setAccessible(true);
-        initMethod.invoke(null);
-
-        assertTrue(
-                MongoDriver.isMongoSqlTranslateLibraryLoaded(),
-                "Library should be loaded when MONGOSQL_TRANSLATE_PATH is set");
-
-        MongosqlLibTest test = new MongosqlLibTest();
-        test.testRunCommand();
-    }
 }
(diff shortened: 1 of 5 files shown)


nbagnard:
> If that is not found it then loads from the path specified by the `MONGOSQL_TRANSLATE_PATH` env variable.

Did we figure out that it is indeed needed for the driver to work with Tableau? Because I think the design decision was to defer adding the variable unless necessary

bucaojit:
@nbagnard I tested it on Mac with Tableau and it was able to find the library that I put in `~/Library/Tableau/Drivers `.   The env variable isn't needed, I guess I got carried away and implemented it.  I can remove the env variable logic.

src/test/java/com/mongodb/jdbc/MongoSQLTranslateLibTest.java
```diff
@@ -0,0 +1,71 @@
...
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class MongoSQLTranslateLibTest {
+    @BeforeEach
+    void setup() throws Exception {
```
nbagnard, line 28: Can you add a comment to explain that this because we need to reset the flag between each test?
I had to read this twice to realize that the setup was needed because the first test was loading the lib and setting the flag to true (hopefully).

nbagnard (approved):
Just a couple of small suggestions and let's discuss the environment variable. Now that it is implemented, it seems a waste to throw it away.

src/test/java/com/mongodb/jdbc/MongoSQLTranslateLibTest.java
bucaojit: Sure I'll add a comment.  It's also because I have `initializeMongoSqlTranslateLibrary();` in the static initialization block so it also gets set there.

EthanHardyMongo (approved):
Looks really good! 💯

PR #282, opened 2024-08-30
SQL-2293: Differentiate cluster type upon connection

mattChiaravalloti:
This PR updates `MongoConnection` to use a `buildInfo` command instead of a `SELECT 1` query to test the connection, and to determine and store the cluster type based on the `buildInfo` result. It also updates the integration tests to be split into `ADFIntegrationTest` and `DCIntegrationTest` (direct cluster), and adds tests demonstrating community server is disallowed but enterprise server is allowed. I'm still working out any kinks in the evergreen and script updates/additions so bear with me on those!

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -471,11 +471,15 @@ functions:
         ;;
         esac
         
+        # Move library file to where MongoDriver class is located
+        mkdir -p build/classes/java/
+        mv -f src/test/resources/MongoSqlLibraryTest/libmongosqltranslate.so build/classes/java/
+        
         # Download and run local community and enterprise mongods.
         ./resources/start_local_mdb.sh $mdb_version_com $mdb_version_ent $arch
         
         # Run the tests.
-        ./gradlew -Dorg.gradle.java.home=${JAVA_HOME} clean integrationTest \
+        ./gradlew -Dorg.gradle.java.home=${JAVA_HOME} integrationTest \
             -x test --tests DCIntegrationTest > gradle_output.log 2>&1 &
         GRADLE_PID=$!
 
--- a/.evg.yml
+++ b/.evg.yml
@@ -58,7 +58,6 @@ buildvariants:
       - name: "build"
       - name: "test-unit"
       - name: "test-adf-integration"
-      - name: "test-dc-integration"
 
   - name: release
     display_name: "Release"
--- a/.evg.yml
+++ b/.evg.yml
@@ -46,7 +46,8 @@ buildvariants:
       - name: "build"
       - name: "test-unit"
       - name: "test-mongo-sql-translate"
-      - name: "test-integration"
+      - name: "test-adf-integration"
+      - name: "test-dc-integration"
 
   - name: amazon2-arm64-jdk-11
     display_name: Amazon Linux 2 ARM64 jdk-11
@@ -56,7 +57,8 @@ buildvariants:
     tasks:
       - name: "build"
       - name: "test-unit"
-      - name: "test-integration"
+      - name: "test-adf-integration"
+      - name: "test-dc-integration"
 
   - name: release
     display_name: "Release"
@@ -82,9 +84,13 @@ tasks:
     commands:
       - func: "run mongosqltranslate library load test"
 
-  - name: "test-integration"
+  - name: "test-adf-integration"
     commands:
-      - func: "run integration test"
+      - func: "run adf integration test"
+
+  - name: "test-dc-integration"
+    commands:
+      - func: "run dc integration test"
 
   - name: "test-smoke"
     commands:
@@ -377,7 +383,7 @@ functions:
         # Test loading library from same directory as where the JDBC driver is located
         ./gradlew runMongoSQLTranslateLibTest -PtestMethod=testLibraryLoadingFromDriverPath
 
-  "run integration test":
+  "run adf integration test":
     command: shell.exec
     type: test
     params:
@@ -388,7 +394,7 @@ functions:
           ./resources/run_adf.sh start &&
           ./gradlew -Dorg.gradle.java.home=${JAVA_HOME} runDataLoader &&
           ./gradlew -Dorg.gradle.java.home=${JAVA_HOME} clean integrationTest \
-              -x test  > gradle_output.log 2>&1 &
+              -x test --tests ADFIntegrationTest > gradle_output.log 2>&1 &
           GRADLE_PID=$!
 
           echo "Gradle process started with PID $GRADLE_PID"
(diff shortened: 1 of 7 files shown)


.evg.yml
```diff
@@ -388,7 +394,7 @@ functions:
           ./resources/run_adf.sh start &&
           ./gradlew -Dorg.gradle.java.home=${JAVA_HOME} runDataLoader &&
           ./gradlew -Dorg.gradle.java.home=${JAVA_HOME} clean integrationTest \
-              -x test  > gradle_output.log 2>&1 &
+              -x test --tests ADFIntegrationTest > gradle_output.log 2>&1 &
```
mattChiaravalloti, line 397: This filter ensures we only run the `ADFIntegrationTest.java` tests for this function. The other, newly added function, `"run dc integration test"`, uses the filter `--tests DCIntegrationTest`. I thought this was a cleaner solution than adding a whole new package to the codebase and a new gradle task.

src/integration-test/java/com/mongodb/jdbc/integration/ADFIntegrationTest.java
```diff
@@ -392,37 +392,4 @@ public void testDifferentRepresentationsForJavaLegacy() {
         }
         assertEquals(4, uuidValues.size(), "Expected 4 different UUID values (including standard)");
     }
-
-    /** Tests that the driver can work with SRV-style URIs. */
-    @Test
-    public void testConnectWithSRVURI() throws SQLException {
```
mattChiaravalloti, line 398: Moved to the `DCIntegrationTest` file since this is not ADF related.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -524,12 +557,15 @@ public SQLXML createSQLXML() throws SQLException {
...
-                throw new SQLException("Connection error");
+            MongoClusterType actualClusterType = determineClusterType();
+
+            if (actualClusterType == MongoClusterType.Community) {
+                // Community edition is disallowed
+                throw new SQLException("Community edition detected. The JDBC driver is intended for use with MongoDB Enterprise edition or Atlas Data Federation.");
             }
+
+            // Set the cluster type.
+            clusterType = actualClusterType;
```
mattChiaravalloti, line 568: @EthanHardyMongo can you clarify if anything else needs to be done as part of this ticket?

Specifically, am I supposed to check that the library is loaded if we discover an enterprise cluster? I'm not sure what the table in the design doc means when it says "Library Available: N". What does the "N" stand for here? ([link](https://docs.google.com/document/d/1rpKyQ-DsPbWsNOMJsp2cp1dKm99lvzp7_uj2HmOHfFI/edit#bookmark=id.h0s8zd1atopw))

Is there anything else to do here besides check that the library is loaded?

mattChiaravalloti: I updated it to throw an exception if the library is not loaded in enterprise mode. Let me know if there is anything else that's supposed to be added!

src/integration-test/java/com/mongodb/jdbc/integration/DCIntegrationTest.java
```diff
@@ -0,0 +1,108 @@
...
+        String authSource = System.getenv("SRV_TEST_AUTH_DB");
+        assertNotNull(authSource, "SRV_TEST_AUTH_DB variable not set in environment");
+
+        Properties p = new java.util.Properties();
+        p.setProperty("user", user);
+        p.setProperty("password", pwd);
+        p.setProperty("authSource", authSource);
+        p.setProperty("database", "test");
+
+        MongoConnection conn = (MongoConnection) DriverManager.getConnection(fullURI, p);
```
mattChiaravalloti, line 56: This is now failing because the library is not loaded. That seems like a good sign! I guess I should set it up so that these tests are run with the library built and in the appropriate place? Or should we defer that to a follow-up ticket and have this test continue to expect a `SQLException`?

bucaojit: I added a [sample](https://github.com/mongodb/mongo-jdbc-driver/tree/master/src/test/resources/MongoSqlLibraryTest) library that you can [move](https://github.com/mongodb/mongo-jdbc-driver/blob/904969e718ee013df4d493f2e8737eedea58536a/.evg.yml#L374) to the `build/classes/java/` directory that should get this past the library loading error.  If that works successfully then that should be good since we can then test the buildInfo logic here.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -190,6 +197,33 @@ protected void checkConnection() throws SQLException {
...
+        // The { buildInfo: 1 } command returns information that indicates
+        // the type of the cluster.
+        BuildInfo buildInfoRes =
+                mongoClient
+                        .getDatabase("admin")
+                        .withCodecRegistry(MongoDriver.registry)
+                        .runCommand(buildInfoCmd, BuildInfo.class);
+
+        // If the "dataLake" field is present, it must be an ADF cluster.
+        if (buildInfoRes.dataLake != null) {
```
bucaojit, line 213: I was thinking this is something ADF would eventually change this to the correct naming, but it looks like they thought to address it and decided not to fix it so I think we're ok here.  [MHOUSE-6693](https://jira.mongodb.org/browse/MHOUSE-6693)

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -190,6 +197,33 @@ protected void checkConnection() throws SQLException {
...
+        } else if (buildInfoRes.modules != null) {
+            // Otherwise, if "modules" is present and contains "enterprise",
+            // this must be an Enterprise cluster.
+            if (buildInfoRes.modules.contains("enterprise")) {
+                return MongoClusterType.Enterprise;
+            }
+        }
+
+        // Otherwise, this is a Community cluster.
+        return MongoClusterType.Community;
```
bucaojit, line 224: I see that the `buildInfo` command returns ` "ok" : <num>`.  Should we check that "ok" was `1`?  If this is something we should check for, that ok was 1, then maybe we would need an `unknown` cluster type when it can't be determined.  That way we wouldn't be defaulting to Community if we could not in fact determine which type it was.

src/main/java/com/mongodb/jdbc/MongoConnection.java
bucaojit: This looks good to me in order to notify users that there was an issue with loading the library.

bucaojit (changes requested):
Great work!  
I had a few suggestions one related to the library to load in the integration test and checking for the runCommand results.

src/main/java/com/mongodb/jdbc/MongoConnection.java
EthanHardyMongo: @mattChiaravalloti Sorry, in retrospect, the table was not very clear. Natacha updated it to make more sense. But to answer your questions, the `N` was supposed to mean `No`. There's now a `Y` section to make this more clear. And yes, if you are in enterprise mode and the library is not available (this information is discovered before connection time and should be available) then you would throw an error.

EthanHardyMongo:
@bucaojit I couldn't get the `test-dc-integration` task to work on the amazon2 variant. Then, I noticed that we don't run the `test-mongo-sql-translate` task on amazon2, so I tried running that one on amazon2, and it also fails. Also, both tasks fail because the library is not loaded or couldn't be found. Here's a patch showing that: https://spruce.mongodb.com/version/66d9aa85d550c80007be5dc9/tasks?sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC

As of now, I've removed the `test-dc-integration` task from amazon2. I think I'll bring this up in stand up.

.evg.yml
```diff
@@ -438,6 +443,92 @@ functions:
...
+          # Check if Gradle process is still running
+          if ! kill -0 $GRADLE_PID 2>/dev/null; then
+              echo "Gradle process has finished."
+              wait $GRADLE_PID
+              EXITCODE=$?
+              break
+          fi
+    
+          sleep 5
+        done
```
mattChiaravalloti, line 521: I copied this from the adf integration test function, but if we aren't running these on amazon2 we may not need any of this? We should be able to delete it and see no difference on ubuntu I think. Not totally sure, to be honest, but may be worth investigating since this is a big chunk of script.

bucaojit:
@EthanHardyMongo we're okay with running the `test-dc-integration` on the `ubuntu` variant, the current changes in this PR also only run it on ubuntu.  I added the amazon2 variant a little while ago to confirm that `run_adf.sh` works there since mhouse was testing that.  
The library I uploaded only works with the ubuntu variant since that's what it was built on.

.evg.yml
bucaojit: Right, it isn't needed if it's only run on Ubuntu.  It would be fine to remove it.

bucaojit (approved):
LGTM!

EthanHardyMongo (approved):

PR #283, opened 2024-09-09
SQL-2299: Use github token for cloning instead of ssh

bucaojit:
Generates the github token for cloning mongohouse.  The token also has permissions for mongohouse dependencies ie mongoast, sqlproxy, bsonutil.  If `GITHUB_TOKEN` is set `run_adf.sh` will use that token, otherwise revert to the way git clone is done currently (in case user does have the ssh key on local system). 

Here's a [patch](https://spruce.mongodb.com/version/66dc88eda92a3900071e48f5/tasks?page=0&sorts=STATUS%3AASC%3BBASE_STATUS%3ADESC&variant=%5Eubuntu2204-64-jdk-8%24) with the github token change, it was run on a host with the ssh key removed that the devprod team provided to test the changes.
+    run_on: DEVPROD-5742-ubuntu2204-small

When approved I'll add similar changes to the other repos that need it: mongo-odbc-driver, mongo-powerbi-connector

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -84,10 +84,12 @@ tasks:
 
   - name: "test-integration"
     commands:
+      - func: "generate github token"
       - func: "run integration test"
 
   - name: "test-smoke"
     commands:
+      - func: "generate github token"
       - func: "run smoke test"
 
   - name: "publish-maven"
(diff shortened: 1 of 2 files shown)


pmeredit (approved):
LGTM!

nbagnard (approved):
LGTM!

PR #288, opened 2024-10-09
SQL-2380: Download and build jars with translation libraries included

mattChiaravalloti:
This PR updates the evergreen file to download the mongosqltranslate library files into the appropriate places before building the driver. I manually verified that the resources directory is being included in the built jar file. Automated testing of this is blocked on or should be done as part of SQL-2381. Also, I have a few open questions I'll include in comments here.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -47,7 +47,8 @@ buildvariants:
       - name: "test-unit"
       - name: "test-mongo-sql-translate"
       - name: "test-adf-integration"
-      - name: "test-dc-integration"
+      # SQL-2289: integration tests. Use the actual mongosqltranslate library for testing instead of mock.
+      # - name: "test-dc-integration"
 
   - name: amazon2-arm64-jdk-11
     display_name: Amazon Linux 2 ARM64 jdk-11
--- a/.evg.yml
+++ b/.evg.yml
@@ -76,19 +76,20 @@ tasks:
     commands:
       # fetch mongosqltranslate libraries for each platform before building
       # macos arm
+#      - func: "fetch libmongosqltranslate"
+#        vars:
+#          arch: "arm"
+#          platform: "macos"
+#          ext: "dylib"
+#          release_version: "snapshot"
+#      # linux x86
+#      - func: "fetch libmongosqltranslate"
+#        vars:
+#          arch: "x86_64"
+#          platform: "linux"
+#          ext: "so"
+#          release_version: "snapshot"
       - func: "fetch libmongosqltranslate"
-        vars:
-          arch: "arm"
-          platform: "macos"
-          ext: "dylib"
-          release_version: "snapshot"
-      # linux x86
-      - func: "fetch libmongosqltranslate"
-        vars:
-          arch: "x86_64"
-          platform: "linux"
-          ext: "so"
-          release_version: "snapshot"
       - func: "build jdbc driver"
 
   - name: "test-unit"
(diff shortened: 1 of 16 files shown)


.evg.yml
```diff
@@ -74,6 +74,41 @@ buildvariants:
...
     commands:
+      # fetch mongosqltranslate libraries for each platform before building
+      # linux arm
+      - func: "fetch libmongosqltranslate"
+        vars:
+          arch: "arm"
+          platform: "linux"
+          lib_prefix: "lib"
+          ext: "so"
+          release_version: "snapshot"
```
mattChiaravalloti, line 85: This is a placeholder for now for testing purposes. I'm not fully understanding the design doc for this. It says when we want to do a release, we'll use a tag that includes the library version -- `v<major>.<minor>.<patch>-libv<major>.<minor>.<patch>`. Are we supposed to parse the mongnosqltranslate library version from that tag? If so, should we default to "snapshot" for patch builds that do not have the tag value set?

.evg.yml
```diff
@@ -74,6 +74,41 @@ buildvariants:
...
+      # macos arm
+      - func: "fetch libmongosqltranslate"
+        vars:
+          arch: "x86_64"
+          platform: "macos"
+          lib_prefix: "lib"
+          ext: "dylib"
+          release_version: "snapshot"
+      # windows
+      - func: "fetch mongosqltranslate windows"
```
mattChiaravalloti, line 111: I have an open thread on the mongosql-rs PR related to these builds suggesting that the windows library follow the same pattern as the other libraries. This would allow us to reuse the same function here as well.

pmeredit (commented):
I really am not sure. I was looking at the same thing for ODBC. We have a release upload of the libraries but they are completely unversioned suggesting that we would only ever have one version available *and* possibly we could get bit by s3 eventual consistency since it will be overwriting keys which is *not* read after write consistent like new keys are.

.evg.yml
nbagnard: Yes, we should.
Ideally, parsing the tag should be done in "export variables" and on top of MDBJDBC_VER, we will now now have another variable like MDBTRANSLIB_VER which you can use in this task.
I checked and there is no separate tickets, so you want to the work as part of this ticket. It's also the only place where it is used really, so it makes sense.

.evg.yml
nbagnard: I agree. I also chimed in on your comment in the mongosql-rs PR.

nbagnard (changes requested):

.evg.yml
```diff
@@ -74,6 +74,37 @@ buildvariants:
...
+          arch: "arm"
+          platform: "macos"
+          lib_prefix: "lib"
+          ext: "dylib"
+      # macos arm
+      - func: "fetch libmongosqltranslate"
+        vars:
+          arch: "x86_64"
+          platform: "macos"
+          lib_prefix: "lib"
```
mattChiaravalloti, line 104: Just to be clear, when we eventually reuse the `"fetch libmongosqltranslate"` function for Windows, this argument will make more sense. Right now, every invocation of the function provides the same value for this argument. But the Windows library will _not_ start with "lib", so when we refactor to use this function for Windows we'll omit this argument for that invocation.

.evg.yml
mattChiaravalloti: Done!

mattChiaravalloti:
Good news is the `build` task is still working as expected, bad news is all these system failures. Ethan seems to have identified the problem for ODBC and is working on addressing it. I'll follow whatever his solution is for ODBC for this repo. I believe we just need to request some type of permission from IT or something like that.

nbagnard:
>  I believe we just need to request some type of permission from IT or something like that.
Sort of. What we need is a GitHub app token and to add it to the Github section of the project settings. Unfortunately, the info is redacted on the projects mapped to the master branch and Oliver is OOO so we can't ask him for the info today, but if IT didn't provide new ones on Monday, ping Oliver and he should be able to help.

.evg.yml
```diff
@@ -581,9 +632,12 @@ functions:
           # Tag triggered runs are releases and the version is set in the tag.
           # Other runs are snapshot builds (periodic builds or patches)
           if [[ "${triggered_by_git_tag}" != "" ]]; then
-            export MDBJDBC_VER=$(echo ${triggered_by_git_tag} | sed s/v//)
+            # tag should be formatted as 'v<JDBC major>.<minor>.<patch>-libv<libmongosqltranslate major>.<minor>.<patch>
+            export MDBJDBC_VER=$(echo ${triggered_by_git_tag} | sed -e 's/\-libv.*//' | sed s/v// )
+            export LIBMONGOSQLTRANSLATE_VER=$(echo ${triggered_by_git_tag} | sed -e 's/v.*\-libv//')
```
nbagnard, line 637: You'd have less potential for problems using cut to retrieve the part of the tag you want and then remove the prefix. `sed -e` is not super cross-platform friendly, using cut is safer.
```suggestion
            export MDBJDBC_VER=$(echo ${triggered_by_git_tag} | cut -d '-' -f 1 | sed s/v// )
            export LIBMONGOSQLTRANSLATE_VER=$(echo ${triggered_by_git_tag} | cut -d '-' -f 2 | sed s/libv//)
```

nbagnard (changes requested):

.evg.yml
```diff
@@ -74,6 +74,37 @@ buildvariants:
...
+          lib_prefix: "lib"
+          ext: "so"
+      # macos arm
+      - func: "fetch libmongosqltranslate"
+        vars:
+          arch: "arm"
+          platform: "macos"
+          lib_prefix: "lib"
+          ext: "dylib"
+      # macos arm
```
pmeredit, line 99:
```suggestion
      # macos x86
```

nbagnard (approved):
LGTM!

mattChiaravalloti:
When #289 merges, I'll update this branch on top of that and it should then be good to go 👍

pmeredit (approved):
LGTM assuming everything passes after 289 :)

.evg.yml
mattChiaravalloti: Field [SQL-2406](https://jira.mongodb.org/browse/SQL-2406) for this.

PR #289, opened 2024-10-15
SQL-2397: evergreen updates

jchemburkar:
Note all the s3 put changes -- see [DEVPROD-6958](https://jira.mongodb.org/browse/DEVPROD-6958) for context. All of these have been failing for as long as we have evergreen history on this project (eg see [this](https://spruce.mongodb.com/task/mongo_jdbc_driver_amazon2_arm64_jdk_11_build_670aa4089b0a9b0008a950e0_24_10_12_16_30_00/logs?execution=0) run).

[Here](https://spruce.mongodb.com/task/mongo_jdbc_driver_amazon2_arm64_jdk_11_build_patch_85ef3aa7a1eb6cc96f9dda3f1a175ae32ca775c5_670d9414dbe83d00074a371c_24_10_14_21_58_56/logs?execution=0) is a successful build run showing the eap jar with the correct name as an artifact.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -133,9 +133,10 @@ tasks:
   - name: sbom
     commands:
       - func: "generate sbom"
-      - func: "upload sbom"
-      - func: "push SBOM Lite to Silk"
-      - func: "pull augmented SBOM from Silk"
+      # TODO: uncomment this before merging branch back into master, to ensure we still upload to silk from the main branch
+      # - func: "upload sbom"
+      # - func: "push SBOM Lite to Silk"
+      # - func: "pull augmented SBOM from Silk"
       - func: "scan sbom"
 
   - name: ssdlc-artifacts-release
--- a/.evg.yml
+++ b/.evg.yml
@@ -7,11 +7,11 @@ variables:
     params:
       role_arn: ${assume_role_arn}
       duration_seconds: 3600
-  - &aws_std_config
+  - &evg_bucket_config
     aws_key: ${AWS_ACCESS_KEY_ID}
     aws_secret: ${AWS_SECRET_ACCESS_KEY}
     aws_session_token: ${AWS_SESSION_TOKEN}
-    bucket: mongo-jdbc-driver-eap
+    bucket: evg-bucket-mongo-jdbc-driver
 
 pre:
   - func: "fetch source"
@@ -152,6 +152,8 @@ tasks:
       - func: "scan sbom"
 
   - name: ssdlc-artifacts-release
+    # TODO: remove this before merging eap branch back into main branch, to allow download center to update.
+    disable: true
     run_on: ubuntu2204-small
     git_tag_only: true
     depends_on:
@@ -233,7 +235,7 @@ functions:
           ls -lrt $SSDLC_DIR
     - command: s3.put
       params:
-        <<: *aws_std_config
+        <<: *evg_bucket_config
         local_file: mongo-jdbc-driver/artifacts/ssdlc/${AUGMENTED_SBOM_NAME}
         remote_file: artifacts/${version_id}/ssdlc/${AUGMENTED_SBOM_NAME}
         content_type: application/json
@@ -243,7 +245,7 @@ functions:
     - *assume_role_cmd
     - command: s3.get
       params:
-        <<: *aws_std_config
+        <<: *evg_bucket_config
         local_file: artifacts/ssdlc/mongodb-jdbc-${MDBJDBC_VER}.sbom.json
         remote_file: artifacts/${version_id}/ssdlc/${AUGMENTED_SBOM_NAME}
         content_type: application/json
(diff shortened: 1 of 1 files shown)


.evg.yml
```diff
@@ -662,14 +667,19 @@ functions:
...
-        aws_secret: ${aws_secret}
-        local_file: mongo-jdbc-driver/build/libs/mongodb-jdbc-${MDBJDBC_VER}.jar
-        remote_file: ${S3_ARTIFACTS_DIR}/mongodb-jdbc-${MDBJDBC_VER}.jar
-        bucket: mciuploads
+        aws_key: ${AWS_ACCESS_KEY_ID}
+        aws_secret: ${AWS_SECRET_ACCESS_KEY}
+        aws_session_token: ${AWS_SESSION_TOKEN}
+        local_file: mongo-jdbc-driver/build/libs/mongodb-jdbc-eap-${MDBJDBC_VER}.jar
+        remote_file: ${S3_ARTIFACTS_DIR}/mongodb-jdbc-eap-${MDBJDBC_VER}.jar
+        bucket: mongo-jdbc-driver-eap
```
mattChiaravalloti, line 682: We repeat this same structure many times now, so let's use a YAML anchor and aliases to reduce it. I think something like this should work:
```
# towards the top of the file
variables:
  - &assume_role_cmd
    command: ec2.assume_role
    params:
      role_arn: ${assume_role_arn}
      duration_seconds: 3600
  - &aws_std_config
    aws_key: ${AWS_ACCESS_KEY_ID}
    aws_secret: ${AWS_SECRET_ACCESS_KEY}
    aws_session_token: ${AWS_SESSION_TOKEN}
    bucket: evg-bucket-mongo-jdbc-driver

...

# In each of these functions...
  "fetch jdbc files":
    - *assume_role_cmd
    - command: s3.get
      params:
        <<: *aws_std_config
        working_dir: mongo-jdbc-driver
        local_file: ...
        remote_file: ...

  "upload unit test classes":
    - *assume_role_cmd
    - command: s3.put
      params:
        <<: *aws_std_config
        ...
        
...
```

Admittedly, I did not test exactly this code but the rough idea should work, potentially with some minor tweaks. This should really help us not only cut down on the size of this config file but also avoid issues in the future where we have to make changes to each of these affected functions in bulk.

mattChiaravalloti (changes requested):
Looking good so far. Thank you for sorting out how to address some of these AWS open questions. Looks like we may still have some access denied failures for semgrep? Regardless, I think now that we know the pattern, we should extract it to a reusable YAML variable 👍

.evg.yml
jchemburkar: Good idea, I'll experiment and try to get this working. Also glad you saw the bucket update, this looks like the missing piece we were talking about yesterday :)

mattChiaravalloti: Based on this part of the evergreen docs: https://docs.devprod.prod.corp.mongodb.com/evergreen/Project-Configuration/Project-Configuration-Files#functions ! Could be a useful reference 👍

jchemburkar: definitely, thanks for linking. I'll do the same in the parallel change to odbc-eap

.evg.yml
```diff
@@ -1,6 +1,18 @@
 stepback: true
 command_type: system
 
+variables:
+  - &assume_role_cmd
+    command: ec2.assume_role
+    params:
+      role_arn: ${assume_role_arn}
+      duration_seconds: 3600
+  - &aws_std_config
```
nbagnard, line 10: If we want to reuse this in our main branch when we merge, we should use a more meaningful name.
Maybe something like `mongo-jdbc-driver-eap_bucket_config` or `evg_bucket_config`? 
I think that having the bucket name as part of the variable name makes it easier to know where we are writing to. Our release bucket or the project specific bucket for example.

Is this bucket a release bucket or a "temp" evg bucket where we put the result of tasks that we want to capture but will only keep for about 60 days (I think that's how much time the data is kept, I might be off)?
Right now, it seems like you expect it to be a temp bucket.

.evg.yml
```diff
@@ -129,9 +145,10 @@ tasks:
...
       - func: "generate sbom"
+      # TODO: uncomment this before merging branch back into master, to ensure we still upload to silk from the main branch
       - func: "upload sbom"
-      - func: "push SBOM Lite to Silk"
-      - func: "pull augmented SBOM from Silk"
+      # - func: "push SBOM Lite to Silk"
+      # - func: "pull augmented SBOM from Silk"
       - func: "scan sbom"
 
   - name: ssdlc-artifacts-release
```
nbagnard, line 154: I think you can get rid of that task completely.
The compliance report links to the augmented SBOM and static code analysis report which we won't have.

nbagnard (approved):
I have only very small requests/comments.
No need to retag me :)

.evg.yml
jchemburkar: I was thinking it would be the temp bucket, since the eap branch doesn't have access to `evg-mongo-jdbc-driver` nor `mciuploads`, and given we dont have any ticketing for updating the release/publishing of the eap version.

nbagnard: This ticket is for publishing the EAP version to S3 because the publishing is done via Maven for the master branch, so I am not sure what Nathan had in mind. We'll ask him when he is back.
I wonder if we are Ok "loosing" the releases after a while or not 🤔 
I'd be happy to keep them around if possible, but if the delay is indeed 60 days, then it should be more than enough.

PR #291, opened 2024-10-29
SQL-2406: Update fetch libmongosqltranslate windows to reuse generic fetch function in JDBC

EthanHardyMongo:


.evg.yml
```diff
@@ -438,16 +453,6 @@ functions:
...
-  "fetch mongosqltranslate windows":
-    - command: s3.get
-      params:
-        working_dir: mongo-jdbc-driver
-        aws_key: ${release_aws_key}
-        aws_secret: ${release_aws_secret}
-        local_file: mongo-jdbc-driver/src/main/resources/x86_64/windows/mongosqltranslate.dll
-        remote_file: mongosqltranslate/mongosqltranslate-v${LIBMONGOSQLTRANSLATE_VER}-win.dll
-        bucket: translators-connectors-releases
-
```
EthanHardyMongo, line 450: Since `fetch libmongosqltranslate` uses `platform` for the local file path and in the name of the remote file, the library will now be put in `mongo-jdbc-driver/src/main/resources/x86_64/win/` instead of `mongo-jdbc-driver/src/main/resources/x86_64/windows/`. Will this be a problem?

src/main/java/com/mongodb/jdbc/utils/NativeLoader.java
```diff
@@ -61,7 +61,7 @@ public class NativeLoader {
 
     private static final String MACOS = "macos";
     private static final String LINUX = "linux";
-    private static final String WINDOWS = "windows";
+    private static final String WINDOWS = "win";
```
EthanHardyMongo, line 64: In `fetch libmongosqltranslate`, `platform` is used both in the path and in the filename. Additionally, the `platform` in the filename is `win`, so the evergreen var has to be `win`, so for consistency, I updated the folder name and this variable to `win`.

mattChiaravalloti (approved):
🙏

PR #295, opened 2024-10-31
SQL-2433: Update logic to extract versions

nbagnard:


.evg.yml
```diff
@@ -729,29 +729,33 @@ functions:
           # Other runs are snapshot builds (periodic builds or patches)
           if [[ "${triggered_by_git_tag}" != "" ]]; then
             # tag should be formatted as 'v<JDBC major>.<minor>.<patch>-libv<libmongosqltranslate major>.<minor>.<patch>
-            export MDBJDBC_VER=$(echo ${triggered_by_git_tag} | cut -d '-' -f 1 | sed s/v// )
-            export LIBMONGOSQLTRANSLATE_VER=$(echo ${triggered_by_git_tag} | cut -d '-' -f 2 | sed s/libv//)
+            export MDBJDBC_VER=$(echo ${triggered_by_git_tag} | awk -F'-libv' '{print $1}' | sed s/v// )
+            export LIBMONGOSQLTRANSLATE_VER=$(echo "$input" | awk -F'-libv' '{print $2}')
```
bucaojit, line 733:
```suggestion
            export LIBMONGOSQLTRANSLATE_VER=$(echo ${triggered_by_git_tag}  | awk -F'-libv' '{print $2}')
```
Is this the right variable to use instead

terakilobyte, line 733: Is `$input` defined here?

```suggestion
            export LIBMONGOSQLTRANSLATE_VER=$(echo ${triggered_by_git_tag} | awk -F'-libv' '{print $2}')
```

terakilobyte (approved):
LGTM. I just had one question and a suggestion if the answer is "no".

PR #298, opened 2024-11-05
SQL-2437:  Set release version

nbagnard:
If the property "isTagTriggered" is set, then:
- The release version (jar names) is the abbreviated git version. For example "3.0.0-alpha-libv1.0.0-alpha".
- The driver version (reported by the driver) is also the abbreviated git version.
If the property "isTagTriggered" is not set, then:
- The release version (jar names) is the abbreviated git version followed by `-SNAPSHOT`. For example "3.0.0-alpha-libv1.0.0-alpha-SNAPSHOT".
- The driver version (reported by the driver) is the full git version. For example "3.0.0-alpha-libv1.0.0-alpha-2-gbe21c46-dirty".

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -8,7 +8,7 @@ plugins {
 }
 
 if (project.hasProperty('isTagTriggered')) {
-    version = getAbbreviatedGitVersion()()
+    version = getAbbreviatedGitVersion()
 } else {
     version = getGitVersion()
 }
--- a/build.gradle
+++ b/build.gradle
@@ -14,8 +14,9 @@ if (project.hasProperty('isTagTriggered')) {
 }
 
 ext {
+    println("Driver version = " + version)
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -7,10 +7,15 @@ plugins {
...
-version =  getGitVersion()
+if (project.hasProperty('isTagTriggered')) {
+    version = getAbbreviatedGitVersion()
+} else {
+    version = getGitVersion()
+}
 
 ext {
     releaseVersion = getReleaseVersion()
+    println(releaseVersion)
```
jchemburkar, line 18: just double checking but is this testing remains or do we want this as part of hte log?

jchemburkar (approved):
looks good to me!

terakilobyte (approved):
LGTM. I think the println will help debug any future issues.

PR #300, opened 2024-11-05
SQL-2423: correctly report db version

terakilobyte:
We now correctly report the major and minor version. I've also updated it to work a bit better with adf version info, and report both adf version and mongosql version. Sample log output of internals:

```sh
[2024-11-06 13:19:23.453] [INFO] [c-2] com.mongodb.jdbc.MongoConnection initConnectionLogger: Connecting using MongoDB Atlas SQL interface JDBC Driver 3.0.0-alpha-libv1.0.0-alpha-11-g56f9799-dirty
[2024-11-06 13:19:25.355] [FINE] [c-2] com.mongodb.jdbc.MongoConnection determineClusterType: BuildInfo{version='7.2.0.v20241030.1.5.1', modules=null, ok=1, dataLake=DataLake{version='v20241030', mongoSQLVersion=1.5.1}}
[2024-11-06 13:19:25.356] [INFO] [c-2] com.mongodb.jdbc.MongoConnection$ConnValidation call: Connecting to cluster type AtlasDataFederation with server version 7.2.0.v20241030.1.5.1
[2024-11-06 13:23:11.157] [INFO] [c-1] com.mongodb.jdbc.MongoConnection initConnectionLogger: Connecting using MongoDB Atlas SQL interface JDBC Driver 3.0.0-alpha-libv1.0.0-alpha-11-g56f9799-dirty
[2024-11-06 13:23:12.993] [FINE] [c-1] com.mongodb.jdbc.MongoConnection determineClusterType: BuildInfo{fullVersion='7.2.0.v20241030.1.5.1', versionArray=[7, 2, 0, 0], majorVersion=7, minorVersion=2, modules=null, ok=1, dataLake=DataLake{version='v20241030', mongoSQLVersion=1.5.1}}
[2024-11-06 13:23:12.994] [INFO] [c-1] com.mongodb.jdbc.MongoConnection$ConnValidation call: Connecting to cluster type AtlasDataFederation with server version 7.2.0.v20241030.1.5.1
[2024-11-06 13:23:40.161] [INFO] [c-2] com.mongodb.jdbc.MongoConnection initConnectionLogger: Connecting using MongoDB Atlas SQL interface JDBC Driver 3.0.0-alpha-libv1.0.0-alpha-11-g56f9799-dirty
[2024-11-06 13:23:41.827] [FINE] [c-2] com.mongodb.jdbc.MongoConnection determineClusterType: BuildInfo{fullVersion='8.0.3', versionArray=[8, 0, 3, 0], majorVersion=8, minorVersion=0, modules=[enterprise], ok=1, dataLake=null}
[2024-11-06 13:23:41.828] [INFO] [c-2] com.mongodb.jdbc.MongoConnection$ConnValidation call: Connecting to cluster type Enterprise with server version 8.0.3
[2024-11-06 13:23:41.834] [INFO] [c-2] com.mongodb.jdbc.mongosql.MongoSQLTranslate getMongosqlTranslateVersion: mongosqlTranslateVersion: v1.0.0
[2024-11-06 13:23:41.835] [INFO] [c-2] com.mongodb.jdbc.mongosql.MongoSQLTranslate checkDriverVersion: Driver Compatibility Status: true
```

### DIFF
--- a/src/main/java/com/mongodb/jdbc/BuildInfo.java
+++ b/src/main/java/com/mongodb/jdbc/BuildInfo.java
@@ -16,48 +16,62 @@
 
 package com.mongodb.jdbc;
 
-import java.util.Optional;
+import java.util.List;
 import java.util.Set;
 import org.bson.codecs.pojo.annotations.BsonCreator;
 import org.bson.codecs.pojo.annotations.BsonProperty;
 
 public class BuildInfo {
-    public String version;
+    private String fullVersion;
+    private List<Integer> versionArray;
     public Set<String> modules;
     public int ok;
+
     public DataLake dataLake;
 
     @BsonCreator
     public BuildInfo(
             @BsonProperty("version") String version,
+            @BsonProperty("versionArray") List<Integer> versionArray,
             @BsonProperty("modules") Set<String> modules,
             @BsonProperty("ok") int ok,
-            @BsonProperty("dataLake") DataLake dataLake) {
-        this.version = version;
-        this.modules = modules;
-        this.ok = ok;
+            @BsonProperty("dataLake") DataLake dataLake)
+            throws IndexOutOfBoundsException {
+        this.fullVersion = version;
+        this.versionArray = versionArray;
+        if (dataLake != null) {
+            this.fullVersion += "." + dataLake.version + "." + dataLake.mongoSQLVersion;
+        }
         this.dataLake = dataLake;
+        this.ok = ok;
+        this.modules = modules;
     }
 
-    public Optional<DataLake> getDataLakePresence() {
-        return Optional.ofNullable(dataLake);
+    public String getFullVersion() {
+        return this.fullVersion;
     }
 
-    public String getDataLakeVersion() {
-        return getDataLakePresence().map(dl -> "." + dl.version).orElse("");
+    public int getMajorVersion() throws IndexOutOfBoundsException {
+        return this.versionArray.get(0);
     }
 
-    public String getDataLakeMongoSQLVersion() {
-        return getDataLakePresence().map(dl -> "." + dl.mongoSQLVersion).orElse("");
+    public int getMinorVersion() throws IndexOutOfBoundsException {
+        return this.versionArray.get(1);
     }
 
     // Override toString for logging
     @Override
     public String toString() {
         return "BuildInfo{"
-                + "version='"
-                + version
+                + "fullVersion='"
+                + fullVersion
                 + '\''
+                + ", versionArray="
+                + versionArray
+                + ", majorVersion="
+                + this.getMajorVersion()
+                + ", minorVersion="
+                + this.getMinorVersion()
                 + ", modules="
                 + modules
                 + ", ok="
--- a/src/main/java/com/mongodb/jdbc/BuildInfo.java
+++ b/src/main/java/com/mongodb/jdbc/BuildInfo.java
@@ -16,10 +16,10 @@
 
 package com.mongodb.jdbc;
 
+import java.util.Optional;
+import java.util.Set;
 import org.bson.codecs.pojo.annotations.BsonCreator;
 import org.bson.codecs.pojo.annotations.BsonProperty;
-import java.util.Set;
-import java.util.Optional;
 
 public class BuildInfo {
     public String version;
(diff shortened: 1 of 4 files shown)


src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -581,6 +585,9 @@ public String getDatabaseProductVersion() throws SQLException {
             return serverVersion;
         }
         serverVersion = conn.getServerVersion();
+        String[] versionParts = serverVersion.split("\\.");
+        serverMajorVersion = parseInt(versionParts[0]);
```
pmeredit, line 589: If you don't check the length, this could throw an exception. Not sure how much we care.

pmeredit (approved):
LGTM other than that it may throw exceptions

terakilobyte:
I've updated this to report ADF version and mongosql version. I'll update the PR description.

src/main/java/com/mongodb/jdbc/BuildInfo.java
```diff
@@ -16,13 +16,54 @@
...
 
-// Simple POJO for deserializing buildInfo results.
 public class BuildInfo {
     public String version;
     public Set<String> modules;
-    public BsonValue dataLake;
     public int ok;
+    public DataLake dataLake;
+
+    @BsonCreator
```
nbagnard, line 30: You don't need to use the annotations when there are no customization, it's a simple POJO.
See JsonSchema or MongoJsonSchema. And the flower example https://www.mongodb.com/docs/drivers/java/sync/v5.2/fundamentals/data-formats/document-data-format-pojo/

Now, if you want, you could bake in the logic for the "calculated" version in your constructor.
If there is a dataLake section, then add the datalake version and mongosql version. I would even make this another variable. Like fullVersion and use that directly in MongoConnection.
It will be cleaner/simpler.

nbagnard (changes requested):
My only concern is with catching all exceptions instead of just NumberFormatException.
The other comments are more around how to use POJOs and codec in Java, and another possible approach for where to process the info.

src/main/java/com/mongodb/jdbc/BuildInfo.java
terakilobyte: Yep going with this approach

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -228,8 +216,20 @@ private MongoClusterType determineClusterType() {
...
+        logger.log(Level.FINE, buildInfoRes.toString());
+
+        this.serverVersion = buildInfoRes.getFullVersion();
+
+        try {
+            this.serverMajorVersion = buildInfoRes.getMajorVersion();
+            this.serverMinorVersion = buildInfoRes.getMinorVersion();
+        } catch (IndexOutOfBoundsException e) {
+            logger.log(Level.SEVERE, e.toString());
+        }
```
nbagnard, line 228: If versionArray can be null, you are not protected against Null Pointer Exception. 
Maybe use a catch all here.

```suggestion
        } catch (Exception e) {
            // Only log issues happening while trying to compute the server version as this is not a blocker.
            logger.log(Level.SEVERE, e.toString());
        }
```

nbagnard (approved):
LGTM!

src/main/java/com/mongodb/jdbc/MongoConnection.java
terakilobyte: I'll add the comment. I can check for null pointer exception, but the versionArray has been returned as part of buildinfo since [4.4](https://www.mongodb.com/docs/v4.4/reference/command/buildInfo/). I stuck with the Index exception in case the return is malformed, though that seems like it'd be an exceedingly rare occurence.

nbagnard: The indexOutOfBound is necessary because you throw. I just think that not being too specific in this case is good, but I agree we should never get there. It's up to you.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -228,8 +217,21 @@ private MongoClusterType determineClusterType() {
...
 
+        logger.log(Level.FINE, buildInfoRes.toString());
+
+        this.serverVersion = buildInfoRes.getFullVersion();
+
+        try {
+            this.serverMajorVersion = buildInfoRes.getMajorVersion();
+            this.serverMinorVersion = buildInfoRes.getMinorVersion();
+            // Only log issues happening while trying to compute the server version as this is not a blocker.
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
nbagnard: I'd be happy to have the stack trace. We can have a generic error message and then the stack trace

PR #301, opened 2024-11-11
SQL-2460: Add handling for non-integer indexes

EthanHardyMongo:
This PR filters out non-integer indexes. At first, I tried to set `ASC_OR_DESC` to null for non-integer indexes as [the spec](https://download.oracle.com/otn_hosted_doc/jdeveloper/905/jdbc-javadoc/oracle/jdbc/OracleDatabaseMetaData.html#getIndexInfo(java.lang.String,%20java.lang.String,%20java.lang.String,%20boolean,%20boolean)) says, "`ASC_OR_DESC String => column sort sequence, "A" => ascending, "D" => descending, may be null if sort sequence is not supported`". However, I then got the following DBeaver error: `Error reading table indexes: SQL Error: java.lang.IllegalArgumentException: The value for key ASC_OR_DESC can not be null`. So in the end, I filtered out all the non-integer indexes which works fine.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
@@ -2662,18 +2662,22 @@ public class MongoDatabaseMetaData implements DatabaseMetaData {
 
         return keys.keySet()
                 .stream()
+                .filter(
+                        key -> {
+                            // If the index is not an integer (e.g., a geospatial index), `keys.getInteger(key)`
+                            // will throw a ClassCastException. In this case, we skip the index because the
+                            // sort sequence is not supported by JDBC.
+                            try {
+                                keys.getInteger(key);
+                            } catch (ClassCastException e) {
+                                return false;
+                            }
+                            return true;
+                        })
                 .map(
                         key -> {
-
-                            // If the index is not an integer (e.g., a geospatial index), `keys.getInteger(key)`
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/jdbc/MongoDatabaseMetaData.java
```diff
@@ -223,7 +223,7 @@ public class MongoDatabaseMetaData implements DatabaseMetaData {
     private static final List<SortableBsonDocument.SortSpec> GET_INDEX_INFO_SORT_SPECS =
             Arrays.asList(
                     new SortableBsonDocument.SortSpec(
-                            NON_UNIQUE, SortableBsonDocument.ValueType.String),
+                            NON_UNIQUE, SortableBsonDocument.ValueType.Boolean),
                     new SortableBsonDocument.SortSpec(
                             INDEX_NAME, SortableBsonDocument.ValueType.String),
                     new SortableBsonDocument.SortSpec(
```
EthanHardyMongo, line 229: Once I fixed the first issue, I got the following error: `Error reading table indexes: SQL Error: org.bson.BsonInvalidOperationException: Value expected to be of type STRING is of unexpected type BOOLEAN`. This was due to `NON_UNIQUE` holding a Boolean rather than a String like we said it would in the sort specs, so I updated the sort specs and `SortableBsonDocument.java` to handle booleans.

mattChiaravalloti (approved):

pmeredit (approved):
LGTM!

PR #302, opened 2024-11-11
SQL-2461: Add getCursorName

pmeredit:


bucaojit (approved):
The changes look good to me.  

One question for @EthanHardyMongo are the changes for this supposed to be in `master` or `on-prem-eap`?

EthanHardyMongo (approved):
LGTM! I think we should put this in the on-prem branch instead?

PR #303, opened 2024-11-12
SQL-2436: Split variables intialization for JDBC

bucaojit:
Split the `export variables` into two commands.  The first command has `silent: false` sets the variables that are generated and then echo's them.  The second command will get those variables as expansions and exports the rest of the variables.  

Let me know if there are more (or fewer) variables that you'd like added to the first command. 

Example of the variables being [logged](https://evergreen.mongodb.com/task_log_raw/mongo_jdbc_driver_eap_release_build_patch_6a76451cd6b0675fb82a87b301e19e70f3e62f45_6733916f2bc09500079c8023_24_11_12_17_33_38/0?type=T#L20).

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -793,11 +793,31 @@ functions:
               fi
           fi
           
+          # set the state needed irrespective of _platform
+          SBOM_WITHOUT_TEAM_NAME=$ARTIFACTS_DIR/ssdlc/sbom_without_team_name.json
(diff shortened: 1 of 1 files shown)


jchemburkar (approved):
lgtm!

nbagnard (commented):
Could we move the SSDLC variables up as well?

nbagnard (approved):
LGTM!

PR #308, opened 2024-11-26
SQL-2437: Don't rebuild the driver from Smoketest

nbagnard:


.evg.yml
```diff
@@ -21,18 +21,22 @@ pre:
   - func: "export variables"
 
 post:
-  - func: "upload jar file"
```
jchemburkar, line 24: Just to be clear, this is the critical fix for the ticket, correct? that we aren't overwriting the jar file with a new one used for smoke testing?

jchemburkar (approved):
lgtm pending my q!

bucaojit (approved):
LGTM

.evg.yml
nbagnard: The critical part is that we are not overwriting it, correct. The other post functions are for copying the results of the smoke test. Which can be critical when it's failing, but it's not related to why the release version was off.

As for why it was all funky, the line you have above it part of the problem. Because our S3 task will put the same artifacts at the same path if they are running on the same variant.
`./gradlew clean shadowJar &&` in run smoke test was another piece of the puzzle. It was rebuilding the driver.

PR #309, opened 2024-11-27
SQL-2469: Support X509 auth in driver using properties and env var

bucaojit:
This adds the option to add the x509 PEM file path to the property or via the `X509_CLIENT_CERT_PATH` environment variable.  The property takes precedence as we expect to have this be modifiable in the Tableau connection dialogue. 

Tested using an Atlas cluster and local enterprise mongod. 
Used the self-signed certificates and the Atlas generated certs. 

I plan to add some unit tests but wanted to get the code up for review.


.evg.yml
```diff
@@ -180,6 +181,47 @@ tasks:
...
+          lib_prefix: "lib"
+          ext: "dylib"
+      # windows
+      - func: "fetch libmongosqltranslate"
+        vars:
+          arch: "x86_64"
+          platform: "win"
+          lib_prefix: ""
+          ext: "dll"
+      - func: "run dc integration test"
```
bucaojit, line 222: For some reason it fails without running the `run dc integration test` first.  I suspect it's environment variable or directory related.  I'll see what the problem is and remove this once that's fixed.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -504,8 +527,19 @@ public static Pair<ConnectionString, DriverPropertyInfo[]> getConnectionSettings
...
 
         ParseResult result = normalizeConnectionOptions(originalConnectionString, info);
-        String user = result.user;
-        char[] password = result.password;
+        String user = null;
+
+        char[] password = null;
+        char[] x509Passphrase = null;
+
+        if (MONGODB_X509.equalsIgnoreCase(result.authMechanism)) {
```
nbagnard, line 535: [note]
Since I played with this recently, I noticed that we can call result.getAuthenticationMechanism and use the Enum AuthenticationMechanism. Not relying on string values is a bit better. More built-in safety and simplified check.
If you feel like it, feel free to make the change on this PR. If not, that's alright too,

src/main/java/com/mongodb/jdbc/utils/X509Authentication.java
```diff
@@ -0,0 +1,144 @@
...
+            Object pemObj;
+
+            // Iterate through PEM objects found in the PEM file and process them based on type:
+            //  - Encrypted/unencrypted private keys
+            //  - X.509 certificates
+            while ((pemObj = pemParser.readObject()) != null) {
+                try {
+                    if (passphrase != null
+                            && passphrase.length > 0
+                            && pemObj instanceof PKCS8EncryptedPrivateKeyInfo) {
```
nbagnard, line 85: Does this mean that only PKCS8 is supported? Not PKCS12?

nbagnard (commented):
This is looking pretty good already! I'll finish the review tomorrow, so that you have it all on Monday ;)

src/integration-test/java/com/mongodb/jdbc/integration/AuthX509IntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+import java.util.Properties;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+public class AuthX509IntegrationTest {
+
+    private static final String LOCAL_PORT_ENV_VAR = "LOCAL_MDB_PORT_ENT";
+    private static final String PASSWORD_ENV_VAR = "ADF_TEST_LOCAL_PWD";
+    private static final String X509_CERT_PATH_PROPERTY = "x509PemPath";
```
terakilobyte, line 33: We're seeing a massive proliferation of environmental variables that are dependent on evergreen or setting them locally. I'd like to start seeing accompanying shell files that encapsulate all of them that could be sourced both locally and in evergreen to make things more portable.

resources/start_local_mdb.sh
```diff
@@ -52,9 +58,55 @@ start_mdb_and_create_user() {
...
+  $mongod_dir/bin/mongod \
+    --dbpath $db_path \
+    --port $port \
+    --tlsMode requireTLS \
+    --tlsCertificateKeyFile "$RESOURCES_DIR/authentication_test/X509/server.pem" \
+    --tlsCAFile "$RESOURCES_DIR/authentication_test/X509/ca.crt" \
+    --bind_ip localhost &
+
+  echo "MongoDB started with X.509 authentication enabled on port $port"
+}
```
terakilobyte, line 99: I asked this question in slack, but should we start capturing these configurations (and data, like users and information) in config files?

terakilobyte (commented):
Just a few thoughts I had while going over it. A lot of great work here!

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -531,17 +565,19 @@ public static Pair<ConnectionString, DriverPropertyInfo[]> getConnectionSettings
         }
         if (password == null
                 && user != null
+                && !MONGODB_X509.equalsIgnoreCase(result.authMechanism)
                 && !MONGODB_OIDC.equalsIgnoreCase(result.authMechanism)) {
             // password is null, but user is not, we must prompt for the password.
             mandatoryConnectionProperties.add(new DriverPropertyInfo(PASSWORD, null));
```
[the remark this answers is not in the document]
bucaojit: Right, it doesn't have to set the username explicitly.  It obtains it from the PEM file.  
We can see it with the openssl command:
```
openssl x509 -in client-unencrypted.pem -subject -noout

subject=C=US, ST=NY, L=NY, O=mongodb, OU=eng
```

That reminds me I was going to bring up.  In the Tableau Connector connection dialogue.  Our idea was to use the `Password` for the passphrase.  Which works fine, one thing that isn't ideal is the user needs to put something in the username field.  So I think if we wanted to we can update the `certificate` drop down to have a single password field in the Tableau dialogue to clean things up.  It works as is now however.

gradle.properties
```diff
@@ -4,6 +4,7 @@ mongodbDriverVersion = 5.2.+
 junitJupiterVersion = 5.5.+
 mockitoVersion = 3.0.+
 googleLintVersion = 1.+
+bouncyCastleVersion = 1.78.+
```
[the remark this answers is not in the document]
bucaojit: It works, I updated it to 1.79 thanks for catching that.

resources/start_local_mdb.sh
```diff
@@ -1,10 +1,16 @@
 #!/bin/bash
 #
-# Usage start_local_mdb.sh <community version> <enterprise version> <architecture>
+# Usage start_local_mdb.sh <community version> <enterprise version> <architecture> [--x509]
+
+# Get the directory where the script is located
+SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+RESOURCES_DIR="$SCRIPT_DIR"
```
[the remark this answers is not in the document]
bucaojit: Sounds good, I made this change.

src/integration-test/java/com/mongodb/jdbc/integration/AuthX509IntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Tests that when x509PemPath property is set, it takes precedence over the
+     * X509_CLIENT_CERT_PATH environment variable. The X509_CLIENT_CERT_PATH points to a valid file,
+     * but the property points to an invalid path. Should fail with FileNotFound.
+     */
+    @Test
+    public void testPropertyPrecedenceFailsIfWrong() {
+        String certPathEnvVar = System.getenv(X509_CLIENT_CERT_PATH);
```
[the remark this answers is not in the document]
bucaojit: Good idea, I added the check to see if the file exists.

src/integration-test/java/com/mongodb/jdbc/integration/AuthX509IntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+import static com.mongodb.jdbc.MongoConnection.X509_CLIENT_CERT_PATH;
+import static org.junit.jupiter.api.Assertions.*;
+
+import java.io.FileNotFoundException;
+import java.sql.*;
+import java.util.Properties;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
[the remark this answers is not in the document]
bucaojit: Oh, you're right it's not needed I've removed it.  I got it from copying another test :-).

src/integration-test/java/com/mongodb/jdbc/integration/AuthX509IntegrationTest.java
```diff
@@ -0,0 +1,141 @@
...
+        assertNotNull(cause, "Expected a cause in the exception");
+        assertTrue(
+                cause instanceof FileNotFoundException,
+                "Expected FileNotFoundException, but got: " + cause.getClass().getName());
+    }
+
+    /** Tests that PEM file without passphrase connects */
+    @Test
+    public void testPropertySetCorrectly() throws SQLException {
+        Connection connection =
```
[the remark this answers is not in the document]
bucaojit: Sure, I put the connections in the try-with-resources block.

src/main/java/com/mongodb/jdbc/utils/X509Authentication.java
bucaojit: Yes, the PEM file format doesn't support PKCS12 I would need to add logic to support reading the PKCS12 file since that's a binary `*.p12` or `*.pks` file.  I focused more on PEM since that's what we get from the Atlas UI and looking up the self-signed [documentation](https://www.mongodb.com/docs/manual/tutorial/configure-x509-client-authentication/) I see PEM files used.  
I can create a ticket to support pkcs12 if that's what we want.  But even with regard to PEM, that's me speculating on what the customer will want to use.

src/integration-test/java/com/mongodb/jdbc/integration/AuthX509IntegrationTest.java
bucaojit: I think for this one I'll create a ticket and we can think of how to improve the environment variable experience.

resources/start_local_mdb.sh
bucaojit: That's a good idea, this is an area we can improve also.  We can chat about it during our on-site brainstorming.  I'll create a ticket.

src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -255,6 +258,18 @@ public Connection connect(String url, Properties info) throws SQLException {
         return conn;
     }
 
+    public static class ConnectionSettings {
```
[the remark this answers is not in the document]
bucaojit: Changed it to `MongoConnectionConfig`

src/main/java/com/mongodb/jdbc/MongoDriver.java
bucaojit: That works, I made the change.

src/main/java/com/mongodb/jdbc/MongoConnection.java
```diff
@@ -71,6 +73,8 @@ public class MongoConnection implements Connection {
     private int serverMinorVersion;
     private String serverVersion;
 
+    public static final String X509_CLIENT_CERT_PATH = "X509_CLIENT_CERT_PATH";
```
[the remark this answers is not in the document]
bucaojit: I made the change 👍

terakilobyte (approved):
LGTM, nice work.

src/main/java/com/mongodb/jdbc/MongoDriver.java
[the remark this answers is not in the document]
nbagnard: @bucaojit  Can you create a follow up ticket for that one. Especially after the conversation with Tableau today, it seems the best course of action. WDYT?

src/main/java/com/mongodb/jdbc/MongoDriver.java
nbagnard: Thank you :)

src/main/java/com/mongodb/jdbc/utils/X509Authentication.java
nbagnard: Sounds good to me. Especially since this is what users can download on Atlas.
We might need to make the doc very clear about this, that's all 👍

nbagnard (approved):
LGMT! Awesome work! 🚀

src/main/java/com/mongodb/jdbc/utils/X509Authentication.java
```diff
@@ -0,0 +1,144 @@
...
+import org.bouncycastle.pkcs.PKCS8EncryptedPrivateKeyInfo;
+import org.bouncycastle.pkcs.jcajce.JcePKCSPBEInputDecryptorProviderBuilder;
+
+public class X509Authentication {
+    private static final String BOUNCY_CASTLE = "BC";
+    private final MongoLogger logger;
+
+    static {
+        if (Security.getProvider(BOUNCY_CASTLE) == null) {
+            Security.addProvider(new BouncyCastleProvider());
```
jyemin, line 41: This makes me quite nervous, as it effects global application state.
Is there an alternative to adding a global provider like this?

bucaojit: Thanks for pointing this out.  I will try finding an alternative to adding the global provider and let you know.

bucaojit: I made a change to pass the locally instantiated provider directly to the operations that required it.  Good catch.

resources/authentication_test/X509/ca.crt
```diff
@@ -0,0 +1,32 @@
+-----BEGIN CERTIFICATE-----
```
jyemin, line 1: Should this whole `resources` directory be moved into either `src/integration_test` or `src/test`?  Located at the top-level like this, it's unclear whether it will be included in the production jar file (not claiming that it is, but the placement makes you wonder).

jyemin (commented):
Just want to confirm that there is no way to implement this feature without taking a dependency on a third-party security provider like Bouncycastle.  I have not looked into it deeply, but I would have expected that one could implement this functionality using only JDK-provided APIs.

resources/authentication_test/X509/ca.crt
bucaojit: I checked the output for `gradlew shadowJar` which outputs our uberjar and it doesn't contain the contents of this `resources/` directory.  We put our utility scripts and some test files in this directory.  I can create a ticket if this should be moved, for now it seems to be okay.

jyemin (approved):

PR #313, opened 2024-12-16
SQL-2514: Report authentication failure and expected SQLSTATE

bucaojit:
Added change to detect when a `MongoSecurityException` is hit in `connect()` to return SQLSTATE 28000 and an updated error message that authentication failed. 

Tested in Tableau and the error dialogue reports `Invalid username or password` as expected.  
Steps are in their [QA doc](https://tableau.github.io/connector-plugin-sdk/docs/manual-test#test-your-connector-with-tableau-desktop) under the section "Connect to the correct database with the wrong credentials".

Here's the output with the changes in this PR:
Tableau_Error_Invalid_creds

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoDriver.java
+++ b/src/main/java/com/mongodb/jdbc/MongoDriver.java
@@ -285,7 +285,7 @@ public class MongoDriver implements Driver {
                     if (cause instanceof com.mongodb.MongoSecurityException) {
                         throw new SQLException(
                                 "Authentication failed. Verify that the credentials are correct.",
-                                "28000",
+                                AUTHENTICATION_ERROR_SQLSTATE,
                                 e);
                     }
                     cause = cause.getCause();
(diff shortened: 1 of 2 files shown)


src/main/java/com/mongodb/jdbc/MongoDriver.java
```diff
@@ -278,6 +279,18 @@ public Connection connect(String url, Properties info) throws SQLException {
                         "Couldn't connect due to a timeout. Please check your hostname and port. If necessary, set a "
                                 + "longer connection timeout in the MongoDB URI.");
             } catch (Exception e) {
+                // Unwrap the cause to detect authentication failures
+                Throwable cause = e;
+                while (cause != null) {
+                    if (cause instanceof com.mongodb.MongoSecurityException) {
```
mattChiaravalloti, line 285: Is it possible this is thrown for any reason other than invalid username/password?

mattChiaravalloti (approved):

src/main/java/com/mongodb/jdbc/MongoDriver.java
bucaojit: In the java driver it's thrown in a few places and it can be more general than username/password like `SASL protocol error: no client response to challenge for credential`.  
It is for authentication type issues so I made the error message that we show more general to say verify credentials.  I believe it's Tableau processing the sqlstate 28000 that causes it to report the username/password even though it can be a more encompassing Invalid Authorization Specialization error.

pmeredit (approved):
LGTM!

PR #316, opened 2025-01-13
upgrade driver version to 5.3.0

terakilobyte:
Just a small one so only one reviewer. I'll also do this for the eap branch once this is approved.


bucaojit (approved):
LGTM!

PR #319, opened 2025-01-15
SQL-2534: Add THIRD-PARTY-LICENSE file generation

pmeredit:
This one generated by gradle seems good enough to me:

https://evg-bucket-mongo-jdbc-driver.s3.amazonaws.com/artifacts/678806ed64f91b0007456ed0/ssdlc/THIRD-PARTY-LICENSE.txt

To be clear: It groups all deps that share a common license together, then it embeds the license after the group. There are multiple slightly different Apache 2.0 license, which is why you see a few of those groups.


.evg.yml
```diff
@@ -252,6 +285,36 @@ functions:
...
+        shell: bash
+        working_dir: mongo-jdbc-driver
+        script: |
+          ${PREPARE_SHELL}
+          set -o errexit
+
+          echo "------------------------------------"
+          echo "Generating THIRD-PARTY-LICENSE.txt"
+          echo "------------------------------------"
+          ./gradlew clean generateLicenseReport || true
```
nbagnard, line 301: Let's move this as part of `build jdbc driver` and modify the renderer to output the file in the resources directory so that it's included in the jar file. Where in the jar, I think doesn't matter. As long as it's inside, we are fine.

You can then also get rid of the separate Evergreen tasks because I don't think that we need to publish it to S3 since the jar inside will be published already.

nbagnard (changes requested):
Looks like a great tool that you found here!
My only comments is around what we discussed this morning and where to put the file.
Can you add Nathan as a reviewer because he is doing the same for ODBC, so seems like a good way to keep things similar for both drivers.

nbagnard (changes requested):
Can you check the produced jar file content?

I don't understand why Junit dependencies show-up. These are not runtime dependencies.
Looking at the tool's github page, it is supposed to only list runtime dependencies already.
My guess is that the integration tests should be skipped/ignored for.
You can double-check with the dependencies listed in the SBOM. They should match the ones in the third-party notice.
Also, I am not sure why it adds directories for dependencies with an embedded license since it is also in the THIRDPARTY-LICENSE.txt file.

nbagnard:
> @nbagnard Do we need to header bit that's in the bic repo?
oh, yes, that'd be nice.
I don't know with that tool how easy it will be to put in in the file directory. If that's a pain, maybe we can have a separate file for the disclaimer, pointing to the THIRDPARTY-LICENSES file.
I'll let @pmeredit see what is best.

pmeredit:
@nbagnard I moved this to the eap branch

pmeredit:
> Can you check the produced jar file content?
> 
> I don't understand why Junit dependencies show-up. These are not runtime dependencies. Looking at the tool's github page, it is supposed to only list runtime dependencies already. My guess is that the integration tests should be skipped/ignored for. You can double-check with the dependencies listed in the SBOM. They should match the ones in the third-party notice. Also, I am not sure why it adds directories for dependencies with an embedded license since it is also in the THIRDPARTY-LICENSE.txt file.

Note that actually the only thing in those directories is the LICENSE files. There is no option to clean this up, unfortunately. If someone complains at some point we can split up the gradle command and do a rm -Rf on those dirs.

terakilobyte:
I see the new `THIRD-PARTY-LICENSE-DISCLAIMER.txt` file included in the jar but not the generated third party licenses file.

pmeredit:
> I see the new `THIRD-PARTY-LICENSE-DISCLAIMER.txt` file included in the jar but not the generated third party licenses file.

Bizarre, it seems somehow gradlew runs everything atomically and the generate license needs to be run separate. I thought this was running locally fine but it looks like it was just due to actually running it twice.

.evg.yml
```diff
@@ -531,17 +531,26 @@ functions:
...
-          EXTRA_PROP="-PisTagTriggered"
-        fi
-        ./gradlew -Dorg.gradle.java.home=${JAVA_HOME} $EXTRA_PROP clean -x test -x integrationTest spotlessApply build shadowjar --rerun-tasks
+    - command: shell.exec
+      type: test
+      params:
+        shell: bash
+        working_dir: mongo-jdbc-driver
+        script: |
+          #./gradlew -Dorg.gradle.java.home=${JAVA_HOME} generateLicenseReport
```
nbagnard, line 540: Can you swap the 2 lines to make sure that we always have JAVA_HOME defined and controlled.
And remove the dead code.

.evg.yml
```diff
@@ -531,17 +531,26 @@ functions:
...
+      params:
+        shell: bash
+        working_dir: mongo-jdbc-driver
+        script: |
+          ${PREPARE_SHELL}
+          if [[ "${triggered_by_git_tag}" != "" ]]; then
+            EXTRA_PROP="-PisTagTriggered"
+          fi
+          #./gradlew -Dorg.gradle.java.home=${JAVA_HOME} $EXTRA_PROP clean -x test -x integrationTest spotlessApply build shadowjar --rerun-tasks
+          ./gradlew $EXTRA_PROP clean -x test -x integrationTest spotlessApply build shadowjar --rerun-tasks
```
nbagnard, line 553: I believe that you need to keep the JAVA HOME property to make use of the toolchain and avoid issue on hosts where JAVA_HOME is not set.

nbagnard (changes requested):
I still don't see the license file in any of the jars.
https://spruce.mongodb.com/task/mongo_jdbc_driver_eap_release_build_patch_4b1bd9a6f9c32002e1cba7e78f655a2eba7da14c_6797d25a443215000729eef1_25_01_27_18_37_15/files?execution=0

pmeredit:
This solution is less pleasant, but it's not too bad. The RELEASE.md file has been updated to make sure that we update the THIRD-PARTY-LICENSE file before a release.

terakilobyte (approved):
Sucks evergreen is having issues with this, but at least it's easily repeatable to generate the file. LGTM

pmeredit:
Closed as Gone Away

PR #320, opened 2025-01-16
SQL-2489: New public signing key

terakilobyte:
In my testing these were the minimal changes to make the gradle task not fail, so here's hoping.

I also set silent back to true for the task.


nbagnard (approved):
LGTM!

I still wonder why only the base 64 encoded value is impacted

PR #323, opened 2025-01-23
update public key and set shell

terakilobyte:
This is a small one, only one reviewer needed imo. You've been helping with it so tagging you in @bucaojit


bucaojit (approved):
LGTM

PR #326, opened 2025-02-24
SQL-2659 : Update building scripts

nbagnard:
This PR changes the dependencies versions to use dynamic versions and not pinned versions.
It also integrates:
- The Evergreen changes to copy the correct files (builds and report) in our mci like bucket.
-  The generation of the thirdparty notice file


nbagnard:
It ports changes from SQL-2455, Commit 4b1bd9a, SQL-2397, SQL-2437, SQL-2436

jchemburkar (approved):
lgtm!

PR #335, opened 2025-03-24
SQL-2682: Pull mongosqltranslate library from Gradle script

bucaojit:
Removed the `fetch libmongosqltranslate` task from evergreen script and uses a gradle task that will download the files. 

In the ticket, there was a comment about whether it could be built without internet.  I didn't find a way to solve that but had an idea of caching the libraries so when there is access, they can be downloaded and the build script will use those cached files unless passed in the `-PupdateLibs=true` property.  Which is really for snapshots since those will be updated and have the same name.  I'm open to other ideas if there are any. 

Also, I deleted the existing library binaries since they will now be downloaded, since the existing ones were a much older version that wasn't working.

The code to get the version from the `LIBMONGOSQLTRANSLATE_VER` will need to be tested at release time.


terakilobyte (approved):
LGTM. I verified the all-jar for libmongosqltranslate libs!

nbagnard (approved):
LGTM!

The manifest will be out of sync for builds using the property or snapshot builds, but I think it's OK, since those are not release builds.

PR #336, opened 2025-03-25
SQL-2661: Log more info

nbagnard:


src/main/java/com/mongodb/jdbc/MongoStatement.java
```diff
@@ -264,12 +296,15 @@ private ResultSet executeDirectClusterQuery(String sql)
     public ResultSet executeQuery(String sql) throws SQLException {
         checkClosed();
         closeExistingResultSet();
-
+        logger.setSqlQuery(sql);
+        long startTime = System.nanoTime();
+        logger.log(Level.INFO, StringEscapeUtils.escapeJava(sql));
```
bucaojit, line 301:
```suggestion
        logger.log(Level.INFO, "\n" + sql);
```
Is the escapeJava added here to handle special characters or something?

If it's not needed, the query gets logged in a more readable way, at least for a simple query like this:
```
# with escapeJava()
[2025-03-26 10:22:08.094] [INFO] [c-1] [stmt-1] com.mongodb.jdbc.MongoStatement executeQuery: SELECT \"c1\".\"_id\" AS \"_id\",\n  \"c1\".\"a\" AS \"a\",\n  \"c1\".\"b\" AS \"b\",\n  \"c1\".\"c\" AS \"c\"\nFROM \"c1\"\nLIMIT 100  

# without escapeJava
[2025-03-26 11:16:48.441] [INFO] [c-1] [stmt-1] com.mongodb.jdbc.MongoStatement executeQuery: 
SELECT "c1"."_id" AS "_id",
  "c1"."a" AS "a",
  "c1"."b" AS "b",
  "c1"."c" AS "c"
FROM "c1"
LIMIT 100  
```

nbagnard: I actually added the escapeJava on purpose.
It is easier to read without it, but harder to grep.
With the query on one line, you can grep `executeQuery` and get all the query in the output. If it is on multiple lines, you can't.
The idea is that it is easy to grep the logs for queries and use search and replace the blank characters \n, \t etc

jchemburkar (approved):
lgtm!

src/main/java/com/mongodb/jdbc/MongoStatement.java
bucaojit: Yeah sounds good. 👍

PR #344, opened 2025-04-04
SQL-2538: Javadoc fixes

nbagnard:
I added a check task to verify the javadoc can be generated so that we can avoid this issue going forward.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -90,6 +90,7 @@ buildvariants:
       - name: "publish-maven"
       - name: "ssdlc-artifacts-release"
       - name: "download-center-update"
+      - name: "javadoc"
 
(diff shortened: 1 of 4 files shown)


.evg.yml
```diff
@@ -403,6 +408,16 @@ functions:
         ${PREPARE_SHELL}
         ./gradlew $IS_RELEASE_PROP spotlessCheck
 
+  "check javadoc":
```
bucaojit, line 411: Nice you already added this.  I created a ticket to add it.

bucaojit (approved):
LGTM

I'm thinking we can also re-add the `silent: true` for the `publish maven` task

PR #345, opened 2025-04-04
SQL-2732: Correct handling of empty select order list

bucaojit:
Initially only `null` was being checked.  The empty selectOrder also is an empty list, added a check for that.


mattChiaravalloti (approved):
👍

pmeredit (approved):
Nice!

PR #351, opened 2025-05-14
Fix possible NPE on MongoDriver#connect()

romanb:
The properties can contain non-string values. If connect() is called with such a property (e.g. intended for a different JDBC driver, as can happen in the context of `DriverManager.getConnection()`), the attempt to lower-case the keys in the MongoDriver results in an NPE, because getProperty() returns null for non-string properties.

The simplest patch is to use get() rather than getProperty(). Alternatively, to avoid the lower-casing if the URL is not actually intended for the MongoDriver, we could add a guard clause that uses acceptsURL().


nbagnard:
Hi @romanb,
Thank you for reporting this issue.
I am able to reproduce the behavior.
I will file an internal bug ticket and we will look at it.

pmeredit:
I replicated your PR here because we can only accept signed commits: https://github.com/mongodb/mongo-jdbc-driver/pull/354

Thanks!

PR #354, opened 2025-05-19
SQL-2831: fix properties issue in Driver#Connect where non-String keys or values would cause NPE

pmeredit:
Based on original user PR: https://github.com/mongodb/mongo-jdbc-driver/pull/351

### DIFF
--- a/src/integration-test/java/com/mongodb/jdbc/integration/ADFIntegrationTest.java
+++ b/src/integration-test/java/com/mongodb/jdbc/integration/ADFIntegrationTest.java
@@ -92,25 +92,19 @@ public class ADFIntegrationTest {
 
     public MongoConnection getBasicConnection(String db, Properties extraProps, String uriOptions)
             throws SQLException {
-        System.out.println("Connecting to database: " + db);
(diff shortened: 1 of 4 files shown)


nbagnard (changes requested):
We should also change the logic to lower case the keys so that if someone has a valid mongo jdbc uri, but is for some reason passing a Properties object with a non-string value, we fail more gracefully than with a NPE in connect().

pmeredit:
> We should also change the logic to lower case the keys so that if someone has a valid mongo jdbc uri, but is for some reason passing a Properties object with a non-string value, we fail more gracefully than with a NPE in connect().

`Properties` was really poorly thought out.

nbagnard (changes requested):
It's not just keys, but also values that we need to watch for.
Unfortunately, not everyone is a fan of using proper throws declaration for their methods and then this is how you end up with this unexpected exceptions :'(

nbagnard (changes requested):

nbagnard (approved):
LGMT :)

PR #357, opened 2025-05-27
Add option to allow disabling the MongoClient cache.

romanb:
In a server environment where the mongo-jdbc driver is used to connect to a variable number of customer databases to run queries at varying frequencies, it is not desirable to keep the MongoClients alive/open across JDBC connections, especially since every MongoClient comes with a background server monitoring thread.

This PR adds an option to disable the mongo client cache, so that the client is owned by the JDBC connection and closed together with it, as was the case before the cache was introduced in https://github.com/mongodb/mongo-jdbc-driver/pull/260.


nbagnard:
Hi Roman,

We have an item in our backlog already for addressing this.
The difference is that we don't plan to go exactly as you are suggesting, but I think it would address your problem.
We want to close the clients and clear the cache once there is no more linked connections.
The work is not currently on our roadmap, but I will bring it up at our next stand up.

pmeredit:
We did a different version of this that has been merged and released in 2.3.0

PR #358, opened 2025-06-10
SQL-2850: JDBC OIDC default scope improvement

bucaojit:
Adding `openid` and `offline_access` by default.  If `clientID + "/.default"` is in the requested scopes will add that as well.  

Verified that it works with our Okta test instance `mongodb://sqlenginesoidctest-lxonw.a.query.mongodb-dev.net` and an Azure idp instance.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
+++ b/src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
@@ -86,7 +86,7 @@ public class OidcAuthFlow {
         // Add custom scopes from request that are supported by the IdP
         List<String> requestedScopes = idpServerInfo.getRequestScopes();
         if (requestedScopes != null) {
-            // Always add clientID/.default if it's in requestedScopes
+            // Always add clientID/.default if it's in requestedScopes, this was needed for Azure OIDC.
             String clientIDDefault = clientID + "/.default";
             if (requestedScopes.contains(clientIDDefault)) {
                 scopes.add(clientIDDefault);
--- a/src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
+++ b/src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
@@ -44,7 +44,9 @@ import com.nimbusds.openid.connect.sdk.op.OIDCProviderMetadata;
 import java.io.IOException;
 import java.net.URI;
 import java.time.Duration;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import javax.security.auth.RefreshFailedException;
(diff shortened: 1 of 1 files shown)


src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -54,13 +56,58 @@ public class OidcAuthFlow {
...
+            String clientIDDefault = clientID + "/.default";
+            if (requestedScopes.contains(clientIDDefault)) {
+                scopes.add(clientIDDefault);
+            }
+            for (String scope : requestedScopes) {
+                if (supportedScopes != null) {
+                    if (supportedScopes.contains(scope)) {
+                        scopes.add(scope);
+                    }
+                }
```
nbagnard, line 100: Let's keep the previous logic with the warning, since we can potentially "ditch" scopes.
```suggestion
if (supportedScopes != null && supportedScopes.contains(scope)) {
                        scopes.add(new Scope.Value(scope));
                    } else {
                        logger.warning(
                                "Requested scope '" + scope + "' is not supported by the IdP");
                    }
```

nbagnard (approved):
LGTM! 
Just have to 2 small suggestions.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
```diff
@@ -54,13 +56,58 @@ public class OidcAuthFlow {
...
+     * scopes.
+     *
+     * @param clientID the OAuth2 client identifier
+     * @param idpServerInfo server information containing requested scopes and other IdP details
+     * @param providerMetadata OIDC provider metadata containing supported scopes
+     * @return a Scope object containing all valid scopes to be requested
+     */
+    public Scope buildScopes(
+            String clientID, IdpInfo idpServerInfo, OIDCProviderMetadata providerMetadata) {
+        Set<String> scopes = new HashSet<>();
```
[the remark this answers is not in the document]
pmeredit: I can't imagine it ever being more than 5-6 things.

pmeredit (approved):
LGTM! with the warning

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
pmeredit: feels a bit odd to check null every time, but as you mentioned this won't be a long loop!

bucaojit: Oh, good catch on the warning I re-added it.  
I also moved the null check to outside of the loop.

src/main/java/com/mongodb/jdbc/oidc/OidcAuthFlow.java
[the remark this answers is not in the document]
bucaojit: Yeah, it won't be a big list.  When I configured the Azure Idp it was 3 entries.  I'll keep this the way it is.

PR #363, opened 2025-07-18
SQL-2911: Updated the X509 logic

nbagnard:
Handle more types of PEM objects and log more diagnostic information.


src/main/java/com/mongodb/jdbc/utils/X509Authentication.java
```diff
@@ -67,51 +74,144 @@ private SSLContext createSSLContext(String pemPath, char[] passphrase) throws Ex
...
+            logger.log(
+                    Level.SEVERE,
+                    "An unexpected error occurred during PEM object processing: "
+                            + e.getClass().getSimpleName()
+                            + " - "
+                            + e.getMessage());
+            throw e;
         }
 
         if (privateKey == null) {
```
bucaojit, line 209: Looking through this code again, I'm wondering if the private key just wasn't present or being processed.  We were seeing this constantly being logged but I think the error message I put here was throwing me off, that the reading of the file was the issue...

I'm going to suggest updating that, you can modify the error message if a clearer one comes to mind.  Also, here's a change so that it will log that the class is missing as well if that's the case. 

```
        StringBuilder missingComponents = new StringBuilder();
        if (privateKey == null) {
            missingComponents.append("Private key not found in the PEM file (encrypted or unencrypted)");
        }
        if (cert == null) {
            if (missingComponents.length() > 0) {
                missingComponents.append("\n");
            }
            missingComponents.append("X.509 certificate not found in the PEM file");
        }
        if (missingComponents.length() > 0) {
            throw new MongoException(missingComponents.toString());
        }
```

src/main/java/com/mongodb/jdbc/utils/X509Authentication.java
```diff
@@ -16,19 +16,26 @@
 
 package com.mongodb.jdbc.utils;
 
+import com.mongodb.MongoException;
```
bucaojit, line 19: We have an [integration test](https://github.com/mongodb/mongo-jdbc-driver/blob/7dbaf15e0b93b55fab0c3f76a6d0ec32ada38b4c/src/integration-test/java/com/mongodb/jdbc/integration/AuthX509IntegrationTest.java) for the PKCS#8 type PEM file.  Just in case, how about we create a ticket to add a test for confirming the PKCS#1 type PEM file? It can be something backlogged for if it is necessary to test.

bucaojit (changes requested):
Looks good, this will help us with troubleshooting these X509 issues.  I had some logging suggestions.

PR #365, opened 2025-07-24
SQL-2914: Disallow Maven publishing for eap builds

bucaojit:
The JDBC Release is failing with an error related to the `closeAndReleaseRepository` task which shouldn't be run for EAP builds.  Added a check for the `isEapBuild` property. 

Added a proposed fix for the standard build as well, but that wouldn't be tested until we release a non-eap build.


nbagnard (approved):
LGTM!

PR #367, opened 2025-08-15
SQL-2920: Add configuration properties to support GSSAPI

bucaojit:
Adding properties:
`jaasconfigpath`: Set a custom jaas.config file path if the default doesn't work or can't be modified.  This was used since the default jaas.conf that was getting picked up by Tableau was in the `jdbcserver.jar` needed to be changed to get a working connection.
`gssnativemode`: Used on Windows to allow the use of MIT Kerberos vs the native Windows Kerberos.  Set this to `false` to use MIT Kerberos. 

Tested on Mac with a locally setup KDC using Docker containers. 
Tested on Windows against ldaptest.10gen.cc using MIT Kerberos.

### DIFF
--- a/src/main/java/com/mongodb/jdbc/MongoConnection.java
+++ b/src/main/java/com/mongodb/jdbc/MongoConnection.java
@@ -16,6 +16,7 @@
 
 package com.mongodb.jdbc;
 
+import static com.mongodb.AuthenticationMechanism.GSSAPI;
 import static com.mongodb.AuthenticationMechanism.MONGODB_OIDC;
 import static com.mongodb.AuthenticationMechanism.MONGODB_X509;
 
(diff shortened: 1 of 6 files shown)


jchemburkar (approved):
lgtm!

bucaojit:
@nbagnard sure, I'll look into adding an integration test with `ldaptest.10gen.cc` when I get back to working on this ticket 👍

bucaojit:
Took some time to get the integration test working with `ldaptest.10gen.cc` to get the secrets and config files working.  Then it turns out it hits the `'Unrecognized pipeline stage name: '$documents''` error because it's an older version of MongoDB.  It doesn't hit an `authentication failed` error, which I hit when trying to authenticate username/pass with the wrong password.  So this seems to me that it gets past the authentication part. 

Retagging @jchemburkar as the latest commit has quite a bit of changes. 

I still have to re-test again on Windows.  This works for me on Mac, and the integration test is run on Linux.

nbagnard (approved):
LGTM! Thank you for setting up the integration test :)

bucaojit:
Confirmed it works on Windows, to the point where it gets the `$documents` error that we expect. 

Settings in Main.java:
```java
   static final String URL = "jdbc:mongodb://drivers%40LDAPTEST.10GEN.CC@ldaptest.10gen.cc/?authSource=$external&authMechanism=GSSAPI";
   
   
         p.setProperty("database", "test");
         p.setProperty("jaasconfigpath", "C:\\kerberos\\jaas.conf");
         p.setProperty("gssnativemode", "false");
         p.setProperty("gssapiserverauth", "false");
```

jaas.conf
```
com.sun.security.jgss.initiate {
    com.sun.security.auth.module.Krb5LoginModule required
    useTicketCache=true
    doNotPrompt=true
    ticketCache="C:/kerberos/kerbfile"
    debug=true;
};
```

PR #375, opened 2025-09-19
SQL-2962: Shade Bouncy Castle dependency to avoid mismatch

bucaojit:


nbagnard (approved):
LGTM!
Do you see any risks to generally shade every dependencies? If not, it seems worth doing it at the same time.

PR #382, opened 2025-12-19
SQL-3029: Update test expired test certificates

nbagnard:
The new expiry date is Nov 25th  2125


bucaojit:
For the test failure if you change this line to 6 instead of 5 it should work.  Or you can add > 5 or something to be less strict, although we probably won't be changing these until 2125 😄 
https://github.com/mongodb/mongo-jdbc-driver/blob/663e8be94ea202a86f6bc4cbaa7fee99c649bfdd/src/test/java/com/mongodb/jdbc/utils/X509AuthenticationTest.java#L319

EthanHardyMongo (approved):
LGTM!

bucaojit (approved):
LGTM

PR #383, opened 2026-01-08
SQL-3022: Update all catch Exception blocks to include message in thrown SQLException

mattChiaravalloti:
This PR adds root cause exception messages to `SQLException`s that are thrown after catching `Exception`s. The ticket highlights connection exceptions in particular, but also suggested doing this anywhere in the codebase where we did `catch (Exception e)` and subsequently threw a `SQLException`. I found 9 instances of this behavior and updated all 9 to include `e.getMessage()` as part of the thrown exception message.

### DIFF
--- a/src/test/java/com/mongodb/jdbc/MongoDriverTest.java
+++ b/src/test/java/com/mongodb/jdbc/MongoDriverTest.java
@@ -1055,9 +1055,17 @@ class MongoDriverTest {
                 e.getMessage());
     }
 
+    static class MockMongoDriver extends MongoDriver {
+        @Override
+        protected MongoConnection getUnvalidatedConnection(String url, Properties info)
+                throws SQLException {
+            throw new SQLException(new com.mongodb.MongoSecurityException(null, "exception"));
+        }
+    }
+
     @Test
     void testConnectExceptionContainsRootCauseForAuthExceptions() throws Exception {
-        MongoDriver d = new MongoDriver();
+        MockMongoDriver d = new MockMongoDriver();
         Properties p = new Properties();
         p.setProperty(DATABASE.getPropertyName(), "test");
 
@@ -1066,7 +1074,7 @@ class MongoDriverTest {
 
         Exception e = assertThrows(SQLException.class, () -> d.connect(url, p));
         assertEquals(
-                "Authentication failed. Verify that the credentials are correct. Root cause: com.mongodb.MongoSecurityException: Failed to login Subject",
+                "Authentication failed. Verify that the credentials are correct. Root cause: com.mongodb.MongoSecurityException: exception",
                 e.getMessage());
     }
 
--- a/src/test/java/com/mongodb/jdbc/MongoDriverTest.java
+++ b/src/test/java/com/mongodb/jdbc/MongoDriverTest.java
@@ -21,6 +21,8 @@ import static com.mongodb.jdbc.utils.X509AuthenticationTest.TEST_PEM_DIR;
 import static org.junit.jupiter.api.Assertions.*;
 
 import com.mongodb.AuthenticationMechanism;
+import com.mongodb.ConnectionString;
+import com.mongodb.MongoConfigurationException;
 import com.mongodb.client.MongoClient;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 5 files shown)


src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -969,10 +971,11 @@ void testGssNativeMode_InvalidValue_ThrowsSQLException() throws SQLException {
         String url = userNoPWDURL + "?authMechanism=GSSAPI";
         p.setProperty(GSS_NATIVE_MODE.getPropertyName(), "invalid");
 
-        assertThrows(
+        Exception e = assertThrows(
                 SQLException.class,
                 () -> d.getUnvalidatedConnection(url, p),
                 "Invalid gssnativemode value should throw SQLException");
+        assertEquals("Failed to create connection. Root cause: Invalid gssnativemode property value: invalid. Valid values are: 'true', 'false'.", e.getMessage());
```
mattChiaravalloti: This test already checked this exceptional case, so I decided to enhance the test to check that the root cause is included.

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -1031,4 +1034,43 @@ void testTlsCaFileMismatchFails() throws Exception {
                 () -> d.getUnvalidatedConnection(url, p),
                 "The connection should fail because of a tlscafile mismatch between the URI and the properties");
     }
+
```
mattChiaravalloti, line 1038: Added new tests for the other cases in this class that could be tested. There is one additional spot that I cannot easily/directly test: `sqlURLEncode`. I think it is okay to skip testing for that since it is trivially obvious that the updated strings will be returned as part of the thrown `SQLException`s.

src/test/java/com/mongodb/jdbc/MongoResultSetTest.java
```diff
@@ -1301,4 +1311,45 @@ void sameMetaDataOnDifferentRowsEvenWithDifferentBsonTypes() throws SQLException
         assertEquals(1, metaData.getColumnCount());
         assertEquals(Types.INTEGER, metaData.getColumnType(1));
     }
+
```
mattChiaravalloti, line 1314: A similar situation applies to this class. I could not easily tests `next` since none of the calls in its `try` block have checked exceptions. Without checked exceptions, it seems `Mockito` cannot mock throwing. Similar to above, it is trivial to see that the exception string is updated and will therefore include the root cause of whatever exception is caught so incomplete testing is okay.

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -1031,4 +1035,47 @@ void testTlsCaFileMismatchFails() throws Exception {
...
+                        SQLException.class,
+                        () -> d.connect(url, p),
+                        "Invalid gssnativemode value should throw SQLException");
+        assertEquals(
+                "Connection failed. Root cause: Failed to create connection. Root cause: Invalid gssnativemode property value: invalid. Valid values are: 'true', 'false'.",
+                e.getMessage());
+    }
+
+    @Test
+    void testConnectExceptionContainsRootCauseForAuthExceptions() throws Exception {
```
mattChiaravalloti, line 1067: Ah, I was testing locally with a mongod running so I was able to trivially trigger the case for this test. We don't have mongod running for our unit tests, of course, so I'll have to rework this one. The general structure of the test will be the same, it is just a matter of forcing the correct exception to trigger the correct code path.

nbagnard (approved):
LGTM!

The ones you could check with unit tests can be added to integration tests if you want.
But I think we are fine as is.

src/test/java/com/mongodb/jdbc/MongoDriverTest.java
```diff
@@ -1031,4 +1035,55 @@ void testTlsCaFileMismatchFails() throws Exception {
...
+        String url = userNoPWDURL + "?authMechanism=GSSAPI";
+        p.setProperty(GSS_NATIVE_MODE.getPropertyName(), "invalid");
+
+        Exception e =
+                assertThrows(
+                        SQLException.class,
+                        () -> d.connect(url, p),
+                        "Invalid gssnativemode value should throw SQLException");
+        assertEquals(
+                "Connection failed. Root cause: Failed to create connection. Root cause: Invalid gssnativemode property value: invalid. Valid values are: 'true', 'false'.",
```
bucaojit, line 1054: Nice work on the thorough testing.  This `Connection failed.` error is the one I was focusing on since it's the one we were hitting with the Tableau Cloud failure.  

I was looking for a way to add a test to have a RuntimeException be thrown to verify that it gets reported but my attempts weren't working.  I did try it manually by adding code to throw a RuntimeException in `ConnValidation.call()` which worked so I think we're good.

bucaojit (approved):
LGTM

PR #385, opened 2026-02-10
SQL-2998, SQL-2807: Update SBOM push/pull using Silkbomb and other common SSDLC functionality via common-test-infra

mattChiaravalloti:
This PR updates the evergreen configs to remove many SSDLC-related functions in favor of common versions in `sql-engines-common-test-infra`. The primary goal was to update `augment sbom` to use new logic described in the ticket; when that functionality was added to `common-test-infra`, it could not be used here until the entire project was set up to use `common-test-infra` for all SSDLC functions. That is why this PR includes updates to all SSDLC functions.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -274,7 +274,16 @@ tasks:
           local_file_path: ${SSDLC_DIR}/${STATIC_CODE_ANALYSIS_NAME}
           published_file_path: ${working_dir}/${STATIC_CODE_ANALYSIS_NAME}
       - func: "generate compliance report"
+        vars:
+          put_to_bucket: "evg-bucket-mongo-jdbc-driver"
+          published_sbom_path: ${working_dir}/${SBOM_FILENAME}
+          published_sarif_path: ${working_dir}/${STATIC_CODE_ANALYSIS_NAME}
+          compliance_report_template_path: ${PROJECT_DIRECTORY}/resources/release/mongo_jdbc_compliance_report_template.md
       - func: "publish compliance report"
+        vars:
+          get_from_bucket: "evg-bucket-mongo-jdbc-driver"
+          local_file_path: ${SSDLC_DIR}/${COMPLIANCE_REPORT_NAME}
+          published_file_path: ${working_dir}/${COMPLIANCE_REPORT_NAME}
 
   - name: ssdlc-artifacts-snapshot
     run_on: ubuntu2204-small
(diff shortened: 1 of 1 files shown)


.evg.yml
```diff
@@ -233,14 +238,22 @@ tasks:
   - name: semgrep
     exec_timeout_secs: 3600 # 1h
     commands:
-      - func: "static code analysis"
+      - func: "generate static code analysis"
```
nbagnard, line 241: You'll need to add a variable for the ruleset to use in the common [script](https://github.com/mongodb/sql-engines-common-test-infra/blob/bf2fa317d637fad9320b8bdcfc549e1d87cd3315/evergreen/scripts/generate_static_code_analysis.sh#L23C1-L23C192).
For most of our projects we use `semgrep --config p/rust` because those are Rust projects.
For the JDBC driver, we need to use `semgrep --config p/java`.

Right now, it is not doing anything. If you look at the verbose logs: "Ran 9 rules on 0 files: 0 findings."

You'll also want to add a variable for setting a different value for `--exclude`. Otherwise, all projects will exclude the "integration_test". For the JDBC driver, we exclude incorrectly "vendor" which is a copy-pasting left over from the BIC. The default exclusion list for Java is enough because there is no vulnerability reported. Which means that we could forgo the `--exclude` option altogether.  We can also set it to "--exclude demo" because demo is an internal helper.

.evg.yml
```diff
@@ -253,9 +266,26 @@ tasks:
...
+        vars:
+          get_from_bucket: "evg-bucket-mongo-jdbc-driver"
+          local_file_path: ${SSDLC_DIR}/${STATIC_CODE_ANALYSIS_NAME}
+          published_file_path: ${working_dir}/${STATIC_CODE_ANALYSIS_NAME}
       - func: "generate compliance report"
+        vars:
+          put_to_bucket: "evg-bucket-mongo-jdbc-driver"
+          published_sbom_path: ${working_dir}/${SBOM_FILENAME}
+          published_sarif_path: ${working_dir}/${STATIC_CODE_ANALYSIS_NAME}
+          compliance_report_template_path: ${PROJECT_DIRECTORY}/resources/release/mongo_jdbc_compliance_report_template.md
```
nbagnard, line 283: I think you can use the generic compliance report template: evergreen/resources/compliance_report_template.md

This way, you will have the "Compliance Report Created Date" info added. I saw no other difference.
You will need to add ```signing_section_bookmark: "verify-integrity-of-mongodb-jdbc-driver-packages"``` and ```repo_name: "mongo-jdbc-driver"```.
This will get the correct link to the README section related to verify the integrity of the artifacts.

.evg.yml
```diff
@@ -268,65 +298,28 @@ tasks:
...
+        vars:
+          get_from_bucket: "evg-bucket-mongo-jdbc-driver"
+          local_file_path: ${ARTIFACTS_DIR}/${SBOM_FILENAME}
+          published_file_path: ${working_dir}/${SBOM_FILENAME}
       - func: "publish static code analysis"
+        vars:
+          get_from_bucket: "evg-bucket-mongo-jdbc-driver"
+          local_file_path: ${SSDLC_DIR}/${STATIC_CODE_ANALYSIS_NAME}
+          published_file_path: ${working_dir}/${STATIC_CODE_ANALYSIS_NAME}
       - func: "generate compliance report"
```
nbagnard, line 310: Same changes than for the release task above apply here.

.evg.yml
```diff
@@ -476,23 +401,26 @@ functions:
         content_type: application/json
         permissions: public-read
 
-  "scan sbom":
+  "scan sbom for jdbc":
```
nbagnard, line 404: You can use [scan_sbom](https://github.com/mongodb/sql-engines-common-test-infra/blob/main/evergreen/scripts/scan_sbom.sh)  in the shared test infra repo. All projects are using Grype and the failure threshold is the same.

nbagnard (changes requested):
Pretty good cleaning work making this all more generic!
I just noticed a few things to adjust and/or where the shared test infra logic could be used.

bucaojit (approved):
Nice work cleaning up the evergreen script. 

The changes look good, I agree with Natacha's comments.

nbagnard:
Note: I am reverting the config file name until this PR is back. It otherwise will block the release process.

PR #388, opened 2026-04-15
SQL-3184: Dynamic dependency upgrade only pull final, ga or release version

nbagnard:
The existing logic would pull the latest version of a dependency and that meant pulling alpha, rc or beta releases if they were available.
For example, for the Java driver, it would  currently pull `5.7.0-beta1`.
We want to stick to stable versions.

### DIFF
--- a/.evg.yml
+++ b/.evg.yml
@@ -189,6 +189,8 @@ tasks:
         variant: "release"
       - name: "test-smoke"
         variant: "release"
+      - name: "make-docs"
+        variant: "release"
     commands:
       - *assume_role_cmd
       - command: s3.get
--- a/.evg.yml
+++ b/.evg.yml
@@ -65,6 +65,7 @@ buildvariants:
     tasks:
       - name: semgrep
       - name: sbom
+      - name: augment-sbom
       - name: ssdlc-artifacts-snapshot
 
   - name: ubuntu2204-64-jdk-8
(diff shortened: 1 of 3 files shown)


.evg.yml
```diff
@@ -239,9 +240,16 @@ tasks:
     commands:
       - func: "generate sbom"
       - func: "upload sbom"
-      - func: "augment sbom"
       - func: "scan sbom"
 
+  - name: augment-sbom
```
nbagnard, line 247: I moved this logic out because will I was working on the logic to excluded non stable builds, one of my patch pulled an older version of bouncy castle and then because we pushed it, Kondukto produced a VULN ticket only applies for that one-off messed up patch.

jchemburkar (approved):
lgtm!

.evg.yml
EthanHardyMongo: So now only periodic builds and github tags will trigger this task? Basically, this prevents patches from inadvertently publishing an SBOM and creating a fake VULN ticket?

EthanHardyMongo (approved):

.evg.yml
nbagnard: Yes :) 
Or this is the intent. If DEVPROD has more unknown voodoo magic going on, I am not aware of it.

PR #396, opened 2026-09-15
SQL 3447: Update LDAP JDBC Driver LDAP Config

jpowell-mongo:
## Summary

This PR updates the LDAP config to use `LDAPTEST.BUILD.10GEN.CC` matching the new LDAP config ahead of the brownout for the legacy `ldaptest.10gen.cc`. 

I've also updated the `AuthGSSAPIIntegrationTest.java` to just use the isValid() method instead of the validation we had around using dbMetadata. I also added some extra properties for debugging in the future.

## Testing

I validated this passes in CI and updated the evergreen.yaml to run the GSSAPI tests now.

## Note

I wasn't sure about the original logic of the test to call `dbmd.getCatalogs();`. It wasn't clear to me that calling GetCatalogs() and expecting to throw an error added value to the test itself. So I've just reduced to just calling isValid() - which we could also argue is redundant as establishing a connection also does a `SELECT 1;`


mattChiaravalloti (approved):
lgtm 💯 

I agree with your changes to the test. I'm pretty confused what the test was even doing before. The new test implementation seems to directly test that connectivity works, which should be the point of this test!

jchemburkar (approved):

